PR #7, opened 2021-04-30
Dholladay00/eospac fortran

dholladay00:
This PR will enable EOSPAC as an eos model for codes that pull in singularity-eos as a library. It also includes a minor fix that was causing build failures on some machines.

### DIFF
--- a/eos/singularity_eos.f90
+++ b/eos/singularity_eos.f90
@@ -39,6 +39,7 @@ module singularity_eos
     init_sg_DavisReactants_f,&
     init_sg_SpinerDependsRhoT_f,&
     init_sg_SpinerDependsRhoSie_f,&
+    init_sg_eospac_f,&
     get_sg_eos_f,&
     finalize_sg_eos_f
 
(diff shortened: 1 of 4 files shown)


eos/singularity_eos.f90
```diff
@@ -78,33 +80,33 @@ end function init_sg_Gruneisen
...
 
   interface
     integer(kind=c_int) function &
-      init_sg_DavisReactants(matid, eos, rho0, e0, P0, T0, A, B, C, G0, Z,&
+      init_sg_DavisReactants(matindex, eos, rho0, e0, P0, T0, A, B, C, G0, Z,&
                           alpha, Cv0) &
       bind(C, name='init_sg_DavisReactants')
       import
-      integer(c_int), value, intent(in)      :: matid
+      integer(c_int), value, intent(in)      :: matindex
```
Yurlungur, line 109: Why this rename?

Yurlungur (approved):
Sorry I didn't see this earlier. Did you try it with and without EOSPAC to make sure it all still works?

Next time @Yurlungur instead of @jonahm-LANL . @Yurlungur is the account I pay more attention to.

Yurlungur:
@dholladay00 go ahead and merge if you like or I can merge if you prefer

eos/singularity_eos.f90
dholladay00: I make sure to test with/without EOSPAC enabled. One complicating factor could be that I will want eospac to install the sesame2spiner executable but then not want the model to be available for the GPU build of the library. I guess as long as there are appropriate caveats and/or protections this should be ok, assuming I that combo is buildable and the compiler doesn't error out.

PR #9, opened 2021-05-10
Minor tweaks and fixes for quality of life downstream

Yurlungur:
This PR contains some minor fixes for quality of life downstream. In particular:
- Warnings are eliminated for cuda builds. Most significantly, I remove the `__host__ __device` tags from defaulted functions, which removes a bunch of warnings. However, I also hide some unused variables, etc.
- I fix some includes in `stellarcollapse2spiner` and disable building this tool by default.

### DIFF
--- a/eos/eos.hpp
+++ b/eos/eos.hpp
@@ -76,11 +76,10 @@ public:
   // move semantics ensures dynamic memory comes along for the ride
   ScaledEOS(T &&t, const Real scale)
       : t_(std::forward<T>(t)), scale_(scale), inv_scale_(1. / scale) {}
-  PORTABLE_FUNCTION ScaledEOS() = default;
-  PORTABLE_FUNCTION ScaledEOS &operator=(ScaledEOS &&) = default;
-  PORTABLE_FUNCTION ScaledEOS &operator=(const ScaledEOS &) = default;
-  PORTABLE_FUNCTION ScaledEOS(const ScaledEOS &) = default;
-  // PORTABLE_FUNCTION ~ScaledEOS() = default;
+  ScaledEOS() = default;
+  ScaledEOS &operator=(ScaledEOS &&) = default;
+  ScaledEOS &operator=(const ScaledEOS &) = default;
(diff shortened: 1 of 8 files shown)


eos/eos.hpp
```diff
@@ -222,11 +220,10 @@ template <typename T> class ShiftedEOS {
 public:
   // move semantics ensures dynamic memory comes along for the ride
   ShiftedEOS(T &&t, const Real shift) : t_(std::forward<T>(t)), shift_(shift) {}
-  PORTABLE_FUNCTION ShiftedEOS() = default;
-  PORTABLE_FUNCTION ShiftedEOS &operator=(ShiftedEOS &&) = default;
-  PORTABLE_FUNCTION ShiftedEOS &operator=(const ShiftedEOS &) = default;
-  PORTABLE_FUNCTION ShiftedEOS(const ShiftedEOS &) = default;
-  // PORTABLE_FUNCTION ~ShiftedEOS() = default;
+  ShiftedEOS() = default;
```
dholladay00, line 223: If we're removing the device markings (which I think we'll regret when we try to remove Cuda warnings in earnest), why not remove all of the `=default` code as well?

Yurlungur: The device markings are apparently automatic for defaulted functions. This compiles and runs on device with zero warnings.

As for removing the defaulted functions completely.... yes, we probably should... I'm not sure which of those are deleted if I don't add them. Probably only the default constructor, which we need. I can give that a try.

dholladay00: ooooohhhhh, this removed the warnings!! That's great! Please try that out, code removal is good.

Yurlungur: Done. We can remove everything except the default constructor, which needs to be explicitly defaulted so it doesn't get deleted. (And we need it for kokkos views.)

Yurlungur:
Fantastic. Thanks, @dholladay00. Merging.

Yeah I remember the fighting with the default constructor for kokkos views with CJ. :D

PR #14, opened 2021-05-29
add histogram for profile stellar collapse

Yurlungur:
Instrument root finding on the host.


dholladay00:
This looks like things we have done before with EOS models. Seems fairly straightforward to me.

dholladay00 (approved):

PR #15, opened 2021-06-07
Dholladay00/scaleshift fortran

dholladay00:
This MR enables passing scaling and shifting parameters through the higher level API (both c/c++ and fortran). The high level API applies the shift and then the scale if both are present for consistency with host codes. 

I still need to bump version numbers and I'm curious with @Yurlungur thinks about how the params are passed in? Should they be defaulted to `nullptr`, should I explicitly overload such that the current call signatures are still valid?

### DIFF
--- a/eos/eos_builder.hpp
+++ b/eos/eos_builder.hpp
@@ -92,8 +92,8 @@ namespace EOSBuilder {
 
   template<typename T>
   EOS applyShiftAndScale(T&& eos,
-			 bool scaled, bool shifted,
-			 Real scale, Real shift) {
+                         bool scaled, bool shifted,
+                         Real scale, Real shift) {
     if (shifted && scaled) {
       ShiftedEOS<T> a(std::forward<T>(eos),shift);
       ScaledEOS<ShiftedEOS<T>> b(std::move(a),scale);
--- a/eos/eos_builder.hpp
+++ b/eos/eos_builder.hpp
@@ -74,8 +74,8 @@ namespace EOSBuilder {
 
   template<typename T>
   EOS applyScaleAndShift(T&& eos,
-					   bool scaled, bool shifted,
-					   Real scale, Real shift) {
+			 bool scaled, bool shifted,
+			 Real scale, Real shift) {
     if (shifted && scaled) {
       ScaledEOS<T> a(std::move(eos),scale);
       ShiftedEOS<ScaledEOS<T>> b(std::move(a),shift);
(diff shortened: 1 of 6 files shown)


eos/eos_builder.hpp
```diff
@@ -89,6 +89,24 @@ namespace EOSBuilder {
...
+      return b;
+    }
+    if (shifted) {
+      return ShiftedEOS<T>(std::forward<T>(eos),shift);
+    }
+    if (scaled) {
+      return ScaledEOS<T>(std::forward<T>(eos),scale);
+    }
+    return eos;
+  }
```
Yurlungur, line 109: If this is the canonical form used in rage, should we remove support for the other one?

eos/singularity_eos.cpp
```diff
@@ -34,39 +35,48 @@ int init_sg_eos(const int nmat, EOS* &eos) {
   return 0;
 }
 
+#define SGAPPLYMOD(A) \
+  EOSBuilder::applyShiftAndScale(A, enabled[0] == 1, enabled[1] == 1,\
+                                 vals[0], vals[1])
```
Yurlungur, line 40: Clever.

Yurlungur (commented):
>  I'm curious with @Yurlungur thinks about how the params are passed in? Should they be defaulted to nullptr, should I explicitly overload such that the current call signatures are still valid

I think I prefer overloads so the old call signatures are still valid.

Yurlungur (approved):

dholladay00:
> > I'm curious with @Yurlungur thinks about how the params are passed in? Should they be defaulted to nullptr, should I explicitly overload such that the current call signatures are still valid
> 
> I think I prefer overloads so the old call signatures are still valid.

It appears that since these are `extern "C"` functions, overloads are not allowed. I'm going to try default arguments.

eos/eos_builder.hpp
dholladay00: I'm not sure it hurts anything to leave it there, but I'm not opposed to removing it either.

Yurlungur: Well let's leave it for now, I suppose. Because riot does it that way. But we should probably unify that at some point.

Yurlungur:
> This is pretty much ready to go. I just need to bump version #, do you want to bump major, minor, or patch number @Yurlungur

I think minor makes sense. What do you think?

dholladay00:
> > This is pretty much ready to go. I just need to bump version #, do you want to bump major, minor, or patch number @Yurlungur
> 
> I think minor makes sense. What do you think?

I don't thing singularity has set official standards for this, but many models bump major version when the API changes. We have only changed the API for non-c++ codes, I could go either way, but I'm inclined to agree with you.

Yurlungur:
> > > This is pretty much ready to go. I just need to bump version #, do you want to bump major, minor, or patch number @Yurlungur
> > 
> > 
> > I think minor makes sense. What do you think?
> 
> I don't thing singularity has set official standards for this, but many models bump major version when the API changes. We have only changed the API for non-c++ codes, I could go either way, but I'm inclined to agree with you.

Makes sense. Let's do that then.

dholladay00:
I'll attempt to tag this commit in main as 1.4.0 after merged @Yurlungur

PR #16, opened 2021-06-24
WIP: Starting an implementation of a ramp modifier. This is not close to b…

dholladay00:
…eing fully implemented, but this is will be helpful for discussion.

@Yurlungur, @jdolence, @chadmeyer, Jeff Peterson (I can't find his handle)

- [ ] reversible discussion
- [ ] FillEOS finish
- [ ] delta e finish
- [ ] caching max discussion
- [ ] caching alpha discussion
- [ ] ctor, other necessary member functions

### DIFF
--- a/eos/eos.hpp
+++ b/eos/eos.hpp
@@ -467,17 +467,15 @@ template <typename T> class EAPRampEOS {
   auto alpha_impl(F&& func, const Real rho, const Real arg2,
 		  Real* lambda = nullptr) const {
     constexpr const Real a_eps {1.e-12};
+    const auto f = [&] (const Real Pl) {
+      const Real a {this->alpha_p(Pl)};
+      return Pl - func(a*rho, arg2, lambda) / a;
+    };
     Real a {0.0};
-    Real a_old {alpha0_};
-    Real diff {0.0};
-    Real P {0.0};
-    for (int i {0}; i < 10; ++i) {
-      P = func(a_old*rho, arg2, lambda) / a_old;
-      a = this->alpha_p(P);
-      diff = fabs(a - a_old) / fabs(a + a_old) * 2.0;
-      if (diff < a_eps) break;
-      a_old = a;
-    }
+    RootFinding1D::RootCounts rc {};
+    RootFinding1D::findRoot(f, 0.0, alpha0_, alpha0_/100.0, alpha0_*100.0,
+			    a_eps, a_eps, a, rc);
+    const Real P {func(a*rho, arg2, lambda) / a};
     return std::tie(a, P);
   }
 
--- a/eos/eos.hpp
+++ b/eos/eos.hpp
@@ -464,7 +464,7 @@ template <typename T> class EAPRampEOS {
   
   template<typename F>
   PORTABLE_FUNCTION
-  auto alpha_impl(F&& func, const Real rho, const Real arg2,
+  Real alpha_impl(F&& func, const Real rho, const Real arg2,
 		  Real* lambda = nullptr) const {
     constexpr const Real a_eps {1.e-12};
     const auto f = [&] (const Real Pl) {
@@ -472,11 +472,13 @@ template <typename T> class EAPRampEOS {
       return Pl - func(a*rho, arg2, lambda) / a;
     };
     Real a {0.0};
+    // using the new solver doesn't allow us to extract the pressure
+    // at the correct alpha. This will result in an extra eos call when
+    // the pressure is needed.
     RootFinding1D::RootCounts rc {};
     RootFinding1D::findRoot(f, 0.0, alpha0_, alpha0_/100.0, alpha0_*100.0,
 			    a_eps, a_eps, a, rc);
-    const Real P {func(a*rho, arg2, lambda) / a};
-    return std::tie(a, P);
+    return a;
   }
 
   // calculate alpha given a rho-sie
@@ -519,8 +521,8 @@ template <typename T> class EAPRampEOS {
   PORTABLE_FUNCTION
   Real PressureFromDensityInternalEnergy(const Real rho, const Real sie,
                                          Real *lambda = nullptr) const {
-    return std::get<1>(this->alpha_impl(&t_.PressureFromDensityInternalEnergy,
-					rho, sie, lambda));
+    const Real alpha {this->alpha_rho_sie(rho, sie, lambda)};
+    return t_.PressureFromDensityInternalEnergy(alpha*rho, sie, lambda) / alpha;
   }
 
   PORTABLE_FUNCTION
@@ -547,8 +549,9 @@ template <typename T> class EAPRampEOS {
   PORTABLE_FUNCTION
   Real PressureFromDensityTemperature(const Real rho, const Real temperature,
                                       Real *lambda = nullptr) const {
-    return std::get<1>(this->alpha_impl(&t_.PressureFromDensityTemperature,
-					rho, temperature, lambda));
+    const Real alpha {this->alpha_rho_temperature(rho, temperature, lambda)};
+    return
+      t_.PressureFromDensityTemperature(alpha*rho, temperature, lambda) / alpha;
   }
 
   PORTABLE_FUNCTION
--- a/eos/eos.hpp
+++ b/eos/eos.hpp
@@ -462,16 +462,17 @@ template <typename T> class EAPRampEOS {
     }
   }
   
-  template<typename F, typename ... ARGS>
+  template<typename F>
   PORTABLE_FUNCTION
-  auto alpha_impl(F func, const Real rho, ARGS... args) const {
+  auto alpha_impl(F&& func, const Real rho, const Real arg2,
+		  Real* lambda = nullptr) const {
     constexpr const Real a_eps {1.e-12};
     Real a {0.0};
     Real a_old {alpha0_};
     Real diff {0.0};
     Real P {0.0};
     for (int i {0}; i < 10; ++i) {
-      P = *func(a_old*rho, args...) / a_old;
+      P = func(a_old*rho, arg2, lambda) / a_old;
       a = this->alpha_p(P);
       diff = fabs(a - a_old) / fabs(a + a_old) * 2.0;
       if (diff < a_eps) break;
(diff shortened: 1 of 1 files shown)


eos/eos.hpp
```diff
@@ -448,6 +448,137 @@ template <typename T> class RelativisticEOS {
...
+    Real diff {0.0};
+    Real P {0.0};
+    for (int i {0}; i < 10; ++i) {
+      P = *func(a_old*rho, args...) / a_old;
+      a = this->alpha_p(P);
+      diff = fabs(a - a_old) / fabs(a + a_old) * 2.0;
+      if (diff < a_eps) break;
+      a_old = a;
+    }
+    return std::tie(a, P);
```
Yurlungur, line 480: I think we should move away from from fixed point iteration quickly. It's slow and not trustworthy. I'd rather use the root finding utilities we already developed.

eos/eos.hpp
```diff
@@ -448,6 +448,137 @@ template <typename T> class RelativisticEOS {
...
+    } else if (P > Pc_) {
+      return 1.0;
+    } else {
+      return alpha0_ + (P - Pe_) / (Pc_ - Pe_) * (1.0 - alpha0_);
+    }
+  }
+  
+  template<typename F, typename ... ARGS>
+  PORTABLE_FUNCTION
+  auto alpha_impl(F func, const Real rho, ARGS... args) const {
```
Yurlungur, line 467: I think you dropped this `&&`

eos/eos.hpp
```diff
@@ -448,6 +448,137 @@ template <typename T> class RelativisticEOS {
...
+    constexpr const Real a_eps {1.e-12};
+    Real a {0.0};
+    Real a_old {alpha0_};
+    Real diff {0.0};
+    Real P {0.0};
+    for (int i {0}; i < 10; ++i) {
+      P = *func(a_old*rho, args...) / a_old;
```
Yurlungur, line 474: I think you dropped this `std::forward<Args>(args)...`

eos/eos.hpp
Yurlungur, line 480: Is returning a `std::pair` the way to do things these days? Not pass in a reference?

eos/eos.hpp
```diff
@@ -448,6 +448,137 @@ template <typename T> class RelativisticEOS {
...
+                                            const Real temperature,
+                                            Real *lambda = nullptr) const {
+    const Real alpha {this->alpha_rho_temperature(rho, temperature, lambda)};
+    return
+      t_.GruneisenParamFromDensityTemperature(alpha*rho, temperature, lambda);
+  }
+
+private:
+  T t_;
+  Real Pe_, Pc_, de_pc_, de_pe_, P0_, T0_, r0_, alpha0_;
```
Yurlungur, line 578: Do some of these quantities change with space? If so, should we build them into `lambda` now rather than wait?

Yurlungur (commented):
Some comments below. I'd add that I think when we implement `FillEos` we should think about it carefully. I imagine we can get something is more performant if you use it, as opposed to the per-var calls, to generate a bunch of quantities at once.

dholladay00:
> Some comments below. I'd add that I think when we implement `FillEos` we should think about it carefully. I imagine we can get something is more performant if you use it, as opposed to the per-var calls, to generate a bunch of quantities at once.

I think the most straightforward way to do this is via lambda. Is the quantity given by the function `nlambda()` a compile time constant, if so the offsets (pmax_index = nlambda of the underlying eos, alpha index = pmax_index +1, etc.). This seems like the most straightforward way to do things, but there is a problem that it will at best reduce performance by 50% (there will be one pressure evaluation + the other evaluations necessary). However, that loss should only be in the pre-crush phase for irreversible Eos's and I am hoping to put in machinery to avoid all extra calls after the crush phase. @Yurlungur I'm happy to think through a better design with you but passing through lambda seems like a decent option that is not overly burdensome to implement.

eos/eos.hpp
dholladay00: @Yurlungur will we always have access to the solver machinery? Perhaps using the machinery if spiner is available and use this fixed point (or some other easy to implement solve) when spiner isn't available via some `#ifdef`s. I should note that my initial implementation had duplicated solver stuff all over the code, so I was pretty pleased with myself for this one, though this has yet to be fully compiled, I should probably instantiate a concrete object at some point to make sure this actually works lol.

dholladay00: I think `pair` supports `std::get<N>` so that is possible, but an auto function signature, a tie, and get was the lazy way and I don't think it misses any potential optimizations or copy elisions or any such thing and I didn't have to think about any concrete types.

eos/eos.hpp
dholladay00: My understanding, albeit poor, is that these quantities are fixed for a given material. The other issue is that the necessary parameters are going to be P-alpha model dependent. There have been discussions about how extensible the ramp implementation should be and those are up for debate and I'd be happy to discuss further.

eos/eos.hpp
Yurlungur: @dholladay00 The solver machinery is in `singularity-eos/utils` always. It's based in `singularity-eos`, so I think it's safe to assume it's always available. Will fixed point always work?

Yurlungur: yeah ok. I'm still stuck in old C++-land, pre move operator, where this is a no-no. :+1:

eos/eos.hpp
Yurlungur: I was just confused... I thought there was some spatially dependent stuff? Is that not the case?

Yurlungur:
> I think the most straightforward way to do this is via lambda. Is the quantity given by the function `nlambda()` a compile time constant, if so the offsets (pmax_index = nlambda of the underlying eos, alpha index = pmax_index +1, etc.). This seems like the most straightforward way to do things, but there is a problem that it will at best reduce performance by 50% (there will be one pressure evaluation + the other evaluations necessary). However, that loss should only be in the pre-crush phase for irreversible Eos's and I am hoping to put in machinery to avoid all extra calls after the crush phase. @Yurlungur I'm happy to think through a better design with you but passing through lambda seems like a decent option that is not overly burdensome to implement.

I like the plan of hiding alpha initial guesses in the lambda, and iterating nlambda the way you describe. But I'm not sure I understand why this should incur a performance hit. I would imagine that would be a significant performance *improvement*. I was more thinking about computing lambda once at the beginning of `FillEos` and then not again. Can't we do that?

I must be missing something.

dholladay00:
> > I think the most straightforward way to do this is via lambda. Is the quantity given by the function `nlambda()` a compile time constant, if so the offsets (pmax_index = nlambda of the underlying eos, alpha index = pmax_index +1, etc.). This seems like the most straightforward way to do things, but there is a problem that it will at best reduce performance by 50% (there will be one pressure evaluation + the other evaluations necessary). However, that loss should only be in the pre-crush phase for irreversible Eos's and I am hoping to put in machinery to avoid all extra calls after the crush phase. @Yurlungur I'm happy to think through a better design with you but passing through lambda seems like a decent option that is not overly burdensome to implement.
> 
> I like the plan of hiding alpha initial guesses in the lambda, and iterating nlambda the way you describe. But I'm not sure I understand why this should incur a performance hit. I would imagine that would be a significant performance _improvement_. I was more thinking about computing lambda once at the beginning of `FillEos` and then not again. Can't we do that?
> 
> I must be missing something.

@Yurlungur I think our comparison baselines are different. Yes, adding the alpha into lambda will be better than not doing so. However, my comparison was against the raw underlying eos call. Suppose you wanted the specific heat and you passed in an alpha that was correct. The solver would have to call the pressure at least once to verify that the passed in alpha was correct before moving to the specific heat call. That means for each call, there is going to be at least 2 eos calls whereas for the underlying eos, there is 1 eos call, that is how I get my factor of 2. Perhaps there is some way to communicate to the underlying eos calls that alpha is converged and skip the extra calls. Again, I think for most ramps, this won't have a huge performance effect across the entirety of the calculation since ramps are typically somewhat short lived.

eos/eos.hpp
dholladay00: I added it, and I'm using lambdas now instead of function pointers, I think this should be easier for the compiler to inline and optimize, plus function pointers are gross.

eos/eos.hpp
dholladay00: No more variadic template. Code is now less template-y and more understandable.

eos/eos.hpp
dholladay00: The spatial varying values are going to be passed in via lambda, but these shouldn't be. We can discuss more offline if needed.

Yurlungur:
> @Yurlungur I think our comparison baselines are different. Yes, adding the alpha into lambda will be better than not doing so. However, my comparison was against the raw underlying eos call. Suppose you wanted the specific heat and you passed in an alpha that was correct. The solver would have to call the pressure at least once to verify that the passed in alpha was correct before moving to the specific heat call. That means for each call, there is going to be at least 2 eos calls whereas for the underlying eos, there is 1 eos call, that is how I get my factor of 2. Perhaps there is some way to communicate to the underlying eos calls that alpha is converged and skip the extra calls. Again, I think for most ramps, this won't have a huge performance effect across the entirety of the calculation since ramps are typically somewhat short lived.

AH, yes. This is true. This is why I'm so focused on `FillEos`. I think you can amortize the cost somewhat there. i.e., if you want specific heat (say) Pressure, you can compute alpha first, so you need 3 EOS calls instead of 4 for the naive approach or 2 for the no-ramp case (assuming you pass in the right alpha in the lambda).

jhp-lanl:
I still don't fully understand the Menikoff and Kober approach to energy but it will throw a wrench in things a bit. I agree with @chadmeyer 's comment that we shouldn't emulate other approaches for the sake of emulation but try to have the right approach here and hopefully it can get some proper V&V in the future.

In that spirit, let me think a bit about how to deal with the energy and I may be able to propose some small changes. For the moment, just know that the energy may need to be a function of alpha and perhaps a compaction potential parameter.

chadmeyer:
I *think* the energy modification is just a function of `alpha` (An integral of the potential maybe?)  I’d have to check for sure.

eos/eos.hpp
dholladay00: We are now using this, one notable downside is that using the solver I'd need to create a custom functor to obtain the pressure at the correct alpha. At the moment I'm just using a lambda and an extra pressure call.

Yurlungur:
Bits have rotted. We're closing this PR for now.

Yurlungur:
Do not delete branch---still needed as a reference.

Yurlungur:
@dholladay00 do we still need this branch? Probably not right?

dholladay00:
I'd still like to have it. This shows how to implement a p-alpha model so I'd like to keep this somewhere so that I don't have re-understand the details.

PR #17, opened 2021-06-30
first pass at reorganziing singularity-eos

Yurlungur:
As discussed in the call with @mauneyc-LANL @dholladay00 and @jdolence reorganize the file structure to use "absolute" paths for includes, and to match better with singularity-eos.

We tested briefly on my VM but more testing on different hardware, and with the downstream codes is required to ensure that, e.g., in-tree builds still work.


Yurlungur:
@dholladay00 @mauneyc-LANL I just pushed a fix for install. But there is a gross aspect, where we also install everything in the `utils` directory. This might cause problems unless we do some curation or quarantining, by putting them in a subdirectory like `singularity-utils` or something. Interested in your thoughts.

Here's the install fix:
```
install(DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/singularity-eos"
  DESTINATION "include"
  FILES_MATCHING PATTERN "*.hpp"
  )

# TODO(JMM): Too aggressive? This will install Eigen, json and kokkos
# headers into the destination.
# I'm leaving this for now, but we should have a longer discussion
# about this.
install(DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/utils/"
  DESTINATION "include"
  FILES_MATCHING PATTERN "*.hpp"
  )
```

dholladay00:
I think that's too aggressive yes. I think we should be progressively building up a list headers based on build options, but I also complicate some of that b/c when I generate a tar file, many of the directories that are in the utils directory by default aren't there, so there will have to be guards like `if utils/kokkos exists and kokkos-intree, then ...` type of stuff.

Yurlungur:
> I think that's too aggressive yes. I think we should be progressively building up a list headers based on build options, but I also complicate some of that b/c when I generate a tar file, many of the directories that are in the utils directory by default won't exist in the tar'd version of the code, so there will have to be guards like `if utils/kokkos exists and kokkos-intree, then ...` type of stuff.

That seems reasonable. SInce you know what this needs to look like do you want to give that a shot and just push to this branch @dholladay00 ? I can also poke at it. This may also get easier if we remove some of these dependencies from in-tree, as we've been discussing.

Yurlungur:
@dholladay00 take a look now. I now try to iteratively add things and more carefully install utils.

I also tested several configurations, which all work, but not yet in-tree builds.

dholladay00:
> @dholladay00 take a look now. I now try to iteratively add things and more carefully install utils.
> 
> I also tested several configurations, which all work, but not yet in-tree builds.

@Yurlungur I really like this solution, what I was thinking was redoing that entire if structure again (unnecessarily). This looks good, I need to test with a tar'd version and make sure it still installs. This is way better than the initial solution and at first glance at least appears like it should work.

Yurlungur:
I can confirm that this reorg works for in-tree builds for `Phoebus`. I didn't check for `riot`, yet. That said, I think from my point of view this is actually good to go now.

dholladay00:
@Yurlungur it appears that the installation of fortran `.mod` files got lost in translation here, could we add that back?

Yurlungur:
I'll look at it when I get back next week. You can also feel free to push your own fix if you get to it first.

Yurlungur:
@dholladay00 try now. I think I fixed the issue with mod files.

singularity-eos/base/eos_error.hpp
```diff
@@ -0,0 +1,26 @@
...
+// reserved by Triad National Security, LLC, and the U.S. Department of
+// Energy/National Nuclear Security Administration. The Government is
+// granted for itself and others acting on its behalf a nonexclusive,
+// paid-up, irrevocable worldwide license in this material to reproduce,
+// prepare derivative works, distribute copies to the public, perform
+// publicly and display publicly, and to permit others to do so.
+//------------------------------------------------------------------------------
+
+#ifndef SINGULARITY_EOS_BASE_EOS_ERROR_HPP_
+#define SINGULARITY_EOS_BASE_EOS_ERROR_HPP_
```
dholladay00, line 16: This appears to follow a convention of `SINGULARITY_EOS_<FILEPATH>_<HEADERNAME>`. I think we should make this consistent across all of our header files (mixed_cell_models.hpp and eos_builder.hpp are examples where this is not consistently applied).

Yurlungur: Yes, I agree. I'll try to go through and check these.

dholladay00:
> @dholladay00 try now. I think I fixed the issue with mod files.

I can confirm the latest changes fix the mod file installation.

singularity-eos/base/eos_error.hpp
Yurlungur: I think I've fixed this everywhere. I added a leading underscore too.

Yurlungur:
@dholladay00 I think I've addressed all concerns. Can you take another look and approve if you're happy with it?

utils/fast-math/logs.hpp
```diff
@@ -17,15 +17,15 @@
 #ifndef _FAST_MATH_LOGS_
```
dholladay00, line 17: @Yurlungur do we want this file to follow the convention as well?

singularity-eos/eos/singularity_eos.hpp
```diff
@@ -15,8 +15,8 @@
 #ifndef EOS_SINGULARITY_EOS_HPP_
 #define EOS_SINGULARITY_EOS_HPP_
```
dholladay00, line 16: missed one

Yurlungur: Fixed

utils/CMakeLists.txt
```diff
@@ -1,3 +1,4 @@
+
```
dholladay00, line 1: Is this space standard?

Yurlungur: That's definitely a typo. Will remove.

dholladay00 (approved):
Looks like everything has been addressed. I'm approving.

PR #19, opened 2021-07-19
Add eospac skip extrapoaltion checks option

annapiegra:
Adds Eospac option to skip internal extrpolation checks. Requires eospac version 6.5 or higher.


Yurlungur (approved):
Thanks, @annapiegra! This is a nice improvement! It looks to me like perhaps this could be set at run time via the json parser, or via a CLI flag. But for now I think the build time option is fine.

PR #20, opened 2021-07-27
Adding optimized log implementation using frexp and approximation to …

dholladay00:
…log on [0.5,1) with the help of Jonah.

Create new fast log10 implementation.
- [ ] Discuss keep or drop other optimized log implementation?

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -10,9 +10,6 @@
 [submodule "utils/variant"]
 	path = utils/variant
 	url = https://github.com/mpark/variant.git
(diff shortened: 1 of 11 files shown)


Yurlungur (approved):
This is awesome. I am in favor of just dropping the `fast-log` CPU-only implementation, purging that submodule, and using this, assuming it's acceptably fast.

The one thing I think we might want is a way to use `std::log10` for debugging purposes. What do you think @dholladay00 ?

Yurlungur:
This is good to go for now. I'm merging it.

PR #21, opened 2021-07-29
Miscellaneous fixes

Yurlungur:
Fix issue introduced by last merge that @annapiegra found. Also cleaned up cmake to make it easier to build for GPUs.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -113,7 +113,7 @@ if (SINGULARITY_USE_EOSPAC)
     message(WARNING "EOSPAC enabled without hdf5.")
   endif()
   if (SINGULARITY_USE_CUDA)
-    message(FATAL_ERROR "EOSPAC not supported with cuda")
+    message(WARNING "EOSPAC not supported with cuda. You may encounter errors.")
   endif()
 endif()
 
@@ -290,7 +290,7 @@ if (SINGULARITY_USE_SINGLE_LOGS)
       SINGULARITY_USE_SINGLE_LOGS)
 endif()
 if (SINGULARITY_FMATH_USE_ORDER_4 AND SINGULARITY_FMATH_USE_ORDER_5)
-  message(ERROR "Order 4 and order 5 interpolation both specified. Please specify only one or zero. If no specification is made, order 7 will be used.")
+  message(FATAL_ERROR "Order 4 and order 5 interpolation both specified. Please specify only one or zero. If no specification is made, order 7 will be used.")
 endif()
 if (SINGULARITY_FMATH_USE_ORDER_4)
   target_compile_definitions(singularity-eos::flags INTERFACE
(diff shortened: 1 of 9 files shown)


CMakeLists.txt
```diff
@@ -328,6 +328,11 @@ if (SINGULARITY_BUILD_TESTS)
 endif()
 # create the libsingularity-eos.a library
 if(SINGULARITY_USE_FORTRAN)
+  if (NOT SINGULARITY_BUILD_CLOSURE)
+    message(FATAL_ERROR "The fortran library relies on closures. "
```
dholladay00, line 332: I'd say if we get here to just turn on build closure and warn that we are doing so.

Yurlungur: I'd rather be explicit. This came up because I wanted to build `compare_to_eospac` for GPU. But I didn't want to bother linking in kokkos kernels. So I want to get an error message that tells me what to use for that scenario.

Yurlungur: I'm willing to change it though if you feel strongly.

sesame2spiner/io_eospac.hpp
```diff
@@ -75,10 +75,8 @@ class Bounds {
       // should be single-precision epsilon b/c that's whats used for the logs
       constexpr Real epsilon = std::numeric_limits<float>::epsilon();
       const Real min_offset = 10*std::abs(epsilon);
-      if (min < 0) offset = std::abs(min) + min_offset;
-      else if ( min == 0 ) {
-        offset = min_offset;
-      }
+      if (min <= 0) offset = 1.1*std::abs(min) + min_offset;
```
dholladay00, line 78: Could I get more context here? Would it be worth commenting in the code how offset is calculated and why?

singularity-eos/eos/eos_spiner.cpp
```diff
@@ -842,7 +842,8 @@ Real SpinerEOSDependsRhoT::lTFromlRhoSie_(const Real lRho,
   if ( whereAmI == TableStatus::OffBottom ) {
     // on the cold curve. No unique temperature.
     // return the temperature at the cold curve crossing.
-    lT = lTColdCrit_.interpToReal(lRho);
+    // lT = lTColdCrit_.interpToReal(lRho);
```
dholladay00, line 845: the comments here probably need to change and we should remove the commented out line of code if this is the final version for the PR.

CMakeLists.txt
dholladay00: It's worth thinking about in the future of not requiring any linear algebra library if we're not building closures, but for now I'll buy it.

Yurlungur: :+1:

sesame2spiner/io_eospac.hpp
Yurlungur: Added a comment. Basically the issue was that the code was choosing an offset for min < 0 to bring min to identically zero, which still doesn't work when you take the log. This situation now handles min == 0 with min_offset and min < 0 with the 1.1 so that you're a little away from identically zero after shifting.

singularity-eos/eos/eos_spiner.cpp
Yurlungur: Done

dholladay00 (approved):

PR #23, opened 2021-08-04
Fix usage of eos_setOption

annapiegra:
eos_SetOption can only set a single option in a call. 
EOS_INSERT_DATA is now set to 4 as was the intention in the original code.


Yurlungur (approved):

Yurlungur:
Thanks for the fix!

PR #24, opened 2021-08-06
This commit includes changes to find gpuoffload eospac libraries, rem…

dholladay00:
…oves some cuda warnings, more faithfully calculate L2 difference norms, appropriately takes into account launch overhead, and chooses more appropriate sizes on GPUs.

### DIFF
--- a/cmake/FindEOSPAC.cmake
+++ b/cmake/FindEOSPAC.cmake
@@ -24,7 +24,7 @@
 FIND_PATH(EOSPAC_INCLUDE_DIR NAMES eos_Interface.h)
 
 # Look for the library.
-FIND_LIBRARY(EOSPAC_LIBRARY NAMES eospac6 libeospac6)
+FIND_LIBRARY(EOSPAC_LIBRARY NAMES eospac6 libeospac6 libeospac6gpu.a)
 
 # handle the QUIETLY and REQUIRED arguments and set EOSPAC_FOUND to TRUE if
 # all listed variables are TRUE
--- a/singularity-eos/base/eos_error.hpp
+++ b/singularity-eos/base/eos_error.hpp
@@ -19,7 +19,7 @@
 #include <stdexcept>
 #define EOS_ERROR(x) (throw std::runtime_error(x))
 #else
-#define EOS_ERROR(x) printf("%s\n", x); exit(1);
+#define EOS_ERROR(x) printf("%s\n", x); assert(false);
 #endif
 #define UNDEFINED_ERROR EOS_ERROR("DEFINE ME\n")
 
--- a/test/compare_to_eospac.cpp
+++ b/test/compare_to_eospac.cpp
@@ -81,7 +81,7 @@ int main(int argc, char* argv[]) {
   constexpr const bool is_host = SpaceAccessibility<HS, KDMS>::accessible;
   constexpr const int def_fine = is_host ? 512 : 2048;
 #else
-  constexpr const int def_fin = 512;
+  constexpr const int def_fine = 512;
 #endif // PORTABILITY_STRATEGY_KOKKOS
   int nFine = argc >= 3 ? std::atoi(argv[2]) : def_fine;
   int nFineRho = nFine + 1;
--- a/test/compare_to_eospac.cpp
+++ b/test/compare_to_eospac.cpp
@@ -74,7 +74,16 @@ int main(int argc, char* argv[]) {
     std::exit(1);
   }
   std::string filename = argv[1];
-  int nFine = argc >= 3 ? std::atoi(argv[2]) : 512;
+#ifdef PORTABILITY_STRATEGY_KOKKOS
+  using Kokkos::SpaceAccessibility;
+  using KDMS = Kokkos::DefaultExecutionSpace::memory_space;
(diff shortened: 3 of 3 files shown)


annapiegra (approved):
While making changes to the test: would it make sense to make matid an input parameter (int matId    = std::atoi(argv[2]); std::vector<int> matids{matId};)?

cmake/FindEOSPAC.cmake
```diff
@@ -24,7 +24,7 @@
 FIND_PATH(EOSPAC_INCLUDE_DIR NAMES eos_Interface.h)
 
 # Look for the library.
-FIND_LIBRARY(EOSPAC_LIBRARY NAMES eospac6 libeospac6)
+FIND_LIBRARY(EOSPAC_LIBRARY NAMES eospac6 libeospac6 libeospac6gpu.a)
```
Yurlungur, line 27: what happens if hte GPU library can't be found?

singularity-eos/base/eos_error.hpp
```diff
@@ -19,7 +19,7 @@
 #include <stdexcept>
 #define EOS_ERROR(x) (throw std::runtime_error(x))
 #else
-#define EOS_ERROR(x) printf("%s\n", x); exit(1);
+#define EOS_ERROR(x) printf("%s\n", x); assert(false);
```
Yurlungur, line 22: :+1:

test/compare_to_eospac.cpp
```diff
@@ -74,7 +74,16 @@ int main(int argc, char* argv[]) {
...
+#ifdef PORTABILITY_STRATEGY_KOKKOS
+  using Kokkos::SpaceAccessibility;
+  using KDMS = Kokkos::DefaultExecutionSpace::memory_space;
+  using HS = Kokkos::HostSpace;
+  constexpr const bool is_host = SpaceAccessibility<HS, KDMS>::accessible;
+  constexpr const int def_fine = is_host ? 512 : 2048;
+#else
+  constexpr const int def_fine = 512;
+#endif // PORTABILITY_STRATEGY_KOKKOS
+  int nFine = argc >= 3 ? std::atoi(argv[2]) : def_fine;
```
Yurlungur, line 86: 👍

Yurlungur (approved):
Looks good. Except for the minor question about the GPU build, I'm happy to merge this in. Thanks for the fixes!

Yurlungur:
> While making changes to the test: would it make sense to make matid an input parameter (int matId = std::atoi(argv[2]); std::vector matids{matId};)?

Originally, I intended it to be possible to run the script on multiple matids in one call. But I think you're right, it's better to just make `matid` an input parameter and run the script multiple times if that's what is desired.

cmake/FindEOSPAC.cmake
dholladay00: it will attempt to find a library that matches any in that list, so it should be fine. I'll test to make sure.

singularity-eos/base/eos_error.hpp
dholladay00: > > While making changes to the test: would it make sense to make matid an input parameter (int matId = std::atoi(argv[2]); std::vector matids{matId};)?
> 
> Originally, I intended it to be possible to run the script on multiple matids in one call. But I think you're right, it's better to just make `matid` an input parameter and run the script multiple times if that's what is desired.

Just for clarity @Yurlungur, what code change is being requested and should I add it in this PR?

Yurlungur:
@dholladay00 

> Just for clarity @Yurlungur, what code change is being requested and should I add it in this PR?

@annapiegra is suggesting replacing the loop over matids, where the matids are a hardcoded vector, with using a single matid provided by the command line. I think this is a good idea. If you want to add it to this PR, go ahead. But I can also look at it later.

dholladay00:
@Yurlungur, is it possible to query the spiner file for all matid's and by default loop over them? I'd think one could even optionally include a subset of ID's on the command line and have that respected (nummats = argc - num_other_args or some similar formula and `matid[m] = atoi(argv[offset+m]);` But if we did this I think it best done in a separate PR.

Yurlungur:
> @Yurlungur, is it possible to query the spiner file for all matid's and by default loop over them? I'd think one could even optionally include a subset of ID's on the command line and have that respected (nummats = argc - num_other_args or some similar formula and `matid[m] = atoi(argv[offset+m]);` But if we did this I think it best done in a separate PR.



> @Yurlungur, is it possible to query the spiner file for all matid's and by default loop over them? I'd think one could even optionally include a subset of ID's on the command line and have that respected (nummats = argc - num_other_args or some similar formula and `matid[m] = atoi(argv[offset+m]);` But if we did this I think it best done in a separate PR.

Yes this is possible @dholladay00. One would loop through the hdf5 groups that are numbers in the file. We can treat this in a separate PR.

cmake/FindEOSPAC.cmake
dholladay00: I confirmed this is the case @Yurlungur

singularity-eos/base/eos_error.hpp
dholladay00: This task will be done in a different PR.

PR #25, opened 2021-08-18
[WIP]: Remove json dependency + remove matid hardcoding in compare_to_eospac

Yurlungur:
Over `singularity-eos`'s life, we've had issues building with `json` on some platforms. In particular, it doesn't cooperate with `nvcc`, making it unusable for cuda builds, and it's not available through spack, forcing us to carry it around as a submodule for the foreseeable future, even with @mauneyc-LANL 's spack work.

Motivated by that, this PR is a potential path forward, where I have removed the `json` dependency completely. `json` is only used in `sesame2spiner` where it defines the input file format. However, it's far richer than we need. As such, I have replaced it with a very simple hand-rolled parser I threw together.  The new version doesn't have input blocks or lists, so instead I make one input file per material. Thus, calls to sesame2spiner now look like this:

```bash
./sesame2spiner/sesame2spiner materials.sp5 ../sesame2spiner/examples/unit_tests/*.dat
```
with all the usual optional arguments provided previously. Now the first sequential argument names the output file name, and the remaining files are input files, one per material. An input file might look like this:
```
matid = 4272
name = steel
rhomin = 1e-2
Tmin = 1
# These shrink logarithm of bounds
# by a fraction of the total interval <= 1
shrinklRhoBounds = 0.15,
shrinklTBounds = 0.15,
shrinkleBounds = 0.5
``` 
so essentially the same format as before.

I also went in to `compare_to_eospac.cpp` and added better CLI parsing. So now matids are specified via the CLI:
```bash
./compare_to_eospac.cpp -n 512 -f materials.sp5 5030 4272
```
where `-n` specifies the `nFine` variable and `-f` specifies the filename (default materials.sp5). The sequential arguments are matids.

This is an API breaking change, albeit only to `sesame2spiner` so I want to deliberate carefully before merging this in. Hence the WIP flag. (The PR is ready otherwise.) Interested parties, please take a look and let me know what you think. In particular pinging @dholladay00 @mauneyc-LANL based on our previous discussion.

@jdolence @brryan and @chadmeyer may be interested in the overall design. @annapiegra may be interested especially in the changes to `compare_to_eospac`.


sesame2spiner/parser.cpp
```diff
@@ -12,103 +10,63 @@
...
-  std::ifstream file(filename.c_str());
-  std::stringstream buffer;
-  if (!file.good()) {
-    std::cerr << "Cannot read file " << filename << std::endl;
-    std::exit(1);
-  }
-  buffer << file.rdbuf();
-  return stringToJson(buffer.str());
+template <> std::string Params::Get(const std::string &key) const {
+  return params_.at(key);
```
brryan, line 48: `unordered_map::at` gives an obscure (at least to me) error if `key` isn't contained. What about adding a check of `Params::Contains` here first, and if the parameter isn't contained, some helpful message like `"Parameter \"" + key + "\" not found!"`?

brryan (approved):
In my opinion dependency removal trumps API-breaking change. Approved (I do think the extra `key` checking I suggested would be useful for many people over time). Although I'm not really involved in actually using this code so others' opinions should be weighted more heavily than mine.

mauneyc-LANL (approved):
Looks fine to me; glad this wasn't too difficult to swap in.

sesame2spiner/parser.cpp
Yurlungur: Yeah that's a good idea.

annapiegra (approved):
Thanks for making compare_to_eospac more user-friendly.

PR #27, opened 2021-08-27
Separate Out EOSPAC functionalit out of sesame2spiner

Yurlungur:
A little bit of cleanup requested by @dholladay00 . Moves the "eospac" wrappers that I wrote for `sesame2spiner` into a separate directory. This lets us use different build/link flags for the eospac-specific stuff, and it also more cleanly separates the pieces of machinery that are used outside `sesame2spiner`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,7 +23,6 @@ option (SINGULARITY_USE_EOSPAC "Pull in eospac" OFF)
 option (SINGULARITY_USE_CUDA "Enable cuda support" OFF)
 option (SINGULARITY_USE_KOKKOSKERNELS
   "Use kokkos-kernels for linear algebra" OFF)
-option (SINGULARITY_INVERT_AT_SETUP "Use eospacs preinverted tables" OFF)
 option (SINGULARITY_BUILD_TESTS "Compile tests" OFF)
 option (SINGULARITY_BUILD_SESAME2SPINER "Compile sesame2spiner" OFF)
 option (SINGULARITY_BUILD_STELLARCOLLAPSE2SPINER "Compile stellarcollapse2spiner" OFF)
@@ -34,7 +33,6 @@ option (SINGULARITY_SUBMODULE_MODE "Submodule mode" OFF)
 option (SINGULARITY_BUILD_CLOSURE "Mixed cell closure" ON)
 option (SINGULARITY_TEST_SESAME "Test the Sesame table readers" OFF)
 option (SINGULARITY_TEST_STELLAR_COLLAPSE "Test the stellar collapse table readers" OFF)
-option (SINGULARITY_EOS_SKIP_EXTRAP "Turn off eospac extrap checks" OFF)
 option (SINGULARITY_USE_SINGLE_LOGS "Use single precision logs. Can harm accuracy." OFF)
 option (SINGULARITY_FMATH_USE_ORDER_4 "4th order interpolant for fast logs. Default is 7th order." OFF)
 option (SINGULARITY_FMATH_USE_ORDER_5 "5th order interpolant for fast logs. Default is 7th order." OFF)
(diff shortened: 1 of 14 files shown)


Yurlungur:
I also changed `INVERT_AT_SETUP` to a runtime parameter, which let's us, e.g., select it for the `EOSPAC` eos class at runtime or set that at runtime in `compare_to_eospac.cpp`.

annapiegra:
Could eospac skip extrapolation checks default to true? I don't think many use the checks and the overhead can be up to 4x. 

How about adding linear interpolation as a runtime option for eospac?

Yurlungur:
> Could eospac skip extrapolation checks default to true? I don't think many use the checks and the overhead can be up to 4x.

The reason I left it as default off is because the latest eospac version on spac is still too early a version to have the checks. Do we know if the spackage will be updated?

> How about adding linear interpolation as a runtime option for eospac?

This is a good idea. Yes, should be straightforward to make that a runtime option. Should we inject any additional points when running in linear interp mode?

annapiegra:
> > Could eospac skip extrapolation checks default to true? I don't think many use the checks and the overhead can be up to 4x.
> 
> The reason I left it as default off is because the latest eospac version on spac is still too early a version to have the checks. Do we know if the spackage will be updated?
> 

Fair enough. 

> > How about adding linear interpolation as a runtime option for eospac?
> 
> This is a good idea. Yes, should be straightforward to make that a runtime option. Should we inject any additional points when running in linear interp mode?

Considering the artifacts that EOS_INSERT_DATA seems to be adding/amplifying I would not add extra points if not inverting. In general, the recommendation is to experiment to find the best settings for  if and how many points should be inserted (optimal number of points can (and will) vary depending on table). I haven't found anything on how inserting data affects linear interpolation results if not inverting (I guess I could run some tests). Depending on how much of eospac you want accessible from singularity adding runtime options for inserting data and number of points would be nice.

Yurlungur:
Hmmm ok. I'll maybe pass that in as a runtime option too then

jhp-lanl:
> Considering the artifacts that EOS_INSERT_DATA seems to be adding/amplifying I would not add extra points if not inverting. In general, the recommendation is to experiment to find the best settings for if and how many points should be inserted (optimal number of points can (and will) vary depending on table). **I haven't found anything on how inserting data affects linear interpolation results if not inverting (I guess I could run some tests)**. Depending on how much of eospac you want accessible from singularity adding runtime options for inserting data and number of points would be nice.

FYI - The document LA-UR-19-23881 is work by a former summer student of mine and Kevin Honnell's that looked specifically at how `EOS_INSERT_DATA` + linear interpolation fares for a simple single-material problem. The simulations were performed in xRAGE. She looked at the root mean square error (so I guess an L2 norm?) of the error over the entire simulation at its conclusion to determine accuracy and we used the timing information from within xRAGE to benchmark. 

Short story: the runtime of EOSPAC was roughly halved using linear interpolation and was not affected by the table density. You could get to within around 2% accuracy of the rational function interpolator with around 3x table density an linear interpolation.

jhp-lanl:
> Hmmm ok. I'll maybe pass that in as a runtime option too then

Is this just for populating the spiner tables or for also running EOSPAC in a host code? If the latter then yes, I think it's important but if the former then I think rational interpolation is fine there by itself.

annapiegra:
> > Hmmm ok. I'll maybe pass that in as a runtime option too then
> 
> Is this just for populating the spiner tables or for also running EOSPAC in a host code? If the latter then yes, I think it's important but if the former then I think rational interpolation is fine there by itself.

Latter (at least that's what I'd like to see)

annapiegra:
> Short story: the runtime of EOSPAC was roughly halved using linear interpolation and was not affected by the table density. You could get to within around 2% accuracy of the rational function interpolator with around 3x table density an linear interpolation.

2% or 0.02%?

jhp-lanl:
> > Short story: the runtime of EOSPAC was roughly halved using linear interpolation and was not affected by the table density. You could get to within around 2% accuracy of the rational function interpolator with around 3x table density an linear interpolation.
> 
> 2% or 0.02%?

Sorry, I think you're right, I misread her plots. The differences were on the order of 0.02%. She has a nice plot of where the differences were in the simulation and these look consistent with an RMS on that order with local spikes up to around order 0.1 or 1% I think.

Yurlungur:
> Latter (at least that's what I'd like to see)

This is what I meant yes.

dholladay00:
What else is required for this to be merged @Yurlungur ?

dholladay00 (approved):

PR #29, opened 2021-09-23
UnitSystem modifier

Yurlungur:
As suggested/requested by @brryan I implement here a `UnitSystem` modifier. The idea with this modifier is that the host code can work in whatever units it likes, and if it initializes an eos with the `UnitSystem` modifier, quantities are changed into cgs for eos calls, and then back out of cgs upon return. 

## Creation via constructor

The `UnitSystem` modifier can be created either using a set of three thermodynamic quantities:
- `rho_unit`, the unit for density
- `sie_unit`, the unit for specific internal energy
- `temp_unit`, the unit for temperature

or via four "base" units:
- `time_unit`
- `mass_unit`
- `length_unit`
- `temp_unit`

The constructor distinguishes between these with tag dispatch. The tags are
- `thermal_units_init_tag`
- `length_time_units_init_tag`

The default is thermal units initialization. 

## Creation via `EOSBuilder`

The `EOSBuilder` also supports the modifier. The names are the same, but instead of tag dispatch, specify `use_length_time` in the params. For example:
```C++
EOSBuilder::EOSType type = EOSBuilder::EOSType::IdealGas;
EOSBuilder::modifiers_t modifiers;
EOSBuilder::params_t base_params, units_params;
base_params["Cv"].emplace<Real>(Cv);
base_params["gm1"].emplace<Real>(gm1);
units_params["use_length_time"].emplace<bool>(true);
units_params["time_unit"].emplace<Real>(time_unit);
units_params["length_unit"].emplace<Real>(length_unit);
units_params["mass_unit"].emplace<Real>(mass_unit);
units_params["temp_unit"].emplace<Real>(temp_unit);
modifiers[EOSBuilder::EOSModifier::UnitSystem] = units_params;
EOS eos = EOSBuilder::buildEOS(type, base_params, modifiers);
```
the default value for `use_length_time` is `false`.

## Other unit quantities

Internally, the modifier takes these base units and constructs the others self-consistently.

## Availability

Currently I have only enabled the modifier for `IdealGas`, `SpinerEOS`, and `StellarCollapse`. Others can be enabled as needed. Also it is currently incompatible with any other modifiers... but of course this too can be changed.

## Code reorganization

To help clean up `eos.hpp`, I also move the modifiers into their own separate header files. pinging @mauneyc-LANL to let him know.

Pinging mostly @brryan, @dholladay00 and @jdolence for review. Also pinging @lroberts36 in case you're interested in seeing what the microphysics machinery in Phoebus looks like.


singularity-eos/eos/eos_builder.cpp
```diff
@@ -28,11 +28,15 @@ EOS EOSBuilder::buildEOS(EOSBuilder::EOSType type,
...
+  if ((shifted || scaled || relativistic || units) && !(isModifiable(type))) {
     EOS_ERROR("Modifiers not supported for this EOS");
   }
-  if (relativistic && (shifted || scaled)) {
-    EOS_ERROR("Relativistic modifier cannot be combined with shift or scale");
+  if (relativistic && (shifted || scaled || units)) {
+    EOS_ERROR("Relativistic modifier currently must be applied alone");
+  }
+  if (units && (shifted || scaled || relativistic)) {
+    EOS_ERROR("Unit System modifier currently must be applied alone");
```
brryan, line 39: What would need to change to combine the unit system modifier with other modifiers? Units are only modifying input and output values, right, so I don't see how it would interact with other modifiers (I'm not a singularity modifiers expert though)

singularity-eos/eos/eos_unitsystem.hpp
```diff
@@ -0,0 +1,250 @@
...
+  // move semantics ensures dynamic memory comes along for the ride
+  // TODO(JMM): Entropy unit needed?
+  UnitSystem(T &&t, eos_units_init::ThermalUnitsInit,
+	     const Real rho_unit, const Real sie_unit, const Real temp_unit)
+    : t_(std::forward<T>(t))
+    , rho_unit_(rho_unit)
+    , sie_unit_(sie_unit)
+    , temp_unit_(temp_unit)
+    , press_unit_(rho_unit*sie_unit)
+    , inv_rho_unit_(1/rho_unit) // inverses computed to avoid division at runtime
```
brryan, line 50: is it better to save the memory rather than doing multiplication instead of division? Maybe obviously yes, just not clear to me how UnitSystem moves through device memory

singularity-eos/eos/eos_unitsystem.hpp
```diff
@@ -0,0 +1,250 @@
...
+    const Real gm1 = t_.GruneisenParamFromDensityTemperature(rho*rho_unit_,
+							     temp*temp_unit_,
+							     lambda);
+    return inv_gruneisen_unit_*gm1;
+  }
+
+  PORTABLE_FUNCTION
+  void DensityEnergyFromPressureTemperature(const Real press, const Real temp,
+					    Real *lambda, Real &rho, Real &sie) const {
+    t_.DensityEnergyFromPRessureTemperature(press*press_unit_, temp*temp_unit_, lambda,
```
brryan, line 167: `PRessure` typo?

singularity-eos/eos/shifted_eos.hpp
```diff
@@ -0,0 +1,156 @@
+//------------------------------------------------------------------------------
```
brryan, line 1: What about a separate subdir for modifiers? Since it's not necessarily obvious that e.g. `eos_unitsystem.hpp` and `shifted_eos.hpp` are both modifiers

test/eos_unit_tests.cpp
```diff
@@ -187,6 +187,59 @@ SCENARIO("Relativistic EOS", "[EOSBuilder][RelativisticEOS]") {
...
+    constexpr Real gm1 = 0.5;
+    EOSBuilder::EOSType type = EOSBuilder::EOSType::IdealGas;
+    EOSBuilder::modifiers_t modifiers;
+    EOSBuilder::params_t base_params, units_params;
+    base_params["Cv"].emplace<Real>(Cv);
+    base_params["gm1"].emplace<Real>(gm1);
+    GIVEN("Units with a thermal unit system") {
+      constexpr Real rho_unit = 1e1;
+      constexpr Real sie_unit = 1e-1;
+      constexpr Real temp_unit = 1;
```
brryan, line 202: `1` probably isn't a great value for testing unit conversions (same for next test case), I would actually opt for oddball numbers with a few digits after the decimal point, small point though

brryan (approved):
LGTM! Thanks for turning this around so fast. It would be great to get this in since we can't use tabular eos with phoebus until we have this feature.

singularity-eos/eos/eos_unitsystem.hpp
Yurlungur: I have no idea. It did end up being a fair amount of memory didn't it? Let's leave it for now, but it's worth experimenting.

singularity-eos/eos/eos_unitsystem.hpp
Yurlungur: good catch.

singularity-eos/eos/shifted_eos.hpp
Yurlungur: Yeah this makes sense to me.

test/eos_unit_tests.cpp
Yurlungur: :+1:

singularity-eos/eos/eos_builder.cpp
```diff
@@ -65,6 +110,14 @@ EOS EOSBuilder::buildEOS(EOSBuilder::EOSType type,
       int matid = mpark::get<int>(base_params["matid"]);
       if (type == EOSType::SpinerEOSDependsRhoT) {
         SpinerEOSDependsRhoT s(filename,matid,reproducibility_mode);
+	if (units) {
```
jhp-lanl, line 113: Could you fix the indentation to match the control flow? These `if (units) {...}` blocks are in the scope of the `if (type == EOSType::SpinerEOSDependsRhoT  || type == EOSType::SpinerEOSDependsRhoSie) {` block but the indentation makes them look to be in a different scope.

Yurlungur: oof yes good catch. Not sure what's going on with the indentation here.

jhp-lanl: Hmmmmm... I just took a closer look and the indentation is all over the place in this file. It may be a harder fix.

I'm just fixing it on my end for now and maybe I can make a separate MR since it touches just about every line in the file and I don't want to pollute the real changes here.

jhp-lanl: That works too! My editor did a reasonable job at trying to fix the indentation but it still required some hand fixing. I'm all in favor of automation :-)

Yurlungur: Will do. I'll submit a separate PR from develop for that. Then this one will look cleaner too.

singularity-eos/eos/modifiers/eos_unitsystem.hpp
```diff
@@ -0,0 +1,250 @@
...
+    , inv_sie_unit_(1/sie_unit)
+    , inv_temp_unit_(1/temp_unit)
+    , inv_press_unit_(1/press_unit_)
+    , inv_dpde_unit_(sie_unit/press_unit_) // thermo derivatives computed consistently
+    , inv_dvdt_unit_(rho_unit*temp_unit)  // TODO(JMM): Is this convention weird?
+    , inv_dpdr_unit_(rho_unit/press_unit_)
+    , inv_dtdr_unit_(rho_unit/temp_unit)
+    , inv_dtde_unit_(sie_unit/temp_unit) // obviously this is also Cv
+    , inv_cv_unit_(temp_unit/sie_unit)
+    , inv_gruneisen_unit_(rho_unit*sie_unit/press_unit_)
```
jhp-lanl, line 60: Since `press_unit_` is defined as `rho_unit * sie_unit`, this will always evaluate to 1. Which makes sense since the Gruneisen parameter is unitless. Up to you whether you feel it necessary to explicitly calculate this ;-)

singularity-eos/eos/eos_builder.cpp
Yurlungur: @jhp-lanl I merged in the clang-format. Diff should look much cleaner now.

singularity-eos/eos/modifiers/eos_unitsystem.hpp
Yurlungur: Oh good catch. Initially I allowed `press_unit_` to be inconsistent with `rho_unit` and `sie_unit`. I think I should remove this to save memory. :)

singularity-eos/eos/modifiers/eos_unitsystem.hpp
```diff
@@ -0,0 +1,250 @@
...
+    , inv_bmod_unit_(1/press_unit_)
+  {}
+  UnitSystem(T &&t, eos_units_init::LengthTimeUnitsInit,
+	     const Real time_unit, const Real mass_unit, const Real length_unit,
+	     const Real temp_unit)
+    : UnitSystem(std::forward<T>(t), eos_units_init::thermal_units_init_tag,
+		 mass_unit / length_unit*length_unit*length_unit,
```
jhp-lanl: This is supposed to be units of density, yes? Are you missing parentheses here?

Yurlungur: wow oops. Yes good catch. :)

jhp-lanl:
Looks good to me

Yurlungur:
Thanks for the reviews, @brryan @jhp-lanl ! Merging it in.

PR #22, opened 2021-08-04
WIP: CMake build refactor

mauneyc-LANL:
NOTE: More work needed on corner cases, please test!

Rework of the CMake build toolchain for `singularity-eos`. Some changes from `singularity-opac` carried over.

Highlights

- Simplified "all end-environment" or "all submodule" selection; controlled with `SINGULARITY_USE_SUBMODULES`. If on, no environment checks for dependencies are carried out and all submodules are configured. If off, looks for required/requested dependencies and fails if not found. This felt most efficient, but if needed different use-cases can be appended.
- Updated submodule inclusion; the (likely still underdeveloped) goal here is to modernize how submodules are included. For submodules dependencies _where it is possible_, after calling `add_subdirectory`, the CMake `find_package` call should behave as tho the package is present in the environment, e.g. like it was a loaded via a module or spack. This is part motivated the above, so that once it is determined if the project is built using it's own submods, or system packages, the rest of the code can forget about it.
- Modernized cmake `install` options, e.g. used as `find_package(singularity-eos)` (Still needs work).

Todo:

- Kokkos/EOSPAC test cases. I'll be looking for cracks but there's a lot of egde cases I'll prolly miss.
- Clean up flags and options, there's some redundancy we can polish out.

@Yurlungur @dholladay00 @brryan

PR #30, opened 2021-09-23
WIP: cmake build updates (redo)

mauneyc-LANL:
### Overview

Note: this PR is in replacement of #22 .

The goal of this PR is to reorganize the CMake build to be more "modern-cmake" like. Further, making a more structured and modular build code that is able to be deployed in a variety of contexts. Adding new features should be simpler as well.

### What's New

- Addition of `export` functionality; if `singularity-eos` is installed into directory visiable to a users `PATH`, then `find_package(singularity-eos)` is all that is required for the user. As a in-tree "use-case", the `*2spiner` directories are now specified as separate projects that must import the `singularity-eos` target (through `find_package`), rather than simply inheriting it.
- New `add_library_target` function defined in `cmake/library.cmake`. This is borrowed from [flecsi](https://gitlab.lanl.gov/flecsi/flecsi/-/blob/2.1/cmake/library.cmake). It provides for a generic "configure/export/install" functionality for any library target, and can be dropped into other codes
- `install` functionality now operating
- "non-external" submodules (spiner, fast-math, ect) have been promoted to "provisional" CMake targets, standardizing their interface. In most cases this is little more than defining an interface include path on a target. Further, submodules with proper target exporting (`util/eigen`, `util/variant`) now utilize `add_subdirectory`, similar to how in-tree `kokkos` has been. 
- configuration header-file generated. This isn't used yet, but could be useful in various locations. With some logic-smithing, a few CMake cache options could be replaced as header `#defines`.

### What's New (from the prior attempt)

- Directory structure is now (mostly) identical, and no include paths need be altered. Additionally, all prior `-DSINGULARITY` cache vars are available (though some may not have functionality at time of writing)
- Dependencies are now handled specificly, rather than trying to add them with a common interface. There are pros/cons here, but I found it much easier to work on each dependency on its own terms, rather than abstract them all. This can be revisited, but I would prefer reorginzing in-tree dependencies and a more formal construction of the submod/external/internal distinctions.

### What's to be Done

- Testing; right now there should be little to do to get this back to functionality, but I'm conflicted on if this should be redrafted as well
- Work-out on diverse platforms
- Fortran functionality isn't worked out yet, but I don't foresee too much work here.
- Better seperation of public/private attributes. Currently any target that uses `singularity-eos::singularity-eos` brings in everything (flags, definitions, links, ect.) Would be proper to give these some limits.
- Odds'n'ends


cmauney:
To reiterate our discussion this afternoon, we need to classify configuration and build constants into (possibly overlapping) category types that define how they are advanced within and without the `singularity-eos` project:

**NOTE**: I maybe wrong about the instances I use as examples, but the point here is to assume I'm ignorant and to help in clarifying this classification. Also, ideally these categories are to some extinct exclusive, but even with purpose made build systems this is hard to impose.

1. "single-use" defines: What lives only to give `singularity-eos` instructions to _configure_ the build. this means what options  _only_ exist to direct the CMake configure stage. variables of the same use as `-DCMAKE_BUILD_TYPE`, `-DSINGULARTITY_KOKKOS_INTALL_DIR`, & `-DSINGULARITY_HIDE_MORE_WARNINGS`
2. "private, in-project" defines: What constants are used in both _configuration_ and _build_ stages that control which code components to activate and which dependencies are necessary to activate those components. `-DSINGULARITY_USE_EOSPAC`, `-DSINGULARITY_EOS_SKIP_EXTRAP`. These live in `CMakeCache` along with the defs in (1), but may have effects (tho crucially not _values_) that propagate into the higher-order variables. These variables should not be visible to downstream projects.
3. "public, transient" defines that out-of-tree builds need to correctly _configure_ and use their build of `singularity-eos`. These defines will be exported as publicly and will be available to reference downstram; that is, `targen_link_libraries(.... singularity-eos::singularity::eos)` will append these defines into downstream build flags.
4. pure "build" definitions, existing in a pre-generated header file but otherwise unused in external project CMake configuration logic (e.g. `PORTIBILITY_STRATEGY...`, 'USE_FMATH_ORDER....`). Downstream builds may still directly use the vars (e.g. including the header), but what is defined there should not in principal influence anything about how the downstream code is built.

Thus, we have 4 variable contexts:
1. Incidental cache variables
2. Required cache variables
[------local CMakeCache | local build compiler boundary-------]
3. exported cache variables
[------downstream CMakeCache | downstream build compiler boundar]
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
4. vertically disjoint sets: constant C/C++ values: maximally complete set across all lower orders that are at least weakly disjoint with all CMakeCache sets

---

What I will do to try and craft this into the build setup is to apply what I *think* I understand the variable to be doing. In practice this means

-> is this switch necessary for sub-components/dependencies build correctly? == non-transitive cache target properties
-> is this switch necessary for in-tree projects to configure and build correctly? == immediately transitive cache target properties
-> is this switch necessary for downstream codes to configure correctly? == some logical union of immediate and deferred target cache properties
-> is this multi-valued option necessary for at most one stage prior?' == always available but only in the build stage (through generated header files)

--

So, that's my plan for the next few days. Feel free to let me know I'm completely mismanaging this. Otherwise, glowing praise will be taken with only the absolute minimum of pride and hubris.

Yurlungur:
This  seems reasonable to me

dholladay00:
It seems possible that the requirements for in-tree builds will be different than built as a library. I suspect that in-tree builds will need many of the defines available. 

```c++
// header.h
void func(args) {
#ifdef OPTION_1
  do_stuff(args);
#else
  do_other_stuff(args);
}
```

```c++
// downstream_code_file.cpp
#include <header.h>

code_func(stuff) {
  auto args = get_args_from_stuff(stuff);
  func(args);
}
```

At the moment, we have code that is very specific to the host code already in singularity-eos and that should probably change in which even downstream codes that don't build in-tree will still need many of the defines.

PR #31, opened 2021-09-23
clang format

Yurlungur:
A long-awaited change: automated formatting. I put a `.clang-format` file I like in the highest level of the repo and ran it on all source.

At the moment, this is just a by hand procedure---run clang-format on files you modify. In the long term, of course, we should automate this. @mauneyc-LANL perhaps this is something that can go in the cmake?

It's possible to automate checking that formatting has been applied in the automated tests, but I don't really want to do that until we can type `make format` and format the whole repo. It would otherwise be a bit cumbersome.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,10 @@
+ColumnLimit: 90
+IndentWidth: 2
+AccessModifierOffset: -1
+
+AllowAllParametersOfDeclarationOnNextLine: true
+AllowShortIfStatementsOnASingleLine: true
+AlwaysBreakTemplateDeclarations: true
+BinPackArguments: true
+BinPackParameters: true
+ConstructorInitializerAllOnOneLineOrOnePerLine: false
--- a/README.md
+++ b/README.md
@@ -83,6 +83,15 @@ make -j
 make test
 ```
 
+## Formatting
+
+This project is formatted with `clang-format` using `clang-format`
+version 10.0.0. If contributing, please run clang-format on the files
+you modify.
+
+For now, we will periodically run clang-format on the repo to
+synchronize it. An automated solution will be forth coming.
+
 ## Copyright
 
 © 2021. Triad National Security, LLC. All rights reserved.  This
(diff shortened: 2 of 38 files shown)


brryan (approved):
I skimmed, looks fine to me. Always nice to have clang format

Yurlungur:
One pair of eyes approves. And this project has grown to the point where I think this kind of machinery is a requirement. I'm merging it.

mauneyc-LANL:
Do we want this on the user side, or as a github action/CI ?

Yurlungur:
If we can get the CI to do it automatically upon commit, that would be amazing. My initial thought was put it on the user side, and then have a CI test fail if formatting wasn't right. But just doing the formatting on the CI side would be even better.

jhp-lanl:
> If we can get the CI to do it automatically upon commit, that would be amazing. My initial thought was put it on the user side, and then have a CI test fail if formatting wasn't right. But just doing the formatting on the CI side would be even better.

If the formatter is run on every commit, that would mean an additional auto-generated commit after every developer commit, right? Could this cause conflicts if the developer doesn't fetch and pull the format changes after each commit?

Yurlungur:
> > If we can get the CI to do it automatically upon commit, that would be amazing. My initial thought was put it on the user side, and then have a CI test fail if formatting wasn't right. But just doing the formatting on the CI side would be even better.
> 
> If the formatter is run on every commit, that would mean an additional auto-generated commit after every developer commit, right? Could this cause conflicts if the developer doesn't fetch and pull the format changes after each commit?

Oh good point.

Yurlungur:
> Yes having to do extra `fetch`/`merge`s after pushing changes to the remote could be frustrating. Another option (maybe this is what @mauneyc-LANL meant) would be to provide installable git hooks e.g. https://github.com/lanl/Draco/blob/develop/environment/git/pre-commit-clang-format so formatting is applied automatically on the user side before pushing by users who opt-in, and the CI still blocks PRs that aren't formatted correctly.

This is kind of what I had in mind initially... although I also like being able to type `make format` and format the source directory.

mauneyc-LANL:
Sure, I'll see what I can conjure up as a cmake target.

PR #33, opened 2021-09-23
Update .gitignore

mauneyc-LANL:
add entries to ignore vscode settings files (`.vscode`), and clangd linting cache (`.cache`)


Yurlungur (approved):
Approved.

PR #35, opened 2021-10-19
Spec changed default of kokkos to off

mauneyc-LANL:
Default for `kokkos` set to off


Yurlungur (approved):
No objections... but why this change?

PR #37, opened 2021-11-02
Set up tests on gitlab

Yurlungur:
A significant update. Several changes:
- Most importantly, incorporates @ktsai7 's work to set up continuous integration on a gitlab mirror.
- Runs clang-format on the repo and bases clang format on clang-format 12
- Revives `SpinerEOS` tests. Cleans them up significantly to compare against eospac directly rather than hardcoded numbers
- Generates spiner tables at test time, rather than storing them as gold files

This has already gone under significant internal review, so if someone could give it a quick ok before we merge in, that would be great.


jhp-lanl:
I gave it a quick look, but I admit I'm probably not one to find serious bugs here. Looks okay to me though.

jhp-lanl (approved):

ktsai7:
@Yurlungur Sorry I didn't catch this earlier but there is a typo in the `.gitlab-ci.yml`

Yurlungur:
Ah no problem. Thanks for the catch. Merged it.

.gitlab-ci.yml
```diff
@@ -0,0 +1,210 @@
...
+      cmake -DSINGULARITY_USE_EOSPAC=ON \
+            -DSINGULARITY_USE_HDF5=ON \
+            -DSINGULARITY_BUILD_SESAME2SPINER=ON \
+            -DSINGULARITY_BUILD_STELLARCOLLAPSE2SPINER=ON \
+            -DSINGULARITY_BUILD_TESTS=ON \
+            -DSINGULARITY_TEST_SESAME=ON \
+            -DSINGULARITY_TEST_STELLAR_COLLAPSE=ON \
+            ..
+    - make -j
+    - ./sesame2spiner/sesame2spiner -s materials.sp5 ../sesame2spiner/examples/unit_tests/*.dat
```
dholladay00, line 151: Doesn't the spiner file need to get moved to the `singularity-eos-data` directory in order for certain tests to run? Will `make test` trigger a re-build such that the spiner file is copied into the necessary build directory and dependent tests are built and linked?

Yurlungur: I changed the path pointing to the file in the `unit_tests` file. Now it points one directory up:
```C++
const std::string eosName = "../materials.sp5";
```

dholladay00: Will `make test` re-check for the existence of that file and then proceed to build the eos test or will it just run the tests as built from the previous make command (which would not include the eos test since the needed file didn't exist at the time the make command was run).

Yurlungur: `make test` won't check for the existence of the sp5 file. But the `eos_unit_tests` binary will be built regardless, as the gold file is no longer listed as a build dependency. So if the commands are run in this order, it'll all work.

PR #42, opened 2021-11-03
use py-sphinx-rtd-theme@0.4.3 

ktsai7:


ktsai7:
@Yurlungur @jhp-lanl this is to fix the issue with installing `npm` through spack if we use the newest sphinx-rtd-theme

Yurlungur (approved):

PR #43, opened 2021-11-04
GitHub niceties

Yurlungur:
A few github niceties I am trying to implement:
- A PR template that tells users what to do when they submit a PR
- gitignore sphinx builds
- A make format target in cmake. Heavily based on the one in parthenon.
- A github action to check formatting, based on @ktsai7 's work in gitlab
- A github action to deploy our docs directly to github pages, using this open source [github pages action](https://github.com/marketplace/actions/github-pages-action).

By necessity I haven't tested all the workflows stuff. It might take me a few tries/commits/pushes to get it all right.


Yurlungur:
This is now ready for review. All listed features work.

Yurlungur:
A few questions regarding documentation that I haven't thought too hard about:
- Should the github pages site be versioned?
- Should we update the pages site on release, or just on every merge to main? I.e., should it be up-to-date with main? Or with releases?

Yurlungur:
1 more update. I added a skeleton to the documentation where we can put things. I also published a preliminary docs page on github pages, so we're live:
https://lanl.github.io/singularity-eos/

.github/workflows/docs.yml
```diff
@@ -0,0 +1,33 @@
...
+
+      steps:      
+        - name: Checkout code
+          uses: actions/checkout@v2
+        - name: Set system to non-interactive mode
+          run: export DEBIAN_FRONTEND=noninteractive
+        - name: install dependencies
+          run: |
+            pip install sphinx
+            pip install sphinx-rtd-theme
```
Yurlungur, line 22: I don't think there's a reason to use spack for this, but I could be convinced otherwise.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,33 @@
...
+        - name: build docs
+          run: |
+            cd doc/sphinx
+            make html
+        - name: Deploy
+          uses: peaceiris/actions-gh-pages@v3.7.3
```
Yurlungur, line 28: I don't know how I feel about using this 3rd-party github action. But it is open source, and it's running only on github. I skimmed the typescript but I didn't follow most of it. If people would feel more comfortable with it, I can write the procedure manually.

.github/workflows/formatting.yml
```diff
@@ -0,0 +1,24 @@
...
+        - name: Checkout code
+          uses: actions/checkout@v2
+        - name: Set system to non-interactive mode
+          run: export DEBIAN_FRONTEND=noninteractive
+        - name: install dependencies
+          run: |
+            sudo apt-get install -y --force-yes -qq git clang-format
+        - name: check formatting
+          continue-on-error: true
+          run: find . -regex '.*\.\(cpp\|hpp\)' | xargs clang-format -style=file -i && git diff --exit-code --ignore-submodules
```
Yurlungur, line 24: Thanks to @ktsai7 for this nice one-liner

jhp-lanl:
>     * Should the github pages site be versioned?

Do you mean just include the project version? I think we can just include `version` and `release` variables in the `conf.py`. Then I think you can just use those in an `html_title` variable there as well maybe?

It looks like there are fancier things you can do with linking versions of the documentation to various releases but I'm not sure how hard that is.

>     * Should we update the pages site on release, or just on every merge to main? I.e., should it be up-to-date with main? Or with releases?

I would suggest web documentation be updated for releases provided that there are instructions for building the documentation locally. I'd think that those who are using master will be working in the repo itself and can build the documentaiton while those working in the host codes may be working with releases and may want to just visit a web page.

ktsai7:
I think we can have versioned documentation for releases and Sphinx should have ways to handle that. That will need to be looked into real quick.

I agree with @jhp-lanl that unless you are working in singularity-eos, the versioned releases doc should be sufficient. For the developers, as long as they have instructions on how to build and update locally, that should be good, but if you want a doc that's even later than the latest release, then perhaps we could have something that's versioned as `develop` or something?

Yurlungur:
I'm fine with just doing things versioned on release. I believe there's a way to do that automatically with the github actions I've set up. Having a develop version would be convenient too, but not important.

Yurlungur:
There appears to be this tool, which seems to be able to generate versions
https://github.com/sphinx-contrib/sphinxcontrib-versioning

Yurlungur:
> There appears to be this tool, which seems to be able to generate versions https://github.com/sphinx-contrib/sphinxcontrib-versioning

sphinxcontrib-versioning seems unmaintained. sphinx-multiversion looks like it works
https://holzhaus.github.io/sphinx-multiversion/master/index.html

Yurlungur:
@ktsai7 @jhp-lanl how's this? I got it to generate documentation for every branch and tag. And it should be straightforward to automatically direct to the latest release. (By default it points to main at the moment.)

Yurlungur:
Also first draft of documentation has gone through RASSTI and has an LA-UR.

jhp-lanl:
This looks great! I'll add my approval but I'd like @ktsai7 's eyes before this gets merged.

jhp-lanl (approved):

annapiegra (approved):

ktsai7 (approved):

Yurlungur:
People seem happy. Merging it in.

PR #48, opened 2021-11-11
fix issues with duplicate names

### LINKED ISSUE #47: sesame2spiner should handle material names
There should be a way to handle sesame files with missing or duplicate material names. Issue found by @dholladay00

Yurlungur:
## PR Checklist

In issue #47 @dholladay00 discovered that if two materials in a sesame2spiner file have the same "name" metadata, HDF5 throws an error. This PR adds guard rails so this can't happen, and also adds guard rails against a similar error when duplicate matids are requested.

Strategy uses a simple `std::unordered_list` to see what has been used already.

Resolves #47

### DIFF
--- a/sesame2spiner/generate_files.cpp
+++ b/sesame2spiner/generate_files.cpp
@@ -17,6 +17,7 @@
 #include <cmath>
 #include <cstdlib>
 #include <string>
+#include <unordered_set>
 #include <vector>
 
 #include <hdf5.h>
@@ -148,6 +149,8 @@ herr_t saveAllMaterials(const std::string &savename,
                         Verbosity eospacWarn) {
   std::vector<Params> params;
   std::vector<int> matids;
+  std::unordered_set<std::string> used_names;
+  std::unordered_set<int> used_matids;
   SesameMetadata metadata;
   hid_t file;
   herr_t status = H5_SUCCESS;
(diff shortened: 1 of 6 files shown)


dholladay00:
@Yurlungur did you test sesame2spiner and verify that it avoids the issues.

Yurlungur:
> @Yurlungur did you test sesame2spiner and verify that it avoids the issues. 

Yes. I added a new set of test inputs in the example directory, which gets run by the CI. These inputs have duplicate names and matids.

That said, it would be good to try it on your problem.

dholladay00:
> > @Yurlungur did you test sesame2spiner and verify that it avoids the issues.
> 
> Yes. I added a new set of test inputs in the example directory, which gets run by the CI. These inputs have duplicate names and matids.
> 
> That said, it would be good to try it on your problem.

These changes fixed the issue I was seeing.

Yurlungur:
> > > @Yurlungur did you test sesame2spiner and verify that it avoids the issues.
> > 
> > Yes. I added a new set of test inputs in the example directory, which gets run by the CI. These inputs have duplicate names and matids.
> > 
> > That said, it would be good to try it on your problem.
> 
> These changes fixed the issue I was seeing. 

Can you review/approve then?

sesame2spiner/generate_files.cpp
```diff
@@ -171,12 +174,27 @@ herr_t saveAllMaterials(const std::string &savename,
...
     eosGetMetadata(matid, metadata, Verbosity::Debug);
     if (printMetadata) std::cout << metadata << std::endl;
     std::string name = params[i].Get("name", metadata.name);
+    if (used_names.count(name) > 0) {
+      std::string new_name = "material_" + std::to_string(i);
+      std::cerr << "...WARNING: Name " << name << " already used. "
+		<< "using a default name: "
+		<< new_name
+		<< std::endl;
+      name = new_name;
```
dholladay00, line 195: what about using the repeated name + an integer @Yurlungur ?

dholladay00: @Yurlungur I'll approve once we resolve how to move forward on this.

Yurlungur: Yeah, name + integer is fine. I did it this way because it's easier than tracking an integer count, but integer count should be easy too. Been under the weather a few days. Will do this when I'm at my desk again.

dholladay00: maybe we use a default name if the "reused name" is an empty string, but if there is an actual name that is re-used, preserving some of that information could be useful to downstream users.

Yurlungur: I'm not sure I can catch all the possible "empty string" names but I'll try.

Yurlungur: @dholladay00 the new version does the following:
- If the name is an empty string or the string "-1", which I noticed some unnamed materials have in their name field, rename the material "material_n" where n is the number of the input deck passed into sesame2spiner
- If matids are repeated, skips the duplicate matid
- If two materials have the same, non-trivial name, `sesame2spiner` will enumerate them. i.e., if you had three materials named "metal," then in the sp5 file they'll be called `metal`, `metal_2`, `metal_3`.

Give it a try.

dholladay00: This works great now!

dholladay00 (approved):

PR #60, opened 2021-12-22
add build doc

Yurlungur:
## PR Summary

Document how to build `singularity-eos` in all its complicated glory.

## PR Checklist

### DIFF
--- a/doc/sphinx/src/building.rst
+++ b/doc/sphinx/src/building.rst
@@ -18,6 +18,7 @@ simply call
 .. _Spack: https://spack.io/
 
 .. code-block:: bash
+
   git clone --recursive git@github.com:lanl/singularity-eos.git
   spack repo add singularity-eos/spack-repo
   spack install singularity-eos
@@ -212,6 +213,7 @@ If you use spack, but would like to build ``singularity-eos`` from
 source, you can install dependencies via, e.g.,
 
 .. code-block:: bash
+
   git clone --recursive git@github.com:lanl/singularity-eos.git
   spack repo add singularity-eos/spack-repo
   spack install --only dependencies singularity-eos+cuda cuda_arch=70
--- a/doc/sphinx/src/building.rst
+++ b/doc/sphinx/src/building.rst
@@ -24,44 +24,45 @@ simply call
 
 to install ``singularity-eos`` into your spack instance. The spackage
 supports a number of relevant variants:
-+-----------------------------+---------+-----------------+-----------------------------+
-| Variant Name [default]      | When    | Allowed Values  | Description                 |
-+=============================+=========+=================+=============================+
-| build_extra [none]          | -----   | none, sesame,   | Build sesame2spiner         |
-|                             |         | stellarcollapse | or stellarcollapse2spiner   |
-+-----------------------------+---------+-----------------+-----------------------------+
-| build_type [RelWithDebInfo] | -----   | Debug, Release, | Equivalent to               |
-|                             |         | RelWitHDebInfo, | -DCMAKE_BUILD_TYPE          |
-|                             |         | MinSizeRel      | in cmake build              |
-+-----------------------------+---------+-----------------+-----------------------------+
-| cuda [off]                  | -----   | on, off         | Build with cuda             |
-+-----------------------------+---------+-----------------+-----------------------------+
-| cuda_arch [none]            | +cuda   | see kokkos spec | The target GPU architecture |
-+-----------------------------+---------+-----------------+-----------------------------+
-| doc [off]                   | -----   | on, off         | Build sphinx docs           |
-+-----------------------------+---------+-----------------+-----------------------------+
-| format [off]                | -----   | on, off         | Support for clang-format    |
-+-----------------------------+---------+-----------------+-----------------------------+
-| fortran [on]                | -----   | on, off         | Provide fortran bindings    |
-+-----------------------------+---------+-----------------+-----------------------------+
-| ipo [off]                   | -----   | on, off         | CMake interprocedural       |
-|                             |         |                 | optimization                |
-+-----------------------------+---------+-----------------+-----------------------------+
-| kokkos [off]                | -----   | on, off         | Enable Kokkos backend       |
-|                             |         |                 | Required for cuda support   |
-+-----------------------------+---------+-----------------+-----------------------------+
-| kokkos-kernels [off]        | -----   | on, off         | Use kokkos-kernels for      |
-|                             |         |                 | linear algebra suport,      |
-|                             |         |                 | which is needed with        |
-|                             |         |                 | mixed-cell closures on GPU  |
-+-----------------------------+---------+-----------------+-----------------------------+
-| mpi [off]                   | -----   | on, off         | Build with parallel HDF5    |
-|                             |         |                 | otherwise build with serial |
-+-----------------------------+---------+-----------------+-----------------------------+
-| opemp [off]                 | +kokkos | on, off         | Build Kokkos openmp backend |
-+-----------------------------+---------+-----------------+-----------------------------+
-| tests [off]                 | -----   | on, off         | Build tests                 |
-+-----------------------------+---------+-----------------+-----------------------------+
+
++-----------------------------+-----------------+-----------------------------+
+| Variant Name [default]      | Allowed Values  | Description                 |
++=============================+=================+=============================+
+| build_extra [none]          | none, sesame,   | Build sesame2spiner         |
+|                             | stellarcollapse | or stellarcollapse2spiner   |
++-----------------------------+-----------------+-----------------------------+
+| build_type [RelWithDebInfo] | Debug, Release, | Equivalent to               |
+|                             | RelWitHDebInfo, | -DCMAKE_BUILD_TYPE          |
+|                             | MinSizeRel      | in cmake build              |
++-----------------------------+-----------------+-----------------------------+
+| cuda [off]                  | on, off         | Build with cuda             |
++-----------------------------+-----------------+-----------------------------+
+| cuda_arch [none]            | see kokkos spec | The target GPU architecture |
++-----------------------------+-----------------+-----------------------------+
+| doc [off]                   | on, off         | Build sphinx docs           |
++-----------------------------+-----------------+-----------------------------+
+| format [off]                | on, off         | Support for clang-format    |
++-----------------------------+-----------------+-----------------------------+
+| fortran [on]                | on, off         | Provide fortran bindings    |
++-----------------------------+-----------------+-----------------------------+
+| ipo [off]                   | on, off         | CMake interprocedural       |
+|                             |                 | optimization                |
++-----------------------------+-----------------+-----------------------------+
+| kokkos [off]                | on, off         | Enable Kokkos backend       |
+|                             |                 | Required for cuda support   |
++-----------------------------+-----------------+-----------------------------+
+| kokkos-kernels [off]        | on, off         | Use kokkos-kernels for      |
+|                             |                 | linear algebra suport,      |
+|                             |                 | which is needed with        |
+|                             |                 | mixed-cell closures on GPU  |
++-----------------------------+-----------------+-----------------------------+
+| mpi [off]                   | on, off         | Build with parallel HDF5    |
+|                             |                 | otherwise build with serial |
++-----------------------------+-----------------+-----------------------------+
+| openmp [off]                | on, off         | Build Kokkos openmp backend |
++-----------------------------+-----------------+-----------------------------+
+| tests [off]                 | on, off         | Build tests                 |
++-----------------------------+-----------------+-----------------------------+
 
 Building Via Cmake
 --------------------
--- a/doc/sphinx/src/building.rst
+++ b/doc/sphinx/src/building.rst
@@ -3,3 +3,216 @@
 Building Singularity-EOS
 =========================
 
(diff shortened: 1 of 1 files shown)


jhp-lanl:
Thanks for working on this!

doc/sphinx/src/building.rst
```diff
@@ -3,3 +3,216 @@
...
+
+.. _Spack: https://spack.io/
+
+.. code-block:: bash
+  git clone --recursive git@github.com:lanl/singularity-eos.git
+  spack repo add singularity-eos/spack-repo
+  spack install singularity-eos
+
+to install ``singularity-eos`` into your spack instance. The spackage
+supports a number of relevant variants:
```
annapiegra, line 26: Is the spackage ready for general use:
I tried: 
`spack install singularity-eos +kokkos +tests build_extra=sesame`
and got:
`py-h5py requires hdf5 variant +mpi, but spec asked for ~mpi`

mauneyc-LANL: Is mpi an external dependency or being scheduled to build by spack? i.e. is there an external, non-buildable mpi in a package.yaml file? Also, is this in the context of an environment? there may be something in `$HOME/.spack/` that is conflicting here.

doc/sphinx/src/building.rst
```diff
@@ -3,3 +3,216 @@
...
++-----------------------------+---------+-----------------+-----------------------------+
+| kokkos-kernels [off]        | -----   | on, off         | Use kokkos-kernels for      |
+|                             |         |                 | linear algebra suport,      |
+|                             |         |                 | which is needed with        |
+|                             |         |                 | mixed-cell closures on GPU  |
++-----------------------------+---------+-----------------+-----------------------------+
+| mpi [off]                   | -----   | on, off         | Build with parallel HDF5    |
+|                             |         |                 | otherwise build with serial |
++-----------------------------+---------+-----------------+-----------------------------+
+| opemp [off]                 | +kokkos | on, off         | Build Kokkos openmp backend |
```
dholladay00, line 61: `openmp`

doc/sphinx/src/building.rst
```diff
@@ -3,3 +3,216 @@
...
+.. _kokkos-kernels: https://github.com/kokkos/kokkos-kernels/
+
+If you use spack, but would like to build ``singularity-eos`` from
+source, you can install dependencies via, e.g.,
+
+.. code-block:: bash
+  git clone --recursive git@github.com:lanl/singularity-eos.git
+  spack repo add singularity-eos/spack-repo
+  spack install --only dependencies singularity-eos+cuda cuda_arch=70
+
```
dholladay00, line 217: If I have a checkout of singularity-eos and I want to build in a non-spack build directory, is there still a way to get the cmake options to match the spack options @mauneyc-LANL ? When there was a deps package, there was a cmake configuration file generated to aid that process, does such a thing still exist, or can we specify via spack where we want the whole singularity-eos spackage to be built?

doc/sphinx/src/building.rst
annapiegra: Removing ~/.spack and adding +mpi concretized (will not concretize w/o +mpi) and currently building.

mauneyc-LANL: I will update the spackage to conflict if `+sesame~mpi`, which seems to be causing the issue (or whatever else variant requires `py-hdf5`

Yurlungur: don't know why hdf5 requires MPI. It shouldn't. But... 🤷 python is always finicky.

doc/sphinx/src/building.rst
Yurlungur: This is a good question that I would also like to know.

Yurlungur:
@dholladay00 @annapiegra @mauneyc-LANL there seem to be some questions regarding the spackage. That said, I'd like to bring the documentation into the code regardless of any changes/fixes to the spack that may be needed. How do you all feel about that?

doc/sphinx/src/building.rst
ktsai7: spack has a `install_tree` config that we could modify on a user level, but I don't know if that's the best approach you guys want here

doc/sphinx/src/building.rst
ktsai7: @Yurlungur I think `py-hdf5` is needed otherwise something doesn't work when we were first adding the CI?

Yurlungur: Right `py-hdf5` is needed for the `stellarcollapse` tests. What confuses me is that it needs `mpi`. There's no reason to build parallel hdf5 for python.

doc/sphinx/src/building.rst
mauneyc-LANL: This feature is not included in the `package.py`, tho I can't recall why I cut it. 

As for controlling where it's build, that would require a global configuration change. Something like

```bash
$ spack location --install-dir gcc
~/spack/opt/spack/linux-centos7-x86_64/gcc-4.9.3-iy4rw...
```

would be helpful in the short term. Long-term, probably something like what `-deps` was doing seemed useful for devs, so we can return to it

mauneyc-LANL:
> @dholladay00 @annapiegra @mauneyc-LANL there seem to be some questions regarding the spackage. That said, I'd like to bring the documentation into the code regardless of any changes/fixes to the spack that may be needed. How do you all feel about that?

Dumb question: should we be putting "sphinx"-style comments in the spackage?

Yurlungur:
> > @dholladay00 @annapiegra @mauneyc-LANL there seem to be some questions regarding the spackage. That said, I'd like to bring the documentation into the code regardless of any changes/fixes to the spack that may be needed. How do you all feel about that?
> 
> Dumb question: should we be putting "sphinx"-style comments in the spackage?

I'm not sure that's needed... we're basically using sphinx rst to hand-write the docs, but we're not doing "doxygen" style automated documentation. I'd be open to that path though if people think it's worth it.

doc/sphinx/src/building.rst
Yurlungur: could we just have both, @mauneyc-LANL ? A deps package and a full package? And the devs can just install the deps package?

mauneyc-LANL: We could but I don't think it's necessary. Let's discuss some options later today (if you're around). I can try to mock up something later today.

jhp-lanl:
> > > @dholladay00 @annapiegra @mauneyc-LANL there seem to be some questions regarding the spackage. That said, I'd like to bring the documentation into the code regardless of any changes/fixes to the spack that may be needed. How do you all feel about that?
> > 
> > 
> > Dumb question: should we be putting "sphinx"-style comments in the spackage?
> 
> I'm not sure that's needed... we're basically using sphinx rst to hand-write the docs, but we're not doing "doxygen" style automated documentation. I'd be open to that path though if people think it's worth it.

I think the only advantage of using docstrings in the `package.py` file (which I think is what is being proposed?) is that the comment will exist where the code change is likely being made. Editing two files just increases the chance that the documentation will be out of date. Though as you pointed out Jonah, there's always that chance since the documentation isn't auto-generated so the advantage may not be significant.

Yurlungur:
> > > > @dholladay00 @annapiegra @mauneyc-LANL there seem to be some questions regarding the spackage. That said, I'd like to bring the documentation into the code regardless of any changes/fixes to the spack that may be needed. How do you all feel about that?
> > > 
> > > 
> > > Dumb question: should we be putting "sphinx"-style comments in the spackage?
> > 
> > 
> > I'm not sure that's needed... we're basically using sphinx rst to hand-write the docs, but we're not doing "doxygen" style automated documentation. I'd be open to that path though if people think it's worth it.
> 
> I think the only advantage of using docstrings in the `package.py` file (which I think is what is being proposed?) is that the comment will exist where the code change is likely being made. Editing two files just increases the chance that the documentation will be out of date. Though as you pointed out Jonah, there's always that chance since the documentation isn't auto-generated so the advantage may not be significant.

Also for the purpose of building the code, commenting the code in the spackage probably isn't useful. All we actually care about is what spack variants are available in the spackage. Though maybe there's a way to automatically generate the table I hand-wrote here, which would be nice.

doc/sphinx/src/building.rst
Yurlungur: yep sounds good

ktsai7: there is the option of having `singularity-eos-deps` spackage and inherit from it to make the `singularity-eos` spackage... at least you won't have to maintain in two places

doc/sphinx/src/building.rst
Yurlungur: Ah I think I found the problem. `py-h5py` has `mpi` on by default in its variant. Do we need to pass `~mpi` down through it?

ktsai7:
> > > > > @dholladay00 @annapiegra @mauneyc-LANL there seem to be some questions regarding the spackage. That said, I'd like to bring the documentation into the code regardless of any changes/fixes to the spack that may be needed. How do you all feel about that?
> > > > 
> > > > 
> > > > Dumb question: should we be putting "sphinx"-style comments in the spackage?
> > > 
> > > 
> > > I'm not sure that's needed... we're basically using sphinx rst to hand-write the docs, but we're not doing "doxygen" style automated documentation. I'd be open to that path though if people think it's worth it.
> > 
> > 
> > I think the only advantage of using docstrings in the `package.py` file (which I think is what is being proposed?) is that the comment will exist where the code change is likely being made. Editing two files just increases the chance that the documentation will be out of date. Though as you pointed out Jonah, there's always that chance since the documentation isn't auto-generated so the advantage may not be significant.
> 
> Also for the purpose of building the code, commenting the code in the spackage probably isn't useful. All we actually care about is what spack variants are available in the spackage. Though maybe there's a way to automatically generate the table I hand-wrote here, which would be nice.

It's not exactly a table, but we could pull in code into the `.rst` files so when someone update the code it would get pull in automatically...

doc/sphinx/src/building.rst
Yurlungur: Right that's what I was thinking.

dholladay00: Definitely don't want to maintain 2 packages, but I'd like to hear what @mauneyc-LANL is thinking. I know the spack people want folks to move away from `-deps` spackages and I'm assuming they have reasons and I'd like to understand that better before going against the grain.

doc/sphinx/src/building.rst
annapiegra: Ok, spack build w/ ~mpi fails with:
  ```
   131                     from /tmp/annap/spack-stage/spack-stage-singularit
            y-eos-main-brgwdvmhfkzipj2bjry2g2e5ws6em2hj/spack-src/singularity-e
            os/eos/singularity_eos.cpp:19:
  >> 132    /tmp/annap/spack-stage/spack-stage-singularity-eos-main-brgwdvmhfkz
            ipj2bjry2g2e5ws6em2hj/spack-src/singularity-eos/closure/mixed_cell_
            models.hpp:272:2: error: #error "Eigen should not be used with Kokk
            os."
     133      272 | #error "Eigen should not be used with Kokkos."
     134          |  ^~~~~
 
```

and 

```
     141    /usr/bin/ld: /projects/artab/users/annap/spack-environments/DWP9X/1
            .8.0/spack/opt/spack/linux-rhel7-power9le/gcc-9.3.0/eospac-6.4.2bet
            a-jxle75damh3twfcaicx7xdboktbyugk2/lib/libeospac6.a(eos_Interpolati
            on.o): In function `eos_BiRationalInterpolate':
     142    eos_Interpolation.c:(.text+0x520c): toc optimization is not support
            ed for 0xe4e50002 instruction.
     143    make[2]: Leaving directory `/ram/tmp/annap/spack-stage/spack-stage-
            singularity-eos-main-brgwdvmhfkzipj2bjry2g2e5ws6em2hj/spack-build-b
            rgwdvm'
     144    [ 73%] Built target sesame2spiner
  >> 145    make[2]: *** [singularity-eos/eos/CMakeFiles/eos_objs.dir/singulari
            ty_eos.cpp.o] Error 1
     146    make[2]: Leaving directory `/ram/tmp/annap/spack-stage/spack-stage-
            singularity-eos-main-brgwdvmhfkzipj2bjry2g2e5ws6em2hj/spack-build-b
            rgwdvm'
  >> 147    make[1]: *** [singularity-eos/eos/CMakeFiles/eos_objs.dir/all] Erro
            r 2
     148    make[1]: *** Waiting for unfinished jobs....

```

annapiegra:
> 

I would add a note saying spack builds are still under development.

doc/sphinx/src/building.rst
mauneyc-LANL: @Yurlungur What was the status of this error? Did we ever decide what to do about it?

Yurlungur: um... I don't understand why we're hitting that error. It *should* only happen if you build with Kokkos. That said that macro is wrong. I will try to push a correction to this branch.

Yurlungur:
> > 
> 
> I would add a note saying spack builds are still under development. Besides that I think it's great to have build instructions!

I'll add that.

dholladay00:
I'm going to approve this as it is a good starting place for building documentation. There are still issues with the spack itself and those can be reported elsewhere, the documentation itself is sufficient for now.

dholladay00 (approved):

PR #63, opened 2021-12-23
Fix the issue in extrapolation

Yurlungur:
## PR Summary

@annapiegra noticed that `SpinerEOSDependsRhoT` was producing an artifact near the top of the table, right around when the code switches to extrapolation. This was caused by two issues:
1. `sesame2spiner` offsets the grid log grid bounds slightly based on STP. The intent is that this regridding sets up exactly one grid point at standard density and one at standard temperature. To do so, I change the grid spacing and bounds slightly from what's specified in the file. This machinery was accidentally removing too many points from the top of the grid. It now removes *exactly* one point, meaning the upper bound of the table is closer to what's expected.
2. There was a bug in the extrapolation routines of `SpinerEOSDependsRhoT` where the Gruneisen parameter used for the ideal gas extrapolation was computed incorrectly.

Both bugs are now fixed. I am also adding a unit test to check for this behavior. 

Finally, this PR includes a new utility in the examples directory, `get_sesame_state`. This is built if you build examples. You pass it a matid, a spiner file, a density, a temperature, and an energy. It then uses both the spiner and eospac backends to print a few relevant calls at (density, temperature) and (density,energy) pairs. This utility was pretty useful for debugging this problem, and I think it'll be useful going forward.

## PR Checklist

- [x]  Format your changes by using the `make format` command after configuring with `cmake`.

### DIFF
--- a/example/CMakeLists.txt
+++ b/example/CMakeLists.txt
@@ -18,3 +18,10 @@ target_link_libraries(get_sound_speed_press PRIVATE
   eos
   singularity-eos::libs
   singularity-eos::flags)
+
+add_executable(get_sesame_state
+  get_sesame_state.cpp)
+target_link_libraries(get_sesame_state PRIVATE
+  eos
+  singularity-eos::libs
+  singularity-eos::flags)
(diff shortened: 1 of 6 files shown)


chadmeyer (approved):
This all seems reasonable to me.

jhp-lanl (approved):

PR #64, opened 2021-12-23
Spack updates

mauneyc-LANL:
Changes:

- `mpi` variant propagated to dependencies correctly
- re-implementation of cmake configuration file generation.
- documentation stubs

Fixing various issues with spack installations that have arisen. Will try and address open issues here.

## PR Checklist

### DIFF
--- a/doc/sphinx/src/building.rst
+++ b/doc/sphinx/src/building.rst
@@ -231,7 +231,7 @@ are consistent
 
 .. code-block:: bash
 
-   spack install --until install singularity-eos
+   spack install singularity-eos
    spack load singularity-eos
    cd <to/build/dir>
    cmake -C $SINGULARITY_SPACK_CMAKE_CONFIG <path/to/source/dir>
--- a/spack-repo/packages/singularity-eos/package.py
+++ b/spack-repo/packages/singularity-eos/package.py
@@ -21,13 +21,10 @@ class SingularityEos(CMakePackage, CudaPackage):
     variant("format", default=False, description="Clang-Format Support")
 
     #depends_on("mpark-variant")
-    depends_on("hdf5+cxx+hl~mpi", when="~mpi")
-    depends_on("hdf5+cxx+hl+mpi", when="+mpi")
     depends_on("eospac")
-    depends_on("py-h5py")
 
     depends_on("cmake@3.12:")
-    depends_on("eigen@3.3.9", when="~kokkos-kernels")
+    depends_on("eigen@3.3.9", when="~cuda")
     depends_on("catch2@2.13.4:2.13.6", when="+tests")
 
     depends_on("py-sphinx", when="+doc")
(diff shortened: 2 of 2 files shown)


ktsai7:
Does `singularity-eos` have to be on `eigen@3.3.9`, `catch2@2.13.4:2.13.6` and `kokkos@develop`?

Would `eigen@3.3.8`, `catch2@2.12.3`, `kokkos@3.2.00` be acceptable as well?

Yurlungur:
> Does `singularity-eos` have to be on `eigen@3.3.9`, `catch2@2.13.4:2.13.6` and `kokkos@develop`?
> 
> Would `eigen@3.3.8`, `catch2@2.12.3`, `kokkos@3.2.00` be acceptable as well?

From my perspective those would all be fine.

Yurlungur (commented):
Thanks for doing this, and adding the cmake config file back in. That's great to have.

ktsai7:
> > Does `singularity-eos` have to be on `eigen@3.3.9`, `catch2@2.13.4:2.13.6` and `kokkos@develop`?
> > Would `eigen@3.3.8`, `catch2@2.12.3`, `kokkos@3.2.00` be acceptable as well?
> 
> From my perspective those would all be fine.

Then it would be nice if we don't required the package version that only start existing on/after spack@v0.17 or spack@develop

cmauney:
> Does `singularity-eos` have to be on `eigen@3.3.9`, `catch2@2.13.4:2.13.6` and `kokkos@develop`?
> 
> Would `eigen@3.3.8`, `catch2@2.12.3`, `kokkos@3.2.00` be acceptable as well?

These have been set in cbdae9f

cmauney:
Spec with GPU support (gcc@9.4.0):

`singularity-eos+cuda+kokkos+kokkos-kernels build_extra=sesame,stellarcollapse cuda_arch=70`

cmauney:
@ktsai7 @Yurlungur @dholladay00 

The workflow has changed somewhat. If you want to just install dependencies, *but not the pre-configured cmake*, you can simply do
```
$> spack install --only dependencies singularity-eos
```
If you *do* want the pre-configured cmake, you have do a regular install
```
$> spack install singularity-eos
```
This will do a full build of the repo and install it into spack's storage. The reason for this is, in order to cleanly install the preconfigured cmake cache, the spackage has to enter the install phase.

I've tried a few hacks to get around this using only internal spack logic, but so far nothing really sticks. One option to overcome this is to use a new variant, such as `+dev_only` or similar, and to short-circuit the spackage install routine to only install the cache file if it is set. I don't *love* this idea, but a custom install routine is not uncommon. Or, we could resurrect the `deps` spackage as an optional dependency, but it would only exist to generate the cmake cache.

Another option is to not bother, and just do a full build+install, as this provides a quick check that the compiler/arch/spec is compatible. This shouldn't interfere with a manual build (tho I don't know if I can guarantee this yet).

I will bang my head against this a little more, but if there are preferences and/or advice on how proceed let me know.

Yurlungur:
> I've tried a few hacks to get around this using only internal spack logic, but so far nothing really sticks. One option to overcome this is to use a new variant, such as `+dev_only` or similar, and to short-circuit the spackage install routine to only install the cache file if it is set. I don't _love_ this idea, but a custom install routine is not uncommon. Or, we could resurrect the `deps` spackage as an optional dependency, but it would only exist to generate the cmake cache.
> 
> Another option is to not bother, and just do a full build+install, as this provides a quick check that the compiler/arch/spec is compatible. This shouldn't interfere with a manual build (tho I don't know if I can guarantee this yet).

I don't have a strong preference between these options... if we were to submit this upstream to the public spack repos, is there a preferred solution that the spack team would encourage?

cmauney:
> > I've tried a few hacks to get around this using only internal spack logic, but so far nothing really sticks. One option to overcome this is to use a new variant, such as `+dev_only` or similar, and to short-circuit the spackage install routine to only install the cache file if it is set. I don't _love_ this idea, but a custom install routine is not uncommon. Or, we could resurrect the `deps` spackage as an optional dependency, but it would only exist to generate the cmake cache.
> > Another option is to not bother, and just do a full build+install, as this provides a quick check that the compiler/arch/spec is compatible. This shouldn't interfere with a manual build (tho I don't know if I can guarantee this yet).
> 
> I don't have a strong preference between these options... if we were to submit this upstream to the public spack repos, is there a preferred solution that the spack team would encourage?

I'm under the impression is they would rather this sort of thing happen outside of spack, which is also an option (e.g. some script)

dholladay00:
I like the script idea, especially if this is a tool that only devs are going to need and the host codes are just going to use spack in the intended way.

Yurlungur:
How would the script machinery work? Installing via spack would provide a script in the binary directory or something?

dholladay00:
hmmm... this is a good question. I guess I was simplistically thinking the script could live in `/utils/scripts`, but I'm not sure how it would know anything about the spack installation.

Yurlungur (approved):
Although I think there's still interest in getting the cmake config independent of the build... I think we should get these changes in regardless, and tackle that question later. Any objections to merging this in?

dholladay00:
lets go ahead and merge

PR #65, opened 2022-01-14
Correct Gruneisen parameter formula

jhp-lanl:
Fix bug in the Gruneisen parameter for the Gruneisen EOS

## PR Summary

As implemented, the energy derivative of the pressure for this EOS would take the following form:
Gruneisen1

However, this form violates a key property of the Gruneisen EOS where the energy derivative should be constant when the parameter b = 0. Changing the first rho to rho_0 fixes this problem in the above equation which results in a Gruneisen parameter given by

Gruneisen2

which is consistent with the energy derivative being constant when b=0. This is also consistent with the corrected EOS in other hydocodes.

Note that this is part of updating documentation so I'll update the documentation in a separate MR.

@chadmeyer can you review this

## PR Checklist

### DIFF
--- a/singularity-eos/eos/eos.hpp
+++ b/singularity-eos/eos/eos.hpp
@@ -174,7 +174,7 @@ class Gruneisen {
   // static constexpr const char _eos_type[] = {"Gruneisen"};
   PORTABLE_INLINE_FUNCTION
   Real Gamma(const Real rho) const {
-    return rho < _rho0 ? _G0 : _G0 + _b * (rho / _rho0 - 1.0);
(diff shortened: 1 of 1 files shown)


Yurlungur (approved):
LGTM. Thanks for the catch. @chadmeyer what do you think?

jhp-lanl:
I'm going to skip adding tests for the moment because I'd like to refactor the code later to be more readable and I'll add tests then

Yurlungur:
There currently don't exist any tests for the Gruneisen EOS anyway. Adding some should be on our TODO list.

jhp-lanl:
@chadmeyer any objections?

chadmeyer:
I trust @jhp-lanl and his judgment on the model.

PR #70, opened 2022-01-21
Fix for JWL EOS

jhp-lanl:
## PR Summary

It appears that the reference energy and pressure for the JWL EOS were swapped as a typo.

The reference curve for the JWL EOS is an isentrope and a quick sanity check is that the pressure should be exactly equal to the volume derivative of the energy along the isentrope. This fix makes that true.

My fix is also consistent with xRAGE.

## PR Checklist

### DIFF
--- a/singularity-eos/eos/eos_jwl.cpp
+++ b/singularity-eos/eos/eos_jwl.cpp
@@ -17,11 +17,11 @@
 
 namespace singularity {
 
-PORTABLE_FUNCTION Real JWL::ReferenceEnergy(const Real rho) const {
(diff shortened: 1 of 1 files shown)


jhp-lanl:
I didn't make any tests yet because this is another area where we need to just make more tests and I intend to tackle that in another MR later.

This was found while making documentation so that will also be forthcoming.

@chadmeyer please let me know if you approve of the change.

Yurlungur (approved):
LGTM but I defer to Chad.

chadmeyer (approved):
This is correct; I made a mistake when I first put it together.

PR #71, opened 2022-01-21
EOS Model documentation

jhp-lanl:
## PR Summary

This PR adds documentation for the analytic EOS models available in the library and provides a placeholder for the tabular EOS and the stellar collapse EOS.

## PR Checklist

- [x] Adds a test for any bugs fixed. Adds tests for new features. (N/A)
- [x] Format your changes by using the `make format` command after configuring with `cmake`. (N/A)


jhp-lanl:
@chadmeyer and @Yurlungur please look things over to make sure I'm not making any obvious mistakes in my documentation. Also feel free to suggest formatting changes.

doc/sphinx/src/models.rst
```diff
@@ -3,3 +3,409 @@
...
+
+``singularity-eos`` contains a number of equations of state that are most useful
+in hydrodynamics codes. All of the equations of state presented here are
+considered "complete" equations of state in the sense that they contain all the
+information needed to derive a given thermodynamic quantity. However, not all
+variables are exposed since the emphasis is on only those that are needed for
+hydrodynamics codes. An incomplete equation of state is often sufficient for
+pure hydrodynamics since it can relate pressure to a given energy-density state,
+but it is missing information that relates the energy to temperature (which in
+turn provides a means to calculate entropy).
```
Yurlungur, line 26: Beyond the scope of this PR, but at some point I'd like to attack entropy more formally. Having access to entropy vs rho and T (for example) would be a generically useful capability. It's needed, actually, for some applications.

doc/sphinx/src/models.rst
```diff
@@ -3,3 +3,409 @@
...
+in the :doc:`EOS API <using-eos>`.
+
++--------------------------+----------------------+------------------------+
+| Function                 | Dependent Variable   | Independent Variables  |
++==========================+======================+========================+
+| :math:`T(\rho, E)`       | Temperature          | Density, Energy        |
++--------------------------+----------------------+                        |
+| :math:`P(\rho, E)`       | Pressure             |                        |
++--------------------------+----------------------+------------------------+
+| :math:`E(\rho, T)`       | Energy               | Density, Temperature   |
```
Yurlungur, line 69: `singularity-eos` never uses total energy. Basically everywhere it uses specific internal energy sie, or \varepsilon.

doc/sphinx/src/models.rst
```diff
@@ -3,3 +3,409 @@
...
+units for thermodynamic quantities:
+
++------------+------------------+---------------------------------------+-----------------------+
+|Quantity    | Default Units    | cgµ conversion                        | Sesame Units          |
++============+==================+=======================================+=======================+
+|:math:`P`   | µbar             | 10\ :sup:`-12` Mbar                   | 10\ :sup:`-10` GPa    |
++------------+------------------+---------------------------------------+-----------------------+
+|:math:`\rho`| g/cm\ :sup:`3`   | 1                                     | 1                     |
++------------+------------------+---------------------------------------+-----------------------+
+|:math:`E`   | erg              | 10\ :sup:`-12` Mbar-cm\ :sup:`3`/g    | 10\ :sup:`-10` kJ/g   |
```
Yurlungur, line 106: See above comment about energy. I suggest we use \varepsilon for specific internal energy here, which has units of ergs/g.

doc/sphinx/src/models.rst
```diff
@@ -3,3 +3,409 @@
...
+
+Ideal Gas
+`````````
+
+The ideal gas (aka perfect or gamma-law gas) model in ``singularity-eos`` takes
+the form
+
+.. math::
+
+    P = \Gamma \rho \hat{e}
```
Yurlungur, line 130: Is this just a me thing? I think of specific energy as \varepsilon, not \hat{e}. But I'm willing to be overruled. I don't think it matters much.

doc/sphinx/src/models.rst
```diff
@@ -3,3 +3,409 @@
...
++--------------------------+----------------------+------------------------+
+| :math:`C_V(\rho, T)`     | Constant Volume      | Density, Temperature   |
++--------------------------+ Specific Heat        +------------------------+
+| :math:`C_V(\rho, E)`     | Capacity             | Density, Energy        |
++--------------------------+----------------------+------------------------+
+| :math:`B_S(\rho, T)`     | Isentropic Bulk      | Density, Temperature   |
++--------------------------+ Modulus              +------------------------+
+| :math:`B_S(\rho, E)`     |                      | Density, Energy        |
++--------------------------+----------------------+------------------------+
+| :math:`\Gamma(\rho, T)`  | Gruneisen Parameter  | Density, Temperature   |
```
Yurlungur, line 85: Hmm using `\Gamma` as the Gruneisen parameter I think is going to be confusing for hydro folks.  Maybe `\Gamma_1` (or whatever subscript it actually is?) might be better. Or alternatively, just use `(\Gamma - 1)` everywhere?

doc/sphinx/src/models.rst
Yurlungur: As I mention in another comment, I think using `\Gamma` here is a little confusing since I think most hydro folks will expect to see
```
P = (\Gamma - 1)rho \varepsilon
```

doc/sphinx/src/models.rst
```diff
@@ -3,3 +3,409 @@
...
+    P_H(\rho) = \rho_0 c_0^2 \mu
+      \begin{cases}
+        1                                                 & \rho < \rho_0 \\
+        \frac{1 + \left(1 - \frac{1}{2}\Gamma_0 \right)\mu - \frac{b}{2} \mu^2}
+          {\left(1 - (s_1 - 1)\mu + s_2 \frac{\mu^2}{1 + \mu}
+            - s_3 \frac{\mu^3}{(1+\mu)^2} \right)^2}      & \rho > \rho_0
+      \end{cases}
+
+where :math:`c_0`, :math:`s_1`, :math:`s_2`, and :math:`s_3` are fitting
+paramters. The units of :math:`c_0` are velocity while the rest are unitless.
```
Yurlungur, line 205: Perhaps worth commenting that the exact fitting form is "model implementation dependent" and our fitting formula may not look like the formula in another code.

doc/sphinx/src/models.rst
```diff
@@ -3,3 +3,409 @@
...
+:math:`a`, :math:`b`:, :math:`k`, and :math:`n`, while :math:`P_\mathrm{C}`,
+:math:`E_\mathrm{C}`, :math:`V_\mathrm{C}` and :math:`T_\mathrm{C}` are tuning
+parameters with units related to their non-subscripted counterparts.
+
+
+Spiner EOS
+````````````
+
+Stellar Collapse EOS
+````````````````````
```
Yurlungur, line 407: Guess it's on me to do these ones. :D

doc/sphinx/src/models.rst
```diff
@@ -3,3 +3,409 @@
...
+
+EOSPAC EOS
+````````````
```
Yurlungur, line 410: @annapiegra do you want to take a stab at this in the same style as above? I can also do it. (Either way we can submit in a different PR. I'd like to get @jhp-lanl 's contribution in. :)

doc/sphinx/src/models.rst
jhp-lanl: For most equations of state it will be a simple add as well. I'm happy to code up the actual isentropes if you create all the plumbing.

doc/sphinx/src/models.rst
jhp-lanl: I'll change this to internal energy to differentiate.

doc/sphinx/src/models.rst
Yurlungur: It's on my to do list, as it's needed for my NGP work. :D

doc/sphinx/src/models.rst
jhp-lanl: You're going to need to define some quantities there because as an EOS person I don't understand `\Gamma_1` and if your `\Gamma` is the adiabatic exponent, I'm pretty sure that the Gruneisen parameter is only equal to `\Gamma - 1` for an ideal gas. While I have seen both lowercase `\gamma` as well as uppercase `\Gamma` for the Gruneisen parameter, I think outside of astrophysics the adiabatic exponent is almost exclusively represented by small `\gamma` so `\Gamma` seems logical for the Gruneisen parameter.

Find me some resources defining the parameters you mention and I'll consider it. But I think changing it will make it confusing for the EOS folks.

doc/sphinx/src/models.rst
jhp-lanl: I'll switch everything to lowercase e as I agree that upper-case E is usually total energy.

jhp-lanl: The ideal gas law was a mistake in the nomenclature that has been brought into consistency with the rest of the document

doc/sphinx/src/models.rst
jhp-lanl: Are you referring to the fact that this looks a lot like a polynomial fit to the Us-up curve?

doc/sphinx/src/models.rst
jhp-lanl: That sounds good. I just wanted to break things up and get _something_ in on this section to lay the groundwork.

doc/sphinx/src/models.rst
jhp-lanl: Yes it was a mistake for energy to be extensive. That has been fixed.

doc/sphinx/src/models.rst
Yurlungur: I can't find the reference for \Gamma_1 at the moment.

How about a compromise: use lower case (\gamma-1) in the ideal gas formula and say that in the special case of an ideal gas, this is equal to the Gruneisen parameter.

Also note that the Gruneisen parameter is referred to as gm1 throughout the implementation of singularity-eos for this reason.

doc/sphinx/src/models.rst
Yurlungur: Basically yes. In other codes I've seen different formulae here with different numbers of "s" coefficients. It has to do with the order of the fit.

doc/sphinx/src/models.rst
jhp-lanl: I'll note that `\Gamma = \gamma - 1` for an ideal gas.

Yes I've seen that in the code, but I think the `\Gamma_1` notation is specifically for astrophysics. The only reference I've found has been here:
https://en.wikipedia.org/wiki/Heat_capacity_ratio#Adiabatic_process.

Aside from E being internal energy, I also think it's good to follow the nomenclature of Menikoff and Plohr since I would consider it a seminal paper at the intersection of EOS and hydrodynamics. They use `\gamma` for the adiabatic exponent and `\Gamma` for the Gruneisen parameter as well.

I think it would be very helpful to add a disambiguation section where we describe the different Gammas and how they are related. If you can find me something that documents your understanding, I'll put it in that section. I'll start with a note about `\gamma` vs `\Gamma` there as well.

jhp-lanl: Menikoff and Plohr for reference:
https://journals.aps.org/rmp/abstract/10.1103/RevModPhys.61.75

Yurlungur: I think disambiguating across all communities is kind of a lost cause. I'm happy enough with the notation with the big and little gammas. I think with the table that's clear enough.

I don't think tying ourselves to the notation of a single paper is wise, but you can mention menikhoff and plohr and your primary source notation wise if you like.

Maybe the better disambiguation we should add is a translation between code and math notation. I.e., the Gruneisen thing. And that energy is often written as sie in the code. Things like that.

jhp-lanl: Well, I wouldn't mind defining multiple Gammas if they're commonly used as you said. My first encounter with them was talking with you and Josh so I'd appreciate knowing what other notations are out there.

doc/sphinx/src/models.rst
jhp-lanl: Added a clarifying sentence

doc/sphinx/src/models.rst
jhp-lanl: As for relating code variables to thermodynamic quantities, I think the API section is a better place for that.

jhp-lanl: @Yurlungur is there anything else you'd like me to do to address this issue?

Yurlungur (approved):
I'm happy with this now. Let's get it merged in. Thanks for all your work on this, @jhp-lanl !

PR #75, opened 2022-01-25
First pass at vector EOS interface

jhp-lanl:
## PR Summary

First attempt at creating a vector interface for the EOS classes. This is very much a WIP and open for discussion.

@dholladay00 and @Yurlungur please let me know what you think

### To do:
- [x] Create default vector interface
- [x] Create unit tests to cover new interface
- [x] Run tests on Darwin

## PR Checklist


singularity-eos/eos/eos.hpp
```diff
@@ -88,6 +122,144 @@ class IdealGas {
...
+  /*
+  Vector versions of the member functions are only available for CPU
+  
+  RealIndexer must have an operator[](int) that returns a Real. e.g., Real*
+  ConstRealIndexer is as RealIndexer, but assumed const type.
+  LambdaIndexer must have an operator[](int) that returns a Real*. e.g., Real**
+  */
+  template<typename RealIndexer, typename ConstRealIndexer, typename LambdaIndexer>
+  inline
+  void TemperatureFromDensityInternalEnergyVector(ConstRealIndexer &&rhos,
```
Yurlungur, line 134: No reason to name these with the `Vector` postfix. Just let multiple dispatch handle it. Because the arguments are different, the functions will be different.

singularity-eos/eos/eos.hpp
```diff
@@ -88,6 +122,144 @@ class IdealGas {
...
                               Real *lambda = nullptr) const;
 
+  /*
+  Vector versions of the member functions are only available for CPU
+  
+  RealIndexer must have an operator[](int) that returns a Real. e.g., Real*
+  ConstRealIndexer is as RealIndexer, but assumed const type.
+  LambdaIndexer must have an operator[](int) that returns a Real*. e.g., Real**
+  */
+  template<typename RealIndexer, typename ConstRealIndexer, typename LambdaIndexer>
```
Yurlungur, line 132: In principle, `rhos` and `sies` could be different `Indexer` types.

singularity-eos/eos/eos.hpp
```diff
@@ -46,6 +46,40 @@
...
 
+
+/*
+  Wrapper for scalar lookups to allow them to take vectors on the CPU
+  
+  RealIndexer must have an operator[](int) that returns a Real. e.g., Real*
+  ConstRealIndexer is as RealIndexer, but assumed const type.
+  LambdaIndexer must have an operator[](int) that returns a Real*. e.g., Real**
+  */
+template<typename EOS, typename LookupFunction, typename ConstRealIndexer,
```
Yurlungur, line 57: I suppose in principle `xs` and `ys` could be different indexer types.

singularity-eos/eos/eos.hpp
```diff
@@ -154,6 +326,146 @@ class Gruneisen {
   void ValuesAtReferenceState(Real &rho, Real &temp, Real &sie, Real &press, Real &cv,
                               Real &bmod, Real &dpde, Real &dvdt,
                               Real *lambda = nullptr) const;
+
+  /*
+  Vector versions of the member functions are only available for CPU
+  
+  RealIndexer must have an operator[](int) that returns a Real. e.g., Real*
+  ConstRealIndexer is as RealIndexer, but assumed const type.
+  LambdaIndexer must have an operator[](int) that returns a Real*. e.g., Real**
```
Yurlungur, line 335: perhaps worth commenting that unlike the scalar versions of the functions, the lambda indexer is not optional here.

We could make it optional by adding an overload with the lambdas missing and a dummy lambda indexer passed in perhaps.

singularity-eos/eos/eos_variant.hpp
```diff
@@ -229,6 +230,249 @@ class Variant {
...
+  }
+
+  template<typename RealIndexer, typename LambdaIndexer>
+  inline
+  void PTofREVector(RealIndexer &&rhos, RealIndexer &&sies, const int num,
+                    LambdaIndexer &&lambdas, RealIndexer &&presses,
+                    RealIndexer &&temps, RealIndexer &&dpdrs,
+                    RealIndexer &&dpdes, RealIndexer &&dtdrs,
+                    RealIndexer &&dtdes) const {
+    return mpark::visit(
```
Yurlungur, line 412: Hmm... I'm not sure how I feel about re-implementing this whole thing here. Maybe it's necessary, I dunno. This whole `PTofRE` machinery might be better being refactored to reduced code duplication, and to enable individual EOSs to implement their own version if desired. (Similar to how you did the `Vector` loop abstractions.)

singularity-eos/eos/eos.hpp
```diff
@@ -46,6 +46,40 @@
...
+  */
+template<typename EOS, typename LookupFunction, typename ConstRealIndexer,
+         typename RealIndexer, typename LambdaIndexer>
+inline
+void VectorizeScalarLookup(const EOS &eos, const LookupFunction &&f,
+                             const int num, ConstRealIndexer &&xs,
+                             ConstRealIndexer &&ys, RealIndexer &&result,
+                             LambdaIndexer &&lambdas) {
+  for (int i = 0; i < num; ++i) {
```
Yurlungur, line 64: This should be a `portableFor`. Look in ports of call for the syntax.

singularity-eos/eos/eos.hpp
```diff
@@ -46,6 +46,40 @@
...
+
+template<typename EOS, typename FillEosFunction, typename RealIndexer,
+         typename LambdaIndexer>
+inline
+void VectorizeFillEos(const EOS &eos, const FillEosFunction &&f,
+                      const int num, RealIndexer &&rhos, RealIndexer &&temps,
+                      RealIndexer &&energies, RealIndexer &&presses,
+                      RealIndexer &&cvs, RealIndexer &&bmods,
+                      const unsigned long output, LambdaIndexer &&lambdas) {
+  for (int i = 0; i < num; ++i) {
```
Yurlungur, line 77: `portableFor`.

singularity-eos/eos/eos_variant.hpp
```diff
@@ -229,6 +230,249 @@ class Variant {
...
+          // Lookup at density and energy
+          eos.PressureFromDensityInternalEnergyVector(rhos, sies, num, presses,
+                                                      lambdas);
+          eos.TemperatureFromDensityInternalEnergyVector(rhos, sies, num, temps,
+                                                         lambdas);
+          // Peturbation factor
+          Real factor = 1. + 1.0e-06;
+
+          // Perturb densities and do lookups
+          for (int i = 0; i < num ; i ++) {
```
Yurlungur, line 425: `portableFor`

singularity-eos/eos/eos_variant.hpp
```diff
@@ -229,6 +230,249 @@ class Variant {
...
+            rhos[i] *= factor;
+          }
+          eos.PressureFromDensityInternalEnergyVector(rhos, sies, num, dpdrs,
+                                                      lambdas);
+          eos.TemperatureFromDensityInternalEnergyVector(rhos, sies, num, dtdrs,
+                                                         lambdas);
+          
+          // Reset densities, perturb energies, and do lookups
+          for (int i = 0; i < num ; i ++) {
```
Yurlungur, line 434: `portableFor`

singularity-eos/eos/eos_variant.hpp
```diff
@@ -229,6 +230,249 @@ class Variant {
...
+            sies[i] *= factor;
+            rhos[i] /= factor;
+          }
+          eos.PressureFromDensityInternalEnergyVector(rhos, sies, num, dpdes,
+                                                      lambdas);
+          eos.TemperatureFromDensityInternalEnergyVector(rhos, sies, num, dtdes,
+                                                         lambdas);
+
+          // Reset the energies to their original values
+          for (int i = 0; i < num ; i ++) {
```
Yurlungur, line 444: `portableFor`

singularity-eos/eos/eos_variant.hpp
```diff
@@ -229,6 +230,249 @@ class Variant {
...
+            sies[i] /= factor;
+          }
+
+          // Calculate the derivatives
+          for(int i = 0; i < num ; i ++){
```
Yurlungur, line 449: `portableFor`

Yurlungur (changes requested):
This is a great start! Thanks for your work on this, @jhp-lanl ! A left a few organizational comments below. There's one broad thing (also discussed below) that I'd like to highlight. Mainly: the vector function calls should be called from host, but they should run on device. This means that using for loops isn't the way to go. We need to use parallel dispatch. In ports of call, we have an abstraction for this. It looks something like this:
```C++
portableFor("loop name", loop_start, loop_end, PORTABLE_LAMBDA(int i) { // loop body });
```
which wraps either an appopriate `Kokkos` call or a raw for loop depending on how the code was built. Loop bounds follow C convention, start is inclusive, end is exclusive.

Take a look at `ports-of-call/portability.hpp` for more details.

jhp-lanl:
> There's one broad thing (also discussed below) that I'd like to highlight. Mainly: the vector function calls should be called from host, but they should run on device.

Ah okay I misunderstood you and @dholladay00 then. I thought you wanted this to be a CPU-only thing top to bottom, not that the vector functions should be host and the loops on the device. I thought about that while I was writing this but I just kept them as regular for loops for the sake of simplicity. Easy enough to change though!

I'll start addressing the individual comments here shortly.

singularity-eos/eos/eos_variant.hpp
jhp-lanl: I re-implemented it simply because I was adapting what was already there. I tried to mirror the existing implementation as closely as I could just with arrays.

I think my main problem here is that different EOS may want to calculate derivatives differently and the 1e-06 factor for the perturbation is just an arbitrary number that may or may not be sufficient for all situations. I'd propose keeping the functionality similar between the scalar and vector versions of these functions, but I'm otherwise open to suggestions for to change this.

singularity-eos/eos/eos.hpp
```diff
@@ -46,6 +46,40 @@
...
+    (eos.*f)(rhos[i], temps[i], energies[i], presses[i], cvs[i], bmods[i],
+              output, lambdas[i]);
+  }
+}
+
```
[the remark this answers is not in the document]
jhp-lanl: Sounds good

singularity-eos/eos/eos.hpp
jhp-lanl: Good point. I'd prefer to mirror the functionality of the scalar functions as much as possible so I'll overload the function

Yurlungur:
> 

Yeah it should be pretty straightforward to change.

singularity-eos/eos/eos_variant.hpp
Yurlungur: My suggestion is we should see if we can make a generic implementation that we call from each EOS, then this can be overloaded as needed.

But we can deal with that in a later PR.

singularity-eos/eos/eos.hpp
jhp-lanl: What circumstances do you think would lead to input indexers having different types? Assuming a collection of states, I'd assume that the input variables would necessarily use the same indexing scheme to travel over the collection of states.

Yurlungur: Yeah you're probably right, so we can leave it for now. I was thinking of some scenario where rho and sie were sparsely populated in different ways somehow.

jhp-lanl: Let's revisit if we find a circumstance where this occurs though. It will be fairly straightforward to change

jhp-lanl:
To do:

- [ ]  Overload member functions to allow for optional specification of lambda
- [ ]  Add vector lookups to EOSPAC
- [ ]  Add unit tests of vector basic vector functionality

singularity-eos/eos/vector_utils.hpp
```diff
@@ -0,0 +1,74 @@
...
+LambdaIndexer must have an operator[](int) that returns a Real*. e.g., Real**
+*/
+template<typename EOS, typename LookupFunction, typename ConstRealIndexer,
+         typename RealIndexer, typename LambdaIndexer>
+inline
+void VectorizeScalarLookup(const EOS &eos, const LookupFunction &&f,
+                           ConstRealIndexer &&xs, ConstRealIndexer &&ys,
+                           RealIndexer &&result, const int num,
+                           LambdaIndexer &&lambdas) {
+  portableFor(
```
dholladay00, line 49: @jhp-lanl perhaps adding a `const char*` for a function name may be helpful here. Another option I've been toying with is to grab additional info from the template parameters, but getting the name of the specific function passed in so far has been challenging. I propose:
- passing in the string (lots of work added, but provides control and lots of information)
- using template param(s) (little work, but less specific, not sure if typeid overhead is perceived on every function call)

Yurlungur: I like the string solution better. Shouldn't be too bad.

dholladay00: I wrote an example of how a macro could be used that may be less objectionable to @jhp-lanl but still provide the specificity we need:
```c++
template<typename T, typename U>
void print_name(const T& a, U&& b, const char* s) {
    printf("%s\n", s);
}

#define PRINT_NAME(a, b) \
print_name(a, b, #b)

int main() {
    PRINT_NAME(gruneisen{}, &gruneisen::funcone); // string is "&gruneisen::funcone"
    return 0;
}
```

Yurlungur: Nice.

jhp-lanl: Sorry I'm a little confused. What did you want me to do with the function name?

jhp-lanl: Resolving because I changed everything lol

jhp-lanl:
@dholladay00 and @Yurlungur I've redone everything with the CRTP design instead. I like this a lot better but as they say, the proof is in the pudding (unit tests). But in the meantime, let me know if you'd like anything changed.

singularity-eos/eos/eos.hpp
```diff
@@ -90,82 +90,18 @@ class IdealGas : public EosBase<IdealGas> {
...
-                                               RealIndexer &&gm1s,
-                                               const int num,
-                                               LambdaIndexer &&lambdas) const;
-  template<typename RealIndexer, typename LambdaIndexer>
-  inline
-  void FillEos(RealIndexer &&rhos, RealIndexer &&temps, RealIndexer &&energies,
-               RealIndexer &&presses, RealIndexer &&cvs, RealIndexer &&bmods,
-               const int num, const unsigned long output,
-               LambdaIndexer &&lambdas) const;
+  // Vector functions that overload the scalar versions declared here
```
dholladay00, line 93: could use a macro, I know preprocessor, but it could really decrease SLOC and prevent spelling errors:
```
#define SG_ADD_USINGS(MOD) \
using EosBase<MOD>::TemperatureFromDensityInternalEnergy;\
using EosBase<MOD>::InternalEnergyFromDensityTemperature;\
using EosBase<MOD>::PressureFromDensityTemperature;\
using EosBase<MOD>::PressureFromDensityInternalEnergy;\
using EosBase<MOD>::SpecificHeatFromDensityTemperature;\
using EosBase<MOD>::SpecificHeatFromDensityInternalEnergy;\
using EosBase<MOD>::BulkModulusFromDensityTemperature;\
using EosBase<MOD>::BulkModulusFromDensityInternalEnergy;\
using EosBase<MOD>::GruneisenParamFromDensityTemperature;\
using EosBase<MOD>::GruneisenParamFromDensityInternalEnergy;\
using EosBase<MOD>::FillEos;
```

jhp-lanl: That's a good idea.

I'm not completely _against_ using preprocessor macros ;-) . I just don't like when they sacrifice code readability for minor gains. In this case I think the code maintenance boon is very significant and the code readability isn't sacrificed.

singularity-eos/eos/eos.hpp
```diff
@@ -1015,7 +1000,7 @@ class StellarCollapse : EosBase<StellarCollapse> {
 #ifdef SINGULARITY_USE_EOSPAC
 // Only really works in serial
 // Not really supported on device
-class EOSPAC : EosBase<EOSPAC> {
+class EOSPAC : public EosBase<EOSPAC> {
```
dholladay00, line 1003: Should eospac be inheriting from the base class since it's not going to be using the base class helper functions?

jhp-lanl: It shouldn't and I plan to change that. I added that as a quick workaround to get things to compile at this stage while I work on getting EOSPAC itself working in a separate way

jhp-lanl: So in short, I plan to keep the inheritance but only use the `using` statements when needed

jhp-lanl: But that reminds me that I need to actually change the scalar version of the member function in the `Variant` class... whoops 😅

jhp-lanl:
For documentation purposes:

I was having troubles with the `ctest` framework where `ctest -V` would cause the vector test to timeout. Without the `-V` flag though, the tests work as intended.

I suspect there might be some undefined behavior occurring with the `INFO` or `REQUIRE` macros (or an interaction between the two) that is fixed when the `REQUIRE` statements are changed to `CHECK` statements.

jhp-lanl:
> 



> For documentation purposes:
> 
> I was having troubles with the `ctest` framework where `ctest -V` would cause the vector test to timeout. Without the `-V` flag though, the tests work as intended.
> 
> I suspect there might be some undefined behavior occurring with the `INFO` or `REQUIRE` macros (or an interaction between the two) that is fixed when the `REQUIRE` statements are changed to `CHECK` statements.

It turns out that the problem was that I had an uninitialized loop counter. Catch seems to work fine so I've removed the comments that made it seem like the issue was there.

jhp-lanl:
Okay with fixes from @dholladay00 and @jonahm-LANL all of the tests are now running and passing on Darwin Power9.

Let's do a final review tomorrow and then get this merged.

FYI - I'm going to focus on EOSPAC in a separate effort.

Yurlungur:
Great work, @jhp-lanl ! Yes, I'll do a careful review of this tomorrow.

singularity-eos/eos/eos_variant.hpp
```diff
@@ -223,12 +223,333 @@ class Variant {
   void DensityEnergyFromPressureTemperature(const Real press, const Real temp,
                                             Real *lambda, Real &rho, Real &sie) const {
     return mpark::visit(
-        [&press, &temp, &lambda, &rho, &sie](const auto &eos) {
+        [&press, &temp, lambda, &rho, &sie](const auto &eos) {
```
dholladay00, line 226: This change should probably be reverted to be consistent with the other lambdas.

jhp-lanl: Fixed

singularity-eos/eos/eos_base.hpp
```diff
@@ -0,0 +1,330 @@
...
+using EosBase<EOSDERIVED>::SpecificHeatFromDensityInternalEnergy;\
+using EosBase<EOSDERIVED>::BulkModulusFromDensityTemperature;\
+using EosBase<EOSDERIVED>::BulkModulusFromDensityInternalEnergy;\
+using EosBase<EOSDERIVED>::GruneisenParamFromDensityTemperature;\
+using EosBase<EOSDERIVED>::GruneisenParamFromDensityInternalEnergy;\
+using EosBase<EOSDERIVED>::PTofRE;\
+using EosBase<EOSDERIVED>::FillEos;
+
+/*
+This is a CRTP that allows for static inheritance so that default behavior for
```
Yurlungur, line 48: Probably worth defining CRTP in this comment and pointing to a wikipedia article or something.

test/CMakeLists.txt
```diff
@@ -31,7 +31,7 @@ if(SINGULARITY_USE_EOSPAC AND SINGULARITY_USE_HDF5 AND SINGULARITY_TEST_SESAME)
                              ${CMAKE_SOURCE_DIR}/utils/ports-of-call
                              ${CMAKE_SOURCE_DIR}/utils/sesame2spiner)
 endif()
-catch_discover_tests(eos_unit_tests)
+catch_discover_tests(eos_unit_tests PROPERTIES TIMEOUT 60)
```
Yurlungur, line 34: What's this timeout for?

singularity-eos/eos/eos.hpp
```diff
@@ -154,6 +159,8 @@ class Gruneisen {
   void ValuesAtReferenceState(Real &rho, Real &temp, Real &sie, Real &press, Real &cv,
                               Real &bmod, Real &dpde, Real &dvdt,
                               Real *lambda = nullptr) const;
+  // Vector functions that overload the scalar versions declared here
+  SG_ADD_BASE_CLASS_USINGS(Gruneisen)
```
Yurlungur, line 163: might be worth adding a comment that this contains any generic functions provided by the base class.

dholladay00:
pending CI and formatting, I think this is good to go!

Yurlungur (changes requested):
This is a tremendous effort, @jhp-lanl ! Great work! I like the the base class solution a lot and I wonder if we can leverage it for other shared capabilities. We should think about that going forward to reduce code complexity.

Some mostly minor nitpicks below. Can you make sure to run clang-formatting before merge as well?

Also it might be good to add some comments about the design into the docs. We should maybe add a docs page for developers, explaining some of our design decisions. The compile time inheritence thing seems like the perfect place to start there. What do you think?

singularity-eos/eos/eos_base.hpp
```diff
@@ -0,0 +1,330 @@
+//------------------------------------------------------------------------------
+// © 2021. Triad National Security, LLC. All rights reserved.  This
```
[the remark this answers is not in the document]
jhp-lanl: Since this needs to be applied to every file in the repo, I'd like to just do this in a separate MR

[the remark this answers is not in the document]
Yurlungur: I'll make an issue.

test/CMakeLists.txt
jhp-lanl: When I was dealing with undefined behavior, the loops would execute for 1500 seconds. While I could CTRL+C the tests, it didn't let the tests end to give me a clue about what was going on. Adding this timeout just makes it so each test has 60 seconds to complete instead of the default 1500.

At worst, this change does nothing at the moment and may need to be changed if somebody introduces a really long test. At best though, it will help somebody debug faster in the future if a test is misbehaving.

singularity-eos/eos/eos.hpp
```diff
@@ -1053,6 +1131,7 @@ class EOSPAC {
   EOS_INTEGER EofRT_table_;
   EOS_INTEGER RofPT_table_;
   EOS_INTEGER TofRP_table_;
+  // EOS_INTEGER PofRE_table_;
```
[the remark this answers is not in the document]
jhp-lanl: It's part of the EOSPAC vector work. I'll add a TODO

test/CMakeLists.txt
Yurlungur: ok seems fine for now.

singularity-eos/eos/eos_base.hpp
```diff
@@ -0,0 +1,330 @@
...
+#define _SINGULARITY_EOS_EOS_EOS_BASE_
+
+#include <string>
+
+#include <ports-of-call/portability.hpp>
+
+// Macro for portableFor naming
+#define SG_MEMBER_FUNC_NAME() \
+  (std::string(typeid(CRTP).name()) \
+    + std::string("::") + std::string(__func__))
```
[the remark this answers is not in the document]
Yurlungur: oops `member_func_name` needs to be marked static.

singularity-eos/eos/eos_base.hpp
```diff
@@ -0,0 +1,314 @@
...
+  //         dtdes[i] = (dtdes[i] - temps[i]) / (sies[i] * (1. - factor));
+  //       }
+  //   );
+
+  //   return;
+  // }
+};
+} // namespace eos_base
+} // namespace singularity
+
```
dholladay00, line 313: not necessary but could add an `#undef SG_MEMBER_FUNC_NAME` for completeness.

singularity-eos/eos/eos.hpp
```diff
@@ -1028,13 +1026,93 @@ class EOSPAC {
...
-  PORTABLE_FUNCTION void PTofRE(const Real rho, const Real sie, Real *lambda, Real &press,
-                                Real &temp, Real &dpdr, Real &dpde, Real &dtdr,
-                                Real &dtde) const;
   PORTABLE_FUNCTION void ValuesAtReferenceState(Real &rho, Real &temp, Real &sie,
                                                 Real &press, Real &cv, Real &bmod,
                                                 Real &dpde, Real &dvdt,
                                                 Real *lambda = nullptr) const;
+  // Vector functions for EOSPAC do not rely on the base class so they are
+  // declared here.
+  // Vector functions that overload the scalar versions declared here
```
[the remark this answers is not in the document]
jhp-lanl: Done

singularity-eos/eos/eos.hpp
jhp-lanl: Done

test/eos_unit_tests.cpp
```diff
@@ -57,15 +58,25 @@ constexpr Real ev2k = 1.160451812e4;
 #endif // SINGULARITY_TEST_SESAME
 #endif // SPINER_USE_HDF
 
-constexpr Real EPS = 5e-2; // within a few percent
-PORTABLE_INLINE_FUNCTION bool isClose(Real a, Real b) {
+PORTABLE_INLINE_FUNCTION bool isClose(Real a, Real b, Real EPS=5e-2) {
   return fabs(b - a) / (fabs(a + b) + 1e-20) <= EPS;
```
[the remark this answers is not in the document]
jhp-lanl: Done

singularity-eos/eos/eos_base.hpp
jhp-lanl: Added the Wiki article

singularity-eos/eos/eos_base.hpp
Yurlungur: I am in favor of this.

Yurlungur (approved):
Approved assuming you add the `#undef` that @dholladay00 suggested. I think it would be good to add some developer documentation about this, but it doesn't have to be here---I leave that to your discretion.

Let me know when you're ready for merge @jhp-lanl .

singularity-eos/eos/eos_base.hpp
jhp-lanl: Sure. Added a `#undef SG_ADD_BASE_CLASS_USINGS` as well.

Yurlungur: I don't think you can `#undef` `SG_ADD_BASE_CLASS_USINGS` because it's needed in `eos.hpp` and other header files right?

jhp-lanl: I assumed the @dholladay00 meant adding an `#undef` right before the `#define` so I just did the same thing for `SG_ADD_BASE_CLASS_USINGS`. Did I misunderstand?

The `#undef` should just make ensure there aren't any name conflicts right?

jhp-lanl: I'll assume that @dholladay00 wanted the `#undef` to go at the end of the file then and I'll remove both of these since they should probably be wrapped by `#ifdef`.

jhp-lanl: Done

PR #82, opened 2022-02-16
Add eospac variant to spackage

annapiegra:
## PR Summary

Make eospac a variant in singularity-eos spackage. (Eospac was already listed as a dependency)

Spack looks for eospac in the spec to determine if eospac should be pulled in:           self.define("SINGULARITY_USE_EOSPAC", "^eospac" in self.spec)

## PR Checklist

Spack env builds on CTS1_intel. 

### DIFF
--- a/spack-repo/packages/singularity-eos/package.py
+++ b/spack-repo/packages/singularity-eos/package.py
@@ -39,6 +39,7 @@ class SingularityEos(CMakePackage, CudaPackage):
 
     # TODO: do we always depend on eospac? 
     depends_on("eospac")
+    variant("eospac", default=True, description="Pull in EOSPAC")
 
     # building/testing/docs
     depends_on("cmake@3.14:")
--- a/spack-repo/packages/singularity-eos/package.py
+++ b/spack-repo/packages/singularity-eos/package.py
@@ -37,8 +37,6 @@ class SingularityEos(CMakePackage, CudaPackage):
     # TODO: decide if this should stay in the source tree, or split out a seperate dependency build
     #depends_on("mpark-variant")
 
-    # TODO: do we always depend on eospac? 
-    depends_on("eospac")
     variant("eospac", default=True, description="Pull in EOSPAC")
 
     # building/testing/docs
@@ -52,6 +50,8 @@ class SingularityEos(CMakePackage, CudaPackage):
 
     # linear algebra when not using GPUs
     depends_on("eigen@3.3.8", when="~cuda")
+
+    depends_on("eospac", when="+eospac")
 
     # set up kokkos offloading dependencies
     for _flag in ("~cuda", "+cuda", "~openmp", "+openmp"):


spack-repo/packages/singularity-eos/package.py
```diff
@@ -39,6 +39,7 @@ class SingularityEos(CMakePackage, CudaPackage):
 
     # TODO: do we always depend on eospac? 
     depends_on("eospac")
```
Yurlungur, line 41: Should this statement turn into a `when`? I think it should only depend on `eospac` when the variant is enabled.

Yurlungur (changes requested):

mauneyc-LANL:
@dholladay00 @annapiegra @Yurlungur ,

I would prefer that the use-case be made clear in the interface. Here, the variant would be (as example, could be different)

`variant("eospac_wrapper", default=True, description="build the EOSPAC wrapper")`

More broadly (and this is probably a separate PR), the CMake has a lot of overlapping inter-dependencies that make it difficult to sort out a clean dependency trees. For example, at CMakeLists.txt:333

```
if(SINGULARITY_BUILD_SESAME2SPINER)
  if(SINGULARITY_USE_EOSPAC)
    if(SINGULARITY_USE_HDF5)
```
Now, the spack spec `singularity-eos~eospac build_extra=sesame` is valid, but due to the logic of the build chain `sesame2spiner` will not be built. 

Further, in nearly all cases where `SINGULARITY_USE_EOSPAC` is checked to alter some extra compiler work, it is used in conjunction with `SINGULARITY_USE_HDF5`, which implies that if we depend on `EOSPAC`, we depend on `HDF5` for any meaningful use-case. However, it is explicitly allowed (though warned) to enable `EOSPAC` without linking to `HDF5`.

This was my original dilemma is writing the `package.py` files; What would be ideal (from my point-of-view) is to enforce a cleaner dependency structure. For example, if I wanted to build ``singularity-eos build_extra=sesame`, this would automatically give `EOSPAC` and `HDF5` as dependencies:

```
# ensure hdf5 and eospac are installed together
depends_on('hdf5', when='+eospac_wrapper')
# do not allow sesame2spiner to be built if there is no eospac dependency
conflicts('~eospac_wrapper', when='build_extra=sesame')
```

The issue then is should the dependency range implied by the current CMake logic what we want, or is there a more "correct" dependency graph.

spack-repo/packages/singularity-eos/package.py
mauneyc-LANL: Correct, and it would be cleaner for the variant to be above the depends.

```
variant("eospac", default=True, description="Pull in EOSPAC")
# maybe some other variants...
```

Yurlungur:
I think it is possible to want eospac but not hdf5 if, for example, the eospac backend is used without spiner.

mauneyc-LANL:
Updated package.py to use eospac variant @annapiegra @Yurlungur

dholladay00:
Does the cmake options file that's generated need to be updated or is that automagically done @mauneyc-LANL ?

mauneyc-LANL:
> Does the cmake options file that's generated need to be updated or is that automagically done @mauneyc-LANL ?

we have 

```
 self.define("SINGULARITY_USE_EOSPAC", "^eospac" in self.spec)
```

as the cmake option, afaik that's the only one

spack-repo/packages/singularity-eos/package.py
dholladay00: has this been resolved?

Yurlungur: yes

mauneyc-LANL: @Yurlungur this still looks labeled as unresolved

Yurlungur: I no longer remember the context of this... What's the issue? (It certainly could still be unresolved.)

Yurlungur:
@annapiegra is this still needed? Or has this been resolved?

annapiegra:
> @annapiegra is this still needed? Or has this been resolved?
Some version of this is needed. I think @danl and @mauneyc-LANL understand the dependencies better.

mauneyc-LANL (approved):

Yurlungur (approved):

Yurlungur:
@mauneyc-LANL @annapiegra should I merge this? I see we have approvals

annapiegra:
> > @mauneyc-LANL @annapiegra should I merge this? I see we have approvals
> 
> This looks fine to me, if @annapiegra is okay with it

Fine with me.

Yurlungur:
Merging then

PR #84, opened 2022-02-23
cmake: fix logic for internal kokkos

junghans:
It is currently impossible to use the bundled kokkos as `find_package(KOKKOS REQUIRED)` will always fail without an external kokkos.


mauneyc-LANL (approved):

Yurlungur (approved):
Oops. Nice find!

Issue #93, opened 2022-03-17
Document class strategy


jhp-lanl:
In a developer guide, document the strategy used for static inheritance and why we went with it.

PR #96, opened 2022-03-17
Add to "Contributing" section of docs

jhp-lanl:
## PR Summary

Adds information about the CRTP to the "Contributing" section of the sphinx documentation

I also added a placeholder for somebody to talk about `mpark::variant` and the `Variant` class.

Fixes https://github.com/lanl/singularity-eos/issues/93

## PR Checklist

- [N/A] Adds a test for any bugs fixed. Adds tests for new features.
- [N/A] Format your changes by using the `make format` command after configuring with `cmake`.


jhp-lanl:
@Yurlungur and @dholladay00 let me know what you think. Just a reminder again, I haven't been programming in C++ for long so I may not understand all of the subtleties here. Feel free to suggest minor changes or just push commits yourself for more major changes

doc/sphinx/src/contributing.rst
```diff
@@ -2,3 +2,120 @@
...
+order to centralize default functionality and avoid code duplication. The two
+main examples of this are the vector overloads and the ``PTofRE`` scalar lookup
+function. In the vector overloads, a simple for loop is used to loop over the
+set of states provided to the function and then call the scalar version on each
+state. The ``PTofRE`` function is designed to provide a common method for
+getting the needing information for a PTE solve from an EOS. Both of these
+features are not dependent on the specific EOS for their definition, but in the
+case of the vector overloads, they *do* need to access methods in the derived
+class.
+
```
Yurlungur, line 30: I'd add a comment like:

> In both cases, these functions have default behavior that may need to be overridden for a given equation of state.

doc/sphinx/src/contributing.rst
```diff
@@ -2,3 +2,120 @@
...
+correct EOS implementation.
+
+The challenge in making this sort of code performance portable is that on a CPU
+this sort of type deduction (i.e. what specific EOS implementaiton to call)
+would be done at runtime through vtables. In a heterogenous computing enviroment
+though, the vtable may point to member functions that do not exist on the
+device, causing a segfault. We could have used a similar technique to the
+modifier classes and pass the EOS as a template paramter, but then the vector
+function calls could only be achieved by creating vector modifiers of all the
+implemented EOS.
```
Yurlungur, line 56: The issue isn't really segfaults. Runtime inheritance is possible on GPU, just cumbersome and less performant. I would rephrase this to something like:

> The standard solution to this problem would be "run-time inheritence," where type deduction is performed at run-time. While this is possible on GPU, it becomes cumbersome, as the user must be very explicit about class inheritence. Moreover, run-time inheritence relies on relocatable device code, which is not as performant on device, thanks to weaker cross-compilation unit optimization. (We note that to fully leverage relocatable device code, the entire library must be made to be header-only.) 
>
>  We could have used a similar technique to the
modifier classes and pass the EOS as a template paramter, but then the vector
function calls could only be achieved by creating vector modifiers of all the
implemented EOS.

doc/sphinx/src/contributing.rst
```diff
@@ -2,3 +2,120 @@
...
+    }
+
+Note that the ``using`` statement needs to be included in order to properly
+overload the scalar functionality with the vector functionality. Otherwise the
+vector member function is hidden by the derived class method rather than
+overloaded.
+
+With several EOS that all inherit from the ``EosBase`` class, we can achieve
+static polymorphism in all of the EOS classes without having to implement
+vector member functions in each class.
```
Yurlungur, line 118: Perhaps also add a comment explaining the macros?

Yurlungur (commented):
Thanks, @jhp-lanl ! Overall, looks good. A few nitpicks below.

doc/sphinx/src/contributing.rst
dholladay00: "We note that to fully leverage relocatable device code (RDC), the entire library must be made header-only" -> "We note that to obtain full performance on device and to build with compilers that don't support (RDC), the entire library must be made header-only"

Yurlungur:
@jhp-lanl we've kind of let this MR slide. Do you want to pull it through? I can also take a stab at making the suggested revisions if you want.

jhp-lanl:
> @jhp-lanl we've kind of let this MR slide. Do you want to pull it through? I can also take a stab at making the suggested revisions if you want.

If you have the cycles, feel free! I'm focusing on the Gruneisen stuff right now, but I can get to this after I've tackled that.

Yurlungur:
Haha ok. I will see if I have a spare few minutes or not. I think you already got us 90% of the way there.

Yurlungur (approved):

Issue #72, opened 2022-01-24
Add pressure/density lookups


jhp-lanl:
We're currently missing pressure-density lookups in the code. These are useful for initializing a material to be in mechanical equilibrium with its surroundings (i.e. same pressure) but where the mass of the material is important so that the density needs to be specified.

I think for most EOS this wouldn't be too hard to implement and just requires some time and work.

Yurlungur:
These lookups are hidden inside `FillEos` for `SpinerEOSDependsRhoT`. But I agree it would be nice to have them generically.

jhp-lanl:
> These lookups are hidden inside `FillEos` for `SpinerEOSDependsRhoT`. But I agree it would be nice to have them generically.

I'm not sure the implementation is consistent though. For example, while the ideal gas EOS has
```c++
  if (output & thermalqs::temperature && output & thermalqs::specific_internal_energy) {
    sie = press / (_gm1 * rho);
  }
```
to get e(P, rho) which is then used to calculate everything else. For the Davis EOS however, it looks like the same lookup would result in using an uninitialized or old energy value :
```c++
  if (output & thermalqs::pressure) press = PressureFromDensityInternalEnergy(rho, sie);
  if (output & thermalqs::temperature)
    temp = TemperatureFromDensityInternalEnergy(rho, sie);
```
As far as I can see, spiner is okay but JWL and the Gruneisen EOS have the same issue. Please let me know if I'm reading this wrong though.

jdolence:
I wonder if the right answer for calls like this is for singularity to provide eos-agnostic (i.e. take an arbitrary EOS as an argument) utilities that just root find.  This would be very easy (I've already written one for a couple downstream codes that could be incorporated with a little bit of hardening) and I think would be fast enough for things like initialization.

jhp-lanl:
> I wonder if the right answer for calls like this is for singularity to provide eos-agnostic (i.e. take an arbitrary EOS as an argument) utilities that just root find. This would be very easy (I've already written one for a couple downstream codes that could be incorporated with a little bit of hardening) and I think would be fast enough for things like initialization.

If possible, I think it would be good to get the exact values for analytic EOS where possible so there's no convergence tolerance to worry about. But I think that would be an excellent idea for the tabular EOS. It should hopefully reduce code duplication.

jdolence:
[...]

PR #97, opened 2022-04-01
Update Gruneisen EOS

jhp-lanl:
## PR Summary

Numerous changes to the Gruneisen EOS

### Changes:

- Redoes how the Gruneisen EOS calculates the pressure and bulk modulus to allow specification of $P_0$. I also cast the equations in something more understandable I think.
- Updated documentation with to better mirror the code and provide a better description of $\Gamma$ vs $\gamma$ in this context
- Added a `rho_max` variable to the model that allows the user to specify a maximum compression for the EOS. Alternatively, if this is not specified, the model will calculate the density limit based off of the singularity that exists in the denominator of the pressure expression. Lookups beyond this density will just return the values at the limit, which should be sufficiently large to reasonably approximate the compressibility limit. Finding the limit for the linear Hugoniot case is trivial, but the cubic case required significantly more effort.
- Split up the `eos_unit_tests.cpp` file into multiple files for organization and to speed up compilation times when the Catch2 hooks have already been compiled.
- Add new Gruneisen EOS tests to ensure that the EOS both reproduced the jump conditions as well as produces a bulk modulus that is consistent with a finite difference approximation.

NOTE: The temperature and the EOS in expansion aren't thermodynamically consistent. This will need to be addressed in the future.

## PR Checklist


jhp-lanl:
@chadmeyer I'm sure you're sick of reviewing my EOS changes but I think you're still the best one to review here.

These changes were mainly necessary to make our implementation operable in Pagosa. Specifically, the P_0 parameter wasn't actually doing anything in our implementation but was used in Pagosa's. I've also decided to adopt the nomenclature of `eostype(m) = 11` in xRAGE since it's neater and I think demonstrates the Mie-Gruneisen form better.

There were diffs but they were in the last two decimals for pressure. The bulk modulus differences were larger, but were expected since the bulk modulus approximately scales with the square root of the pressure.

jhp-lanl:
> I just worry that without GPU test coverage, it'll be easy to make a change that causes a segfault or something down the line.

Fair enough. I may just loop it then rather than checking the full vector functionality.

Aiming for full code coverage is always my goal, so I should have just not been lazy :-)

Yurlungur:
Looping should be fine, since it inherits from the full vector stuff, which we already know works.

jhp-lanl:
@Yurlungur I made it run on GPUs and decided to just use the same strategy as the vector EOS test since it seemed the most similar to what I was already doing.

If there's an easier way to just copy one value at a time in a for loop, I'd be happy to change the code to do that. But at least it runs on the GPU now.

Yurlungur (approved):

Yurlungur:
There's a way to do that @jhp-lanl using either deep copies or `parReduce` but this is fine. I don't think it's any cleaner than what you've done.

Yurlungur:
I would like to wait until @chadmeyer reviews before merging if possible.

chadmeyer:
It would seem that there is no ability in github to comment on unchanged lines, so I'm just going to quote the section:
```c++
PORTABLE_FUNCTION void Gruneisen::DensityEnergyFromPressureTemperature(
    const Real press, const Real temp, Real *lambda, Real &rho, Real &sie) const {
  sie = _Cv * (temp - _T0);
  // We have a branch at rho0, so we need to decide, based on our pressure, whether we
  // should be above or below rho0
  Real Pref = PressureFromDensityInternalEnergy(_rho0, sie);
  if (press < Pref) {
    rho = (press + _C0 * _C0 * _rho0) / (_C0 * _C0 + _G0 * sie); <=== THIS LINE IS PROBABLY WRONG NOW!
  } else { // We are in compression; iterate
    auto residual = [&](const Real r) {
      return press - PressureFromDensityInternalEnergy(r, sie);
    };
    Real rho1 = _rho0, res1 = residual(rho1), slope = _G0 * sie + _C0 * _C0, rho2, res2,
         rhom, resm;
    rho2 = (rho > rho1 + 1e-3) ? rho : rho1 + res1 / slope;
    res2 = residual(rho2);
    for (int i = 0; i < 20; ++i) {
      slope = (rho2 - rho1) / (res2 - res1 + 1.0e-10);
      rhom = rho2 - res2 * slope;
      resm = residual(rhom);
      if (resm / press < 1e-8) break;
      rho1 = rho2;
      res1 = res2;
      rho2 = rhom;
      res2 = resm;
    }
    rho = rhom;
  }
}
```

I think the second half of the routine is OK because it iterates using functions that contain pressure and the `_P0` parameter, but I think this particular branch will be wrong.  @jhp-lanl, do you agree?

jhp-lanl:
> I think the second half of the routine is OK because it iterates using functions that contain pressure and the _P0 parameter, but I think this particular branch will be wrong. @jhp-lanl, do you agree?

@chadmeyer Good catch! Let me see what I can do here to reduce code duplication and make this consistent with my changes. I'll add a test as well.

Yurlungur:
I assume we're waiting on this until we resolve current questions downstream, @jhp-lanl @annapiegra ?

jhp-lanl:
> I assume we're waiting on this until we resolve current questions downstream, @jhp-lanl @annapiegra ?

We're waiting on a definitive formulation for the bulk modulus since it seems like the current formulations fail to reproduce the finite difference solution. We're also waiting on me since I've been busy with other things, but I'm going to be working on this tomorrow.

jhp-lanl:
The last thing to resolve on this MR is the introduction of a maximum density for the EOS. I'll do this next week

jhp-lanl:
@jonahm-LANL and @chadmeyer I've introduced the maximum density limit in the EOS now. It was made a bit complicated by the fact that we can have up to a cubic fit for the Hugoniot which could lead to multiple compression limits. Some of these limits may be excluded by other properties of the EOS, but instead of trying to run down all those limits, I figured it was easier to just naively assume that any fit could be given and that it could produce a compression limit for the EOS. If you guys can review the algorithm, I'm fairly certain (due to all the tests I've added) that the compression limit code is doing what it's supposed to.

@annapiegra and @mauneyc-LANL this will add the capability to introduce a maximum density limit if you want to merge it into your testing. I also have also added a few tests that make me fairly certain the pressure and bulk modulus are accurate in singularity: I have one test comparing the calculated bulk modulus to what you would get with finite differences, and I have another test comparing a state on the reference curve to what you would get from the Rankine-Hugoniot jump conditions. Both tests pass, which makes me fairly confident in the implementation we have here.

doc/sphinx/src/models.rst
```diff
@@ -214,40 +230,66 @@ The user should note that this implies that :math:`e=0` at the reference
...
       \begin{cases}
         \Gamma_0                                          & \rho < \rho_0 \\
-        \Gamma_0 \frac{\rho_0}{\rho} 
-           + b(1 - \frac{\rho_0}{\rho})                   & \rho >= \rho_0
+        \Gamma_0 (1 - \eta) + b\eta                       & \rho \geq \rho_0
       \end{cases}
 
-and when the unitless user parameter :math:`b=0`, this ensures the the Gruneisen
+When the unitless user parameter :math:`b=0`, this ensures the the Gruneisen
 parameter is of a form where :math:`\rho\Gamma =` constant in compression.
```
Yurlungur, line 251: Given that compression is now a variable, this phrasing has become confusing I think. Maybe "in the so-called compression limit, when when eta < 0"

singularity-eos/eos/eos.hpp
```diff
@@ -125,10 +125,19 @@ class Gruneisen : public EosBase<Gruneisen> {
...
+            const Real rho_max)
+      : _C0(C0), _s1(s1), _s2(s2), _s3(s3), _G0(G0), _b(b), _rho0(rho0), _T0(T0), _P0(P0),
+        _Cv(Cv), _rho_max(rho_max) {}
+  // Constructor when rho_max isn't specified automatically determines _rho_max
+  PORTABLE_INLINE_FUNCTION
   Gruneisen(const Real C0, const Real s1, const Real s2, const Real s3, const Real G0,
             const Real b, const Real rho0, const Real T0, const Real P0, const Real Cv)
       : _C0(C0), _s1(s1), _s2(s2), _s3(s3), _G0(G0), _b(b), _rho0(rho0), _T0(T0), _P0(P0),
-        _Cv(Cv) {}
+        _Cv(Cv), _rho_max(0.99 * GetRhoMax(s1, s2, s3, rho0)) {}
```
Yurlungur, line 138: I assume the 0.99 is a safety factor?

test/eos_unit_test_helpers.hpp
```diff
@@ -0,0 +1,101 @@
...
+// granted for itself and others acting on its behalf a nonexclusive,
+// paid-up, irrevocable worldwide license in this material to reproduce,
+// prepare derivative works, distribute copies to the public, perform
+// publicly and display publicly, and to permit others to do so.
+//------------------------------------------------------------------------------
+
+#ifndef _SINGULARITY_EOS_TEST_TEST_HELPERS_
+#define _SINGULARITY_EOS_TEST_TEST_HELPERS_
+
+#include "catch2/catch.hpp"
```
Yurlungur, line 18: Missing `#ifndef CATCH_CONFIG_RUNNER` guard

singularity-eos/eos/eos_gruneisen.cpp
```diff
@@ -12,12 +12,155 @@
...
+  // Try to find the minimum bounded value. If none exists, return a value less than the
+  // minimum
+  const Real minroot = std::min(root1, root2);
+  const Real maxroot = std::max(root1, root2);
+  if (minroot > min && minroot < max) {
+    return minroot;
+  } else if (maxroot > min && maxroot < max) {
+    return maxroot;
+  } else {
+    return min - 1.;
```
Yurlungur, line 35: how should I interpret this else case? Is this a fail case? If so perhaps we should return a specific flag number, like signal_nan?

singularity-eos/eos/eos_gruneisen.cpp
```diff
@@ -12,12 +12,155 @@
...
+          // will lie outside of the extrema. Further, the only possibility for multiple
+          // bound roots will occur when the extremum are both positive
+          if (min_extremum > 0) {
+            maxbound = std::min(min_extremum, maxbound);
+          }
+        }
+      }
+      if (poly(minbound) * poly(maxbound) <= 0) {
+        // Root is appropriately bounded
+        using RootFinding1D::bisect;
```
Yurlungur, line 112: I might try the new regula falsi method instead of bisection, thought it probably doesn't matter too much

singularity-eos/eos/eos_gruneisen.cpp
```diff
@@ -12,12 +12,155 @@
...
+        // Single root with multiplicity of 3 (inflection point is root)
+        // Also guards against divide by zero
+        root = -s2 / 3 / s3;
+      } else {
+        // One root has multiplicity of 2 (local extremum is a root)
+        root1 = (-9 * s3 - s2 * s1) / 2 / (square(s2) - 3 * s3 * s1);
+        root2 = (-4 * s3 * s2 * s1 - 9 * square(s3) + cube(s2)) /
+                (-s3 * (square(s2) - 3 * s3 * s1));
+        root = find_min_bounded_val(root1, root2, 0, 1);
+      }
```
Yurlungur, line 80: Given that this only gets called in initialization, is there a reason to do the analytic root finding at all? Maybe we could save some code and always use the numerical root finder?

If we did the numerical root finding always, we could use the analytic formulae as checks that everything is working properly in the unit tests.

singularity-eos/eos/eos_gruneisen.cpp
```diff
@@ -12,12 +12,155 @@
...
+    return maxroot;
+  } else {
+    return min - 1.;
+  }
+}
+/*
+The Gruneisen EOS diverges at a specific compression. Ensure that the maximum density is
+below the smallest singularity in the reference pressure curve
+*/
+PORTABLE_FUNCTION Real Gruneisen::GetRhoMax(const Real s1, const Real s2, const Real s3,
```
Yurlungur, line 42: It occurs to me that querying what rhomax is from the eos might be desirable but we don't want to re-do thhis calculation. I suggest renaming this `ComputeRhoMax` but exposing a method that return the value stored in the class for debugging purposes.

singularity-eos/eos/eos_gruneisen.cpp
```diff
@@ -12,12 +12,155 @@
...
+      if (discriminant > 0) {
+        // Three real roots. We need to use the extrema to ensure we have a proper bracket
+        // in which to search for the root.
+        const Real extremum1 =
+            (2 * s2 + std::sqrt(square(-2 * s2) - 4 * 3 * s3 * s1)) / (-2 * 3 * s3);
+        const Real extremum2 =
+            (2 * s2 - std::sqrt(square(-2 * s2) - 4 * 3 * s3 * s1)) / (-2 * 3 * s3);
+        const Real min_extremum = std::min(extremum1, extremum2);
+        const Real max_extremum = std::max(extremum1, extremum2);
+        if (s3 < 0) {
```
Yurlungur, line 94: I like the extrema trick to find the root you want. Nice idea.

singularity-eos/eos/eos_gruneisen.cpp
```diff
@@ -12,12 +12,155 @@
...
+        // Root doesn't lie within physical bounds for eta so no maximum density exists
+        root = -1.;
+      }
+    } // Cubic discriminant >= 0
+  }   // Linear/Quadratic/Cubic
+  if (root > 0) {
+    return rho0 / (1 - root);
+  } else {
+    // No bounded root exists so there is no upper-limit on the compression
+    return 1.e99;
```
Yurlungur, line 140: should you return infinity here? Or will that cause problems?

test/test_eos_gruneisen.cpp
```diff
@@ -0,0 +1,546 @@
...
+  /* These tests all test the functionality that finds the roots of the polynomial that
+     makes up the denominator of the reference pressure curve, i.e.
+     P(x) = 1 - s1 * x - s2 * x**2 - s3 * x**3,
+     in order to find the maximum compression allowed by the EOS.
+
+     It's very probable that many of these cases may not be physical because they produce
+     unstable shocks (i.e. Us - up < c) but we admit these cases for the moment since
+     a more comprehensive set of bounds on the EOS parameters is not currently available.
+     For a linear Us-up relationship, the point at which up > Us occurs also corresponds
+     to the singularity in the
```
Yurlungur, line 282: incomplete sentence

test/test_eos_gruneisen.cpp
```diff
@@ -0,0 +1,546 @@
...
+        THEN("The returned Gamma(rho, e) should be equal to the true value") {
+          array_compare(num, density, energy, h_gruneisen, gruneisen_true, "Density",
+                        "Energy");
+        }
+      }
+    }
+  }
+}
+
+SCENARIO("Aluminum Gruneisen EOS Sound Speed and Pressure Comparison", "[GruneisenEOS]") {
```
Yurlungur, line 176: Would it be worth checking that the answer is reasonable way in compression? Maybe you're doing that?

Yurlungur (approved):
Overall this looks good and these changes seem reasonable, and well tested to me. A few comments below. Also @chadmeyer should double check the physics reasoning.

doc/sphinx/src/models.rst
```diff
@@ -214,40 +230,66 @@ The user should note that this implies that :math:`e=0` at the reference
 temperature, :math:`T_0`. Given this simple relationship, the user should
 treat the temperature from this EOS as only a rough estimate.
 
-The Grunesien parameter is given by
+First defining the compression as
```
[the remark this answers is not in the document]
jhp-lanl: I think it's grammatically correct, but I'm using the passive voice pretty heavily. I'll figure out a way to make it more active

singularity-eos/eos/eos_gruneisen.cpp
jhp-lanl: I might have coded this to be a little too general here, because my intent was to just fail at an arbitrary number outside of the bounds on the minimum side. Since I'm only operating on `eta` in compression, any negative value would suffice. I'll change this.

singularity-eos/eos/eos_gruneisen.cpp
jhp-lanl: The reason I'm doing analytic root finding is because I need to guarantee that the  minimum root is found. I briefly looked into the numerics of this and decided that there was no guaranteed way to find the _minimum_ bounded root without more information.

Feel free to find a better way though.

singularity-eos/eos/eos_gruneisen.cpp
jhp-lanl: Yeah it wasn't available yet when I was coding this up. I just wanted to avoid secant really. I'll change this to the false position root finder

singularity-eos/eos/eos_gruneisen.cpp
jhp-lanl: I just realized I never implemented the cutoff in the actually lookups 🤦 .

My intent was just to min the input density against the maximum. It looks like `std::numerical_limits::infinity` needs a special `max()` function. Is there a way to avoid this?

doc/sphinx/src/models.rst
Yurlungur: I think including rho Gamma = constant is good, I just want to reference what "in compression" means here.

singularity-eos/eos/eos.hpp
Yurlungur: Might be good to define a static class variable `static constexpr Real RHOMAX_SAFETY = 0.99` or something... and maybe a comment indicating this is consistent with Pagosa.

singularity-eos/eos/eos_gruneisen.cpp
Yurlungur: I think `std::min(rho, std::numerical_limits::infinity());` does what you want? Am I missing something?

doc/sphinx/src/models.rst
jhp-lanl: Ah gotcha. That makes sense! I'll change that

singularity-eos/eos/eos_gruneisen.cpp
jhp-lanl: > 👍 Maybe add a comment explaining the logic as well.

There is a comment at the top of the function explaining the behavior. Were you wanting something more verbose?

singularity-eos/eos/eos_gruneisen.cpp
jhp-lanl: Yeah I don't know what I was thinking... I'll implement it

test/test_eos_gruneisen.cpp
jhp-lanl: Not yet! I should put in a simple check that the values beyond max compression simply return the value at max compression.

singularity-eos/eos/eos_gruneisen.cpp
jhp-lanl: Are you talking about the cubic case specifically? I think the extrema equations are still valid when the discriminant is also zero so I could probably relax the condition to `if (discriminant >= 0) {` and eliminate that code. The good thing about writing that code is that I do at least test it so the tests should automatically tell me whether it's feasible to eliminate the code.

For the quadratic and linear cases I don't think you're saving much. For the quadratic case you'd still need to find the extremum anyway and then have a conditional for the sign of `s2` to determine whether you're greater than or less than the extremum.

Yurlungur: Ah yeah I guess the extremum trick only works for cubic.

singularity-eos/eos/eos_gruneisen.cpp
Yurlungur: nah just a comment at this else case like "any negative number will do" or something like that.

singularity-eos/eos/eos_gruneisen.cpp
jhp-lanl: I added some comments to the extremum trick to make it more apparent that the trick works since the roots must be in a region of the polynomial that is _decreasing_. This is because the polynomial is pinned to 1 at $\eta = 0$, so there must be a decreasing region between $\eta = 0$ and $\eta = 1$ for there to be a possibility of any roots.

I also realized another reason to delete the analytic root part: the roots will only have multiplicity (i.e. the determinant is zero) when an extremum _**is**_ a root anyway. So I'm now requiring an explicit sign change across the bounds and otherwise testing the bounds themselves since the iterative approaches have a hard time converging if the root is an extremum.

Yurlungur: That sounds good! We should double check it all works in the tests, of course. We can use the code you just replaced as a sanity check. :)

test/test_eos_gruneisen.cpp
jhp-lanl: Test added

singularity-eos/eos/eos_gruneisen.cpp
```diff
@@ -127,11 +270,24 @@ PORTABLE_FUNCTION void Gruneisen::DensityEnergyFromPressureTemperature(
...
+  }
+  else if (thermalqs::density & output || thermalqs::specific_internal_energy & output) {
+    // Error out on density or energy output because they're currently required as inputs
+    std::stringstream errorMessage;
+    errorMessage << "Gruneisen FillEos: Density and energy are currently required inputs "
+                 << "except when pressure and temperature are inputs"
+                 << std::endl;
+    EOS_ERROR(errorMessage.str().c_str());
+  }
+  const Real rho = std::min(rho_in, _rho_max);
```
jhp-lanl, line 290: @jonahm-LANL note that I needed to modify the `FillEos` function a bit to add the density check. You may want to look at this as well.

This is related to issue https://github.com/lanl/singularity-eos/issues/72

singularity-eos/eos/eos_gruneisen.cpp
jhp-lanl: I believe I already have tests that capture the behavior of the previous code so in that sense I think it works.

Interestingly the tests for root multiplicity (i.e. roots are extrema) failed when using the iterative root find because the tolerances were too tight. The main issue is that the root find tolerance is too loose for this particular situation where the dy is basically zero. I would have assumed that regula falsi would converge directly to a bound if that bound is a root though, so I'm not quire sure what's going on.

Either way, skipping the root find when one of the bounds is a solution is probably the intelligent thing to do here anyway.

Yurlungur: Yeah that seems fine. My experience with Regula Falsi has been it's a little sensitive compared to bisection or secant actually. It's very fast, but it seems to fail out more frequently.

singularity-eos/eos/eos_gruneisen.cpp
```diff
@@ -12,90 +12,233 @@
...
+  } else if (s3 == 0 && s2 != 0) {
+    // Quadratic Us-up
+    discriminant = square(s1) + 4 * s2;
+    if (discriminant < 0) {
+      root = -1; // Imaginary roots, so no limit
+    } else {
+      root1 = (s1 + std::sqrt(discriminant)) / (-2 * s2);
+      root2 = (s1 - std::sqrt(discriminant)) / (-2 * s2);
+      root = find_min_bounded_val(root1, root2, 0, 1);
+    }
```
mauneyc-LANL, line 67: I'm always a bit nervous about math w/ implicit integer conversions
```c++
root = 1 / s1;
```
and testing FP equality w/ a literal integer `0`.

```c++
if (s2 == 0 && s3 == 0 && s1 > 0) {
```

I can also see a lot of issues with division by strictly "non-zero" FP numbers that are only a few bits different than `0`.

Should there be a tolerance on these conditions? e.g. `if (std::abs(s2)<TOL /* ... */ )`.
(I may be missing something key here, tho)

Yurlungur: THis is a good catch. I second @mauneyc-LANL 's suggestion. Might be good to divide by denom + a little bit too.

jhp-lanl: > I'm always a bit nervous about math w/ implicit integer conversions
> ```c++
> root = 1 / s1;
> ```

Is switching to
```c++
root = 1. / s1;
```
enough to alleviate this issue or should we define some of
```c++
constexpr Real ONE = 1.;
constexpr Real TWO = 1.;
constexpr Real HALF = 1.;
...
```
in `constants.hpp`?

> Should there be a tolerance on these conditions? e.g. if (std::abs(s2)<TOL /* ... */ ).
Sure that's not a bad idea. If that denominator is close to zero, I suppose it's probably good enough to use the linear case anyway and I'll make those comparisons to zero a bit more loose.

Yurlungur: I think `1. / s1` should resolve the implicit conversion question. Let's not do `constexpr Real ONE = 1`.

I would be more nervous about divide by zero issues and exact comparison of real numbers with `==`. It's easy for floating point math to never exactly equal zero, so the tolerance trick is much more likely to be what we actually want.

Yurlungur: So for `root = 1. / s1` I would actually do:
```C++
root = 1. / (s1 + MY_SIGN(s1)*EPS);
```
where `EPS` is some small number, like machine epsilon, and `MY_SIGN` evaluates to 1 if s1 >= 0 and -1 otherwise.

jhp-lanl: Yeah I went with explicit comparisons because it's not a derived quantity but rather user-supplied parameters. I'd hope a user wouldn't try to provide a quadratic fit parameter when it's 1e-109, but who knows 🙃 . I'll make the changes shortly

Yurlungur: I think you probably already know all this. But this discussion of floating points reminded me of some nice online resources on working with floating point numbers :)

- [What every computer scientist should know about floating-point arithmetic](https://docs.oracle.com/cd/E19957-01/806-3568/ncg_goldberg.html)
- [Five tips for floating point programming](https://www.codeproject.com/Articles/29637/Five-Tips-for-Floating-Point-Programming)
- [Avoiding overflow and underflow](https://www.codeproject.com/Articles/25294/Avoiding-Overflow-Underflow-and-Loss-of-Precision)

singularity-eos/eos/eos_gruneisen.cpp
```diff
@@ -12,90 +12,233 @@
...
+    Real minbound = 0;
+    Real maxbound = 1;
+    if (discriminant >= 0) {
+      // Three real roots (roots may have multiplicity). We need to use the extrema to
+      // ensure we have a proper bracket in which to search for the root (if they exist).
+      // Note: If the discriminant is positive, then `square(s2) - 3 * s3 * s1` must
+      //       necessarily be positive (the reverse does not hold). Thus the extrema below
+      //       are not imaginary.
+      const Real extremum1 =
+          (2 * s2 + std::sqrt(square(-2 * s2) - 4 * 3 * s3 * s1)) / (-2 * 3 * s3);
```
mauneyc-LANL, line 83: This is mostly for discussion, and am not really advocating for any change.

There could be some FLOPs saved here with a bit of simplification (may be wise to double-check me here)

```c++
ex1 = (s2 + std::sqrt(square(s2) - 3. * s3 * s1)/(-3. * s3);
```

and, if you wanna go REALLY crazy with it

```c++
auto threeS3 = 3. * s3;
auto I_threeS3 = 1. / threeS3; // division is expensive, so if there's reuse my instinct is to move to multipication
auto t1 = std::sqrt(square(s2) - threeS3 * s1);
Real ex1 = (s2 + t1) * I_threeS3;
Real ex2 = (s2 - t1) * I_threeS3;
```
This maybe a bit much, and likely as not most compilers are good enough in optimization here. There's also an argument to be made that in some contexts it's better to "load" FLOPS. On the other hand, keeping the FP arithmetic less "expressive" can help contain round-off and cancellation errors.

(side note: the quad formula is notorious for reaking havoc on precision, see Theorem 2 of https://docs.oracle.com/cd/E19957-01/806-3568/ncg_goldberg.html, https://www.johndcook.com/blog/2018/04/28/quadratic-formula/ for a nice summary).

Yurlungur: good catch

singularity-eos/eos/eos_gruneisen.cpp
```diff
@@ -49,45 +52,48 @@ PORTABLE_FUNCTION Real Gruneisen::ComputeRhoMax(const Real s1, const Real s2, co
...
 
+  // Comparisons to zero are actually made to
+  static constexpr Real EPS = std::numeric_limits<Real>::epsilon();
+
   // First find the eta root. A negative root indicates that there is no maximum density.
-  Real root = 0; // Non-sensical root means a root hasn't been found yet
+  Real root = -1.; // Non-sensical root means none exists
   Real discriminant, root1, root2;
-  if (s2 == 0 && s3 == 0 && s1 > 0) {
+  if (is_near_zero(s2, EPS) && is_near_zero(s3, EPS) && s1 > EPS) {
```
Yurlungur, line 61: are the s's all guaranteed to be positive? i.e., should this be `std::abs(s1) > EPS`?

singularity-eos/eos/eos_gruneisen.cpp
mauneyc-LANL: > I think `1. / s1` should resolve the implicit conversion question. Let's not do `constexpr Real ONE = 1`.
> 
> I would be more nervous about divide by zero issues and exact comparison of real numbers with `==`. It's easy for floating point math to never exactly equal zero, so the tolerance trick is much more likely to be what we actually want.

Offering an opinion: I'm generally in favor of named constants over literal constants, as it avoids issues with having to sort out when/how implict conversions happen. I'm sympathetic to the idea that `ONE` ect., is a bit much.

It may be worthwhile to have and use some number constants (a bit beyond the scope of this PR)

```c++
inline constexpr Real e          = /*...*/;
inline constexpr Real pi         = /*...*/;
inline constexpr Real ln2        = /*...*/;
inline constexpr Real sqrt2      = /*...*/;
inline constexpr Real inv_sqrt2  =/*...*/;
// and so on...
```

singularity-eos/eos/eos_gruneisen.cpp
mauneyc-LANL: > (side note: the quad formula is notorious for reaking havoc on precision, see Theorem 2 of https://docs.oracle.com/cd/E19957-01/806-3568/ncg_goldberg.html, https://www.johndcook.com/blog/2018/04/28/quadratic-formula/ for a nice summary).

I see @Yurlungur got the jump on me with this

jhp-lanl: This code should be just executed at initialization so I'm not too concerned about performance here, BUT I think your first suggestion enhances readability since it's a pretty small simplification that doesn't make it any less obvious what is being calculated here.

I think the second suggestion makes the code slightly less readable though. The fact that you were able to instantly recognize the quadratic formula I think is a good thing here.

singularity-eos/eos/eos_gruneisen.cpp
jhp-lanl: I would first argue that there is no guarantee since we don't enforce anything about the selection of the s values so I wouldn't assume anything about sign.

But in this case the critical test is that `s1` is positive, not that it is nonzero. If `s1` is negative, then the root would be negative anyway and it's not relevant.

However, you do remind me that if `s1 < 1`, the root would be greater than 1 anyway which is nonsensical. I'm not guarding against that though so really the condition should be `s1 > 1.`. I'll fix that

singularity-eos/eos/eos_gruneisen.cpp
mauneyc-LANL: > Is switching to
> 
> ```c++
> root = 1. / s1;
> ```
> 
> enough to alleviate this issue or should we define some of
> 
> ```c++
> constexpr Real ONE = 1.;
> constexpr Real TWO = 2.;
> constexpr Real HALF = ONE / TWO;
> ...
> ```
> 
> in `constants.hpp`?

using `1.` is usually what I default to doing

singularity-eos/eos/eos_gruneisen.cpp
Yurlungur: sounds good.

singularity-eos/eos/eos_gruneisen.cpp
Yurlungur: > > I think `1. / s1` should resolve the implicit conversion question. Let's not do `constexpr Real ONE = 1`.
> > I would be more nervous about divide by zero issues and exact comparison of real numbers with `==`. It's easy for floating point math to never exactly equal zero, so the tolerance trick is much more likely to be what we actually want.
> 
> Offering an opinion: I'm generally in favor of named constants over literal constants, as it avoids issues with having to sort out when/how implict conversions happen. I'm sympathetic to the idea that `ONE` ect., is a bit much.
> 
> It may be worthwhile to have and use some number constants (a bit beyond the scope of this PR)
> 
> ```c++
> inline constexpr Real e          = /*...*/;
> inline constexpr Real pi         = /*...*/;
> inline constexpr Real ln2        = /*...*/;
> inline constexpr Real sqrt2      = /*...*/;
> inline constexpr Real inv_sqrt2  =/*...*/;
> // and so on...
> ```

We actually have *some* of this, thanks to @brryan in the `singularity-eos/base/constants.hpp` file. It could be extended to include some particularly common non-rational numbers that aren't physical I would think.

Yurlungur: Obviously that's a story for another time of course

singularity-eos/eos/eos_gruneisen.cpp
jhp-lanl: FYI - I think I can get away with losing precision in the discriminants of the quadratic formulas here for two reasons:
1.  The value of the extrema calculations don't need to be really precise since they are used to bracket the real root more appropriately.
2.  When I'm directly calculating the roots for the quadratic case, I can reasonably rely on `s1` to be of order unity. That way although I may lose precision in the discriminant if it's near zero, I don't really care since I'll basically get two very close estimates of the root. If the lower estimate is too conservative or too affected by machine precision issues, then we can re-examine this issue later.

jhp-lanl:
I think I've addressed everybody's comments at this point and all tests pass both here and on gitlab. Let me know by tomorrow if you have other concerns.

singularity-eos/eos/eos.hpp
```diff
@@ -125,10 +125,19 @@ class Gruneisen : public EosBase<Gruneisen> {
  public:
   Gruneisen() = default;
   PORTABLE_INLINE_FUNCTION
+  Gruneisen(const Real C0, const Real s1, const Real s2, const Real s3, const Real G0,
+            const Real b, const Real rho0, const Real T0, const Real P0, const Real Cv,
+            const Real rho_max)
+      : _C0(C0), _s1(s1), _s2(s2), _s3(s3), _G0(G0), _b(b), _rho0(rho0), _T0(T0), _P0(P0),
+        _Cv(Cv), _rho_max(rho_max) {}
```
chadmeyer, line 132: Should there be a check here in the constructor to ensure that user-specified rho_max is sane (and, for instance, less than the result of `ComputeRhoMax`)?

Yurlungur: I would prefer to warn rather than die if rho_max is specified to be greather than `computeRhoMax`  but a warning would be good maybe.

jhp-lanl: I'll create a hotfix for this shortly.

Just to be sure, @Yurlungur , we don't have a full-fledged warning mechanism other than print statements, right?

Yurlungur: We don't. Here's my suggested design for now:
```C++
#ifndef NDEBUG
if (rho max condition violated) {
printf("some warning message);
}
#endif
```

I'll try to submit some more formal warning machinery as a PR at some point

jhp-lanl: Updated here:
https://github.com/lanl/singularity-eos/pull/130

PR #101, opened 2022-04-15
Dholladay00/sap ramp

dholladay00:
## PR Summary
This PR will introduce a new modifier that enables ramping an EOS in a way that is compatible with ramp implementations in downstream host codes. 

This PR is still in a very draft state, WIP PR put in place for comments since there are a few possible paths of pursuit. This can provide a location for discussion.

## PR Checklist


singularity-eos/eos/modifiers/ramps_eos.hpp
```diff
@@ -0,0 +1,163 @@
...
+    Real senergy;
+    switch (t_.PreferredInput()) {
+    case thermalqs::density | thermalqs::temperature:
+      t_.FillEos(rho, temp, energy, press, cv, bmod, output, lambda);
+      energy = energy + shift_;
+      break;
+    case thermalqs::density | thermalqs::specific_internal_energy:
+      senergy = energy - shift_;
+      t_.FillEos(rho, temp, senergy, press, cv, bmod, output, lambda);
+      break;
```
Yurlungur, line 122: should we cover more cases?

singularity-eos/eos/modifiers/ramps_eos.hpp
```diff
@@ -0,0 +1,163 @@
...
+  PORTABLE_FUNCTION
+  void ValuesAtReferenceState(Real &rho, Real &temp, Real &sie, Real &press, Real &cv,
+                              Real &bmod, Real &dpde, Real &dvdt,
+                              Real *lambda = nullptr) const {
+    t_.ValuesAtReferenceState(rho, temp, sie, press, cv, bmod, dpde, dvdt, lambda);
+    sie += shift_;
+  }
+
+  // Vector functions that overload the scalar versions declared here.
+  SG_ADD_BASE_CLASS_USINGS(SAPRampEOS<T>)
```
Yurlungur, line 154: specific overloads might be necessary

Yurlungur (commented):
Overall this seems reasonable and basically dirt simple, which I'm a fan of.

jhp-lanl:
I'll give this a thorough review tomorrow

singularity-eos/eos/modifiers/ramps_eos.hpp
dholladay00: this is copy/paste code from shifted eos. I will work finishing updating the code when I get a chance.

singularity-eos/eos/singularity_eos.cpp
```diff
@@ -35,15 +35,30 @@ int init_sg_eos(const int nmat, EOS *&eos) {
...
+					   vals[4], vals[5])
+
+constexpr const int def_en[4] = {0, 0, 0, 0};
+constexpr const double def_v[6] = {0.0, 0.0, 0.0, 0.0, 0.0, 0.0};
 
 int init_sg_IdealGas(const int matindex, EOS *eos, const double gm1, const double Cv,
-                     int const *const enabled, double const *const vals) {
+                     int const *const enabled, double const *const vals_i) {
   assert(matindex >= 0);
+  Real vals[6] = {vals_i[0], vals_i[1], vals_i[2], vals_i[3], vals_i[4], vals_i[5]};
```
dholladay00, line 56: This is how I'm currently handling the multiple ways to initialize a ramp that covers standard initialization and initialization from p-alpha model parameters. If this looks ok, I'll do this for all of the other Eos's. Would like some eyes maybe @Yurlungur and @jhp-lanl

Yurlungur: This seems fine to me. The names are a little vague, though. What is `vals` vs `vals_i`?

Yurlungur: I think letting them be non-const is the cleaner and clearer design

singularity-eos/eos/modifiers/ramps_eos.hpp
```diff
@@ -0,0 +1,213 @@
...
+  }
+  PORTABLE_FUNCTION
+  Real SpecificHeatFromDensityTemperature(const Real rho, const Real temperature,
+                                          Real *lambda = nullptr) const {
+    return t_.SpecificHeatFromDensityTemperature(rho, temperature, lambda);
+  }
+  PORTABLE_FUNCTION
+  Real BulkModulusFromDensityTemperature(const Real rho, const Real temperature,
+                                         Real *lambda = nullptr) const {
+    return t_.BulkModulusFromDensityTemperature(rho, temperature, lambda);
```
jhp-lanl, line 148: I _believe_ most implementations of a ramp utilize the slope of the ramp itself in the bulk modulus calculation

singularity-eos/eos/modifiers/ramps_eos.hpp
jhp-lanl: Won't the vector calls just access the scalar versions where the modifier is already applied?

singularity-eos/eos/modifiers/ramps_eos.hpp
```diff
@@ -0,0 +1,213 @@
...
+  PORTABLE_FUNCTION
+  Real GruneisenParamFromDensityTemperature(const Real rho, const Real temperature,
+                                            Real *lambda = nullptr) const {
+    return t_.GruneisenParamFromDensityTemperature(rho, temperature, lambda);
+  }
+  PORTABLE_FUNCTION
+  void FillEos(Real &rho, Real &temp, Real &energy, Real &press, Real &cv, Real &bmod,
+               const unsigned long output, Real *lambda = nullptr) const {
+    // density must be an input
+    assert(!(output & thermalqs::density));
```
jhp-lanl, line 159: We'll need P-T initialization for this to be used in EAP. Since the ramp is bilinear, I'd think it would be pretty easy to write a function to invert it so that you can get rho(P) instead of P(rho).

The tricky part will be figuring out if you're on the ramp or not without knowledge of `r1`. Maybe you can call `FillEos` assuming no ramp and then compare the output density of that to the value you get out of the ramp and see which one is greater. If the EOS density is greater than the ramp then you're on the ramp.

So maybe something like
```c++
if (output & thermalqs:density) {
  assert(!(output & thermalqs::pressure));  // Need pressure if density isn't specified
  Real temp_in = temp; 
  Real energy_in = energy;
  Real press_in = press;
  t_.Filleos(rho, temp_in, energy_in, press_in, cv, bmod, lambda);
  const Real rho_ramp = get_ramp_density(press);
  if (rho > rho_ramp) {
    rho = rho_ramp;
  }
}
```
There may be more I'm missing though like how to modify the bulk modulus...

singularity-eos/eos/modifiers/ramps_eos.hpp
```diff
@@ -0,0 +1,213 @@
...
+  PORTABLE_FUNCTION
+  Real GruneisenParamFromDensityInternalEnergy(const Real rho, const Real sie,
+                                               Real *lambda = nullptr) const {
+    return t_.GruneisenParamFromDensityInternalEnergy(rho, sie, lambda);
+  }
+  PORTABLE_FUNCTION
+  Real PressureFromDensityTemperature(const Real rho, const Real temperature,
+                                      Real *lambda = nullptr) const {
+    const Real p_ramp{get_ramp_pressure(rho)};
+    const Real p_eos{t_.PressureFromDensityTemperature(rho, temperature, lambda)};
```
jhp-lanl, line 137: for my own edification, is there a particular reason to use curly bracket initialization here rather than just the assignment operator?

jhp-lanl (changes requested):

singularity-eos/eos/modifiers/ramps_eos.hpp
Yurlungur: > We'll need P-T initialization for this to be used in EAP. Since the ramp is bilinear, I'd think it would be pretty easy to write a function to invert it so that you can get rho(P) instead of P(rho).

@jdolence and I actually got a request from Ristra recently for a general API for this for all singularity-eos EOS's. I think it's a great opportunity for the base class. We can have a default functionality that uses a numerical root find and overload as needed.

singularity-eos/eos/modifiers/ramps_eos.hpp
dholladay00: Yes, like I said (even though I'm inconsistent with it), it is a good habit to have. Though for POD types it's not needed.

Yurlungur: I don't think in this context it matters much... Despite being around @dholladay00 for a long time now, I still haven't picked up this habit for scalars. :P

singularity-eos/eos/modifiers/ramps_eos.hpp
jhp-lanl: > > We'll need P-T initialization for this to be used in EAP. Since the ramp is bilinear, I'd think it would be pretty easy to write a function to invert it so that you can get rho(P) instead of P(rho).
> 
> @jdolence and I actually got a request from Ristra recently for a general API for this for all singularity-eos EOS's. I think it's a great opportunity for the base class. We can have a default functionality that uses a numerical root find and overload as needed.

You mean putting a general version of `FillEos` in the base class? At the very least we definitely need to standardize its behavior across all of the classes, but yes I agree that a general implementation in the base class would be a good idea. I'll create an issue for this.

But regardless, I would still think we need a special `FillEos` here since the ramp complicates things.

Yurlungur: > You mean putting a general version of FillEos in the base class? 

I meant providing a `rho(P, T)` and/or `rho(P, e)` function in the base class. That way it's available for `FillEos` to use for individual EOS's. But yeah I agree we need to standardize behaviour

jhp-lanl: > > You mean putting a general version of FillEos in the base class?
> 
> I meant providing a `rho(P, T)` and/or `rho(P, e)` function in the base class. That way it's available for `FillEos` to use for individual EOS's. But yeah I agree we need to standardize behaviour

Ah okay, yes those are probably good ideas

singularity-eos/eos/modifiers/ramps_eos.hpp
dholladay00: how do I tell when to use the ramp slope vs the eos evaluated one?

jhp-lanl: You probably need to do the ramp pressure vs EOS pressure test to figure out if you're on the ramp or not

jhp-lanl: To add to this, most ramps assume that the ramp defines a roughly isentropic path for the material which makes the slope of the ramp a direct input (i.e. no thermodynamic relations needed) to the isentropic bulk modulus calculation.

Whether this is physical or not is another question, but if we want to match the current implementations I would just use the slope of the ramp.

singularity-eos/eos/modifiers/ramps_eos.hpp
dholladay00: @jhp-lanl could you take a look at the most recent changes and see if they address your concerns?

dholladay00: @jhp-lanl could you take a look at the most recent changes and see if they address your concerns?

singularity-eos/eos/singularity_eos.cpp
dholladay00: @Yurlungur this has now been addressed.

singularity-eos/eos/modifiers/ramps_eos.hpp
```diff
@@ -171,6 +186,16 @@ class SAPRampEOS : public EosBase<SAPRampEOS<T>> {
     }
     // call internal filleos
     t_.FillEos(rho, temp, energy, press, cv, bmod, ramp_out, lambda);
+    // fill ramp density
+    Real rho_ramp{rho};
+    if (output & thermalqs::density) {
+      assert(!(output & thermalqs::pressure));
+      rho_ramp = get_ramp_density(press);
+    }
+    rho = rho_ramp > rho ? rho_ramp : rho;
```
jhp-lanl, line 195: Starting a new code comment to keep things clean.

I think the logic here is backwards. The ramp is used to decrease the EOS density so you only want to use it when it's less than the EOS density.

jhp-lanl (changes requested):
Just one issue in the comparison of the ramp and EOS densities. After that's fixed I approve.

utils/scripts/format.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 #------------------------------------------------------------------------------
-# © 2021-2022. Triad National Security, LLC. All rights reserved.  This
+# © 2022. Triad National Security, LLC. All rights reserved.  This
```
jhp-lanl, line 4: Are we changing the copyright again?

jhp-lanl (changes requested):
Sorry one more thing I noticed: there are no tests for the ramp itself and only that the modifier doesn't do anything with default settings (a good test!).

I would propose three tests:
1. Verify B_S and P on the first leg of the ramp
2. Verify B_S and P on the second leg of the ramp
3. Verify that B_S and P are the same as the EOS for a density beyond the ramp

utils/scripts/format.sh
Yurlungur: oh oops. No this was on me. I copy-pasted this script out of Phoebus.

singularity-eos/eos/modifiers/ramps_eos.hpp
```diff
@@ -0,0 +1,239 @@
...
+
+using namespace eos_base;
+
+template <typename T>
+void pAlpha2SAPRampParams(const T &eos, const Real alpha0, const Real Pe, const Real Pc,
+                          Real &r0, Real &a, Real &b, Real &c) {
+  // get reference conditions
+  Real rho0, T0, sie0, P0, cv0, bmod0, dpde0, dvdt0;
+  Real rmid, r1;
+  eos.ValuesAtReferenceState(rho0, T0, sie0, P0, cv0, bmod0, dpde0, dvdt0);
```
Yurlungur, line 41: This is a conversation for a later time---but with the new PTE solver, do we *need* `ValuesAtReferenceState`? If not, perhaps we should deprecate it and remove it.

singularity-eos/eos/modifiers/ramps_eos.hpp
```diff
@@ -0,0 +1,239 @@
...
+
+  PORTABLE_INLINE_FUNCTION
+  int nlambda() const noexcept { return t_.nlambda(); }
+
+  PORTABLE_FUNCTION
+  unsigned long PreferredInput() const { return t_.PreferredInput(); }
+
+  PORTABLE_FUNCTION void PrintParams() const {
+    t_.PrintParams();
+    printf("r0=%e\na=%e\nb=%e\nc=%e\n", r0_, a_, b_, c_);
```
Yurlungur, line 209:
```suggestion
    printf("Ramp parameters: r0=%.14e\na=%.14e\nb=%.14e\nc=%.14e\n", r0_, a_, b_, c_);
```

Yurlungur (changes requested):
Overall looks good. Some nitpicks below. I also agree with @jhp-lanl that this needs tests of the ramp physics.

Also @jhp-lanl how do you feel about helping us fill out the ramp physics in the docs?

singularity-eos/eos/modifiers/ramps_eos.hpp
```diff
@@ -0,0 +1,239 @@
...
+               : t_.BulkModulusFromDensityTemperature(rho, temperature, lambda);
+  }
+  PORTABLE_FUNCTION
+  Real GruneisenParamFromDensityTemperature(const Real rho, const Real temperature,
+                                            Real *lambda = nullptr) const {
+    return t_.GruneisenParamFromDensityTemperature(rho, temperature, lambda);
+  }
+  PORTABLE_FUNCTION
+  void FillEos(Real &rho, Real &temp, Real &energy, Real &press, Real &cv, Real &bmod,
+               const unsigned long output, Real *lambda = nullptr) const {
```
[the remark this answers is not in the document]
Yurlungur: Oops. Already an issue. #135.

singularity-eos/eos/singularity_eos.hpp
```diff
@@ -27,44 +27,42 @@ extern "C" {
 int init_sg_eos(const int nmat, EOS *&eos);
 
 int init_sg_IdealGas(const int matindex, EOS *eos, const double gm1, const double Cv,
-                     int const *const enabled, double const *const vals);
+                     int const *const enabled, double *const vals);
```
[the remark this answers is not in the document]
Yurlungur: oops I asked for this and forgot. Resolved.

singularity-eos/eos/modifiers/ramps_eos.hpp
```diff
@@ -0,0 +1,239 @@
...
+    // output passed into internal filleos can't include pressure
+    const unsigned long ramp_out = output & ~thermalqs::pressure;
+    // if pressure is output, calculate it first
+    if (output & thermalqs::pressure) {
+      // maybe switch case on preferred input and check for not output of the input
+      // for now sie lookup is prioritized
+      if (!(output & thermalqs::specific_internal_energy)) {
+        press = this->PressureFromDensityInternalEnergy(rho, energy, lambda);
```
Yurlungur, line 182: Why not just call `PressureFromDensityInternalEnergy`? Why dereference `this`?

singularity-eos/eos/modifiers/ramps_eos.hpp
```diff
@@ -0,0 +1,239 @@
...
+      } else if (!(output & thermalqs::temperature)) {
+        press = this->PressureFromDensityTemperature(rho, temp, lambda);
+      }
+    }
+    // call internal filleos
+    t_.FillEos(rho, temp, energy, press, cv, bmod, ramp_out, lambda);
```
Yurlungur, line 188: Won't pressure be overwritten here?

singularity-eos/eos/modifiers/ramps_eos.hpp
```diff
@@ -0,0 +1,239 @@
...
+  SAPRampEOS() = default;
+
+  auto GetOnDevice() { return SAPRampEOS<T>(t_.GetOnDevice(), r0_, a_, b_, c_); }
+  inline void Finalize() { t_.Finalize(); }
+
+  PORTABLE_INLINE_FUNCTION
+  Real get_ramp_pressure(Real rho) const {
+    const Real p_ramp{rho < r0_     ? 0.0
+                      : rho < rmid_ ? a_ * (rho / r0_ - 1.0)
+                                    : b_ * (rho / r0_ - c_)};
```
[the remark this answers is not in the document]
dholladay00: There are asserts in the ctor, but perhaps we should disallow r0 = 0.

singularity-eos/eos/modifiers/ramps_eos.hpp
```diff
@@ -0,0 +1,239 @@
...
+    return p_ramp;
+  }
+
+  PORTABLE_INLINE_FUNCTION
+  Real get_ramp_density(Real P) const {
+    const Real rho_ramp{P < Pmid_ ? r0_ * (P / a_ + 1.0) : r0_ * (P / b_ + c_)};
```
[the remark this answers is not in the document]
dholladay00: there are asserts in the ctor for the ramp parameters.

singularity-eos/eos/modifiers/ramps_eos.hpp
dholladay00: notice that I pass in the `ramp_out` which ensures pressure is not an output value.

Yurlungur: ah ok I see.

singularity-eos/eos/modifiers/ramps_eos.hpp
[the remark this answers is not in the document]
Yurlungur: We should disallow r0=0 if we're dividing by it. But even small r0 could cause problems.

singularity-eos/eos/modifiers/ramps_eos.hpp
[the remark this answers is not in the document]
Yurlungur: You check for exactly zero. We should have division by zero guards because a small number divided by a small number can still give the wrong answer.

singularity-eos/eos/modifiers/ramps_eos.hpp
```diff
@@ -0,0 +1,239 @@
...
+
+template <typename T>
+void pAlpha2SAPRampParams(const T &eos, const Real alpha0, const Real Pe, const Real Pc,
+                          Real &r0, Real &a, Real &b, Real &c) {
+  // get reference conditions
+  Real rho0, T0, sie0, P0, cv0, bmod0, dpde0, dvdt0;
+  Real rmid, r1;
+  eos.ValuesAtReferenceState(rho0, T0, sie0, P0, cv0, bmod0, dpde0, dvdt0);
+  // calculate r0
+  r0 = rho0 / alpha0;
```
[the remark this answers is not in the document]
jhp-lanl: It should be greater than or equal to one, but I realize one of the tests currently relies on it being less than 1.

singularity-eos/eos/modifiers/ramps_eos.hpp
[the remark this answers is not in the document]
jhp-lanl: I believe the tests will cause `r0` to be fairly small since they are ramping an ideal gas. The basic idea is that $\rho_0 = \rho(P_0, T_\mathrm{ref}) / \alpha_0$. For an ideal gas, the density will be on order 1e-03.

singularity-eos/eos/modifiers/ramps_eos.hpp
dholladay00: We at least need the density at reference state in order for us to initialize the parameters from p-alpha parameters.

singularity-eos/eos/modifiers/ramps_eos.hpp
dholladay00: done

test/test_eos_unit.cpp
```diff
@@ -64,6 +66,38 @@ constexpr Real ev2k = 1.160451812e4;
...
+  REQUIRE(isClose(test_e.PressureFromDensityTemperature(1, 1),
+                  ref_e.PressureFromDensityTemperature(1, 1), 1.e-15));
+  REQUIRE(isClose(test_e.SpecificHeatFromDensityTemperature(1, 1),
+                  ref_e.SpecificHeatFromDensityTemperature(1, 1), 1.e-15));
+  REQUIRE(isClose(test_e.BulkModulusFromDensityTemperature(1, 1),
+                  ref_e.BulkModulusFromDensityTemperature(1, 1), 1.e-15));
+  REQUIRE(isClose(test_e.GruneisenParamFromDensityTemperature(1, 1),
+                  ref_e.GruneisenParamFromDensityTemperature(1, 1), 1.e-15));
+  REQUIRE(isClose(test_e.MinimumDensity(), ref_e.MinimumDensity(), 1.e-15));
+  REQUIRE(isClose(test_e.MinimumTemperature(), ref_e.MinimumTemperature(), 1.e-15));
```
jhp-lanl, line 97: This isn't a hold on the MR, but one thing I've been trying to do is use the Catch2 `INFO` function since `REQUIRE(isClose(...))` will just return `false` without any more helpful information. The `INFO` macro is a print statement that will execute only when the test fails so you can get some more information about why it's failing. This has _massively_ helped in my debugging of the Gruneisen EOS because just knowing the two values can help me determine what's going crazy.

The other thing to consider is using `CHECK` rather than `REQUIRE` for a big block like this. `CHECK` fails the test still if the condition is false, but it won't end the test at the first failure. This can be helpful for debugging when you can see a pattern in how multiple values are off.

test/test_eos_unit.cpp
```diff
@@ -193,14 +227,90 @@ SCENARIO("EOS Builder and Modifiers", "[EOSBuilder],[Modifiers][IdealGas]") {
...
+        // check bulk moduli on both pieces of ramp
+        REQUIRE(
+            isClose(bmodrt1, igra.BulkModulusFromDensityTemperature(rho_t1, T0), 1.e-12));
+        REQUIRE(
+            isClose(bmodrt2, igra.BulkModulusFromDensityTemperature(rho_t2, T0), 1.e-12));
+        // check bulk modulus below and beyond ramp matches unmodified ideal gas
+        REQUIRE(isClose(0.8 * r0 * gm1 * (gm1 + 1.0) * Cv * T0,
+                        igra.BulkModulusFromDensityTemperature(0.8 * r0, T0), 1.e-12));
+        REQUIRE(isClose(1.2 * r1 * gm1 * (gm1 + 1.0) * Cv * T0,
+                        igra.BulkModulusFromDensityTemperature(1.2 * r1, T0), 1.e-12));
```
jhp-lanl, line 313: Can you break this up into multiple `THEN` statements? I think you just need to transform your comments into `THEN` statements so that you have five tests instead of one long one.

jhp-lanl (changes requested):
Small changes

test/test_eos_unit.cpp
```diff
@@ -64,6 +66,38 @@ constexpr Real ev2k = 1.160451812e4;
...
+  REQUIRE(isClose(test_e.BulkModulusFromDensityTemperature(1, 1),
+                  ref_e.BulkModulusFromDensityTemperature(1, 1), 1.e-15));
+  REQUIRE(isClose(test_e.GruneisenParamFromDensityTemperature(1, 1),
+                  ref_e.GruneisenParamFromDensityTemperature(1, 1), 1.e-15));
+  REQUIRE(isClose(test_e.MinimumDensity(), ref_e.MinimumDensity(), 1.e-15));
+  REQUIRE(isClose(test_e.MinimumTemperature(), ref_e.MinimumTemperature(), 1.e-15));
+  return;
+}
+
 SCENARIO("Test that fast logs are invertible and run on device", "[FastMath]") {
```
jhp-lanl, line 101: It looks like a bunch of these tests have nothing to do with the fast logarithms even if they're grouped there. I'm going to create an issue to break these out

singularity-eos/eos/modifiers/ramps_eos.hpp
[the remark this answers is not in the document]
dholladay00: I'd say that number is not small relative to machine epsilon. I am wary of judging relative orders of magnitude in a way that prevents users from running the code without a special build. I will try to add @Yurlungur's smart divides thing, but I do not want to further restrict densities besides r0 needs to be a positive and non-zero floating point value.

singularity-eos/eos/modifiers/ramps_eos.hpp
```diff
@@ -0,0 +1,239 @@
...
+
+  PORTABLE_INLINE_FUNCTION
+  Real get_ramp_density(Real P) const {
+    const Real rho_ramp{P < Pmid_ ? r0_ * (P / a_ + 1.0) : r0_ * (P / b_ + c_)};
+    return rho_ramp;
+  }
+
+  PORTABLE_INLINE_FUNCTION
+  Real get_ramp_dpdrho(Real rho) const {
+    const Real dpdr{rho < r0_ ? 0.0 : rho < rmid_ ? a_ / r0_ : b_ / r0_};
```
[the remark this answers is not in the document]
dholladay00: I'll try to implement smart/robust divide, but I still prefer to use the ternary where possible. *I* had formatted the ternary in a readable way, but I'm not allowed to format. And I don't want to restrict r0 beyond the > 0 constraint.

jonahm-LANL:
If you have a way you want to format, you can use
 // clang-format disable
On a piece of code.

Sent with BlackBerry Work
(www.blackberry.com)


From: Daniel Holladay ***@***.******@***.***>>
Date: Friday, Jul 01, 2022, 11:14 AM
To: lanl/singularity-eos ***@***.******@***.***>>
Cc: Subscribed ***@***.******@***.***>>
Subject: [EXTERNAL] Re: [lanl/singularity-eos] Dholladay00/sap ramp (PR #101)


@dholladay00 commented on this pull request.

________________________________

In singularity-eos/eos/modifiers/ramps_eos.hpp<https://urldefense.com/v3/__https://github.com/lanl/singularity-eos/pull/101*discussion_r912128188__;Iw!!Bt8fGhp8LhKGRg!EGzueB5w418lcM_PRvqsfAKthTD4veOegN7RxrvbBp61DRNSR8HtzW1OouRyR1_67qVCOt_PmuOExDzQeLh4sdi9$>:

> +  Real get_ramp_pressure(Real rho) const {
+    const Real p_ramp{rho < r0_     ? 0.0
+                      : rho < rmid_ ? a_ * (rho / r0_ - 1.0)
+                                    : b_ * (rho / r0_ - c_)};
+    return p_ramp;
+  }
+
+  PORTABLE_INLINE_FUNCTION
+  Real get_ramp_density(Real P) const {
+    const Real rho_ramp{P < Pmid_ ? r0_ * (P / a_ + 1.0) : r0_ * (P / b_ + c_)};
+    return rho_ramp;
+  }
+
+  PORTABLE_INLINE_FUNCTION
+  Real get_ramp_dpdrho(Real rho) const {
+    const Real dpdr{rho < r0_ ? 0.0 : rho < rmid_ ? a_ / r0_ : b_ / r0_};


I'll try to implement smart/robust divide, but I still prefer to use the ternary where possible. I had formatted the ternary in a readable way, but I'm not allowed to format. And I don't want to restrict r0 beyond the > 0 constraint.

test/test_eos_unit.cpp
dholladay00: done.

test/test_eos_unit.cpp
dholladay00: done.

jhp-lanl (approved):
Looks good to me. @Yurlungur has @dholladay00 satisfied your comments?

singularity-eos/eos/modifiers/ramps_eos.hpp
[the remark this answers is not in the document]
dholladay00: done

singularity-eos/eos/modifiers/ramps_eos.hpp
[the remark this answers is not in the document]
dholladay00: now using robust divide.

singularity-eos/eos/modifiers/ramps_eos.hpp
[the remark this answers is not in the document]
dholladay00: now using robust divide

singularity-eos/eos/modifiers/ramps_eos.hpp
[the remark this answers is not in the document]
dholladay00: now using robust divide

singularity-eos/eos/modifiers/ramps_eos.hpp
```diff
@@ -0,0 +1,239 @@
...
+  // calculate r1
+  auto r1_func = [&](const Real x) { return eos.PressureFromDensityTemperature(x, T0); };
+  // finds where r1_func = Pc
+  RootFinding1D::findRoot(r1_func, Pc, rmid, r0, rho_ub, 1.e-12, 1.e-12, r1, co);
+  // a
+  a = r0 * Pe / (rmid - r0);
+  // b
+  b = r0 * (Pc - Pe) / (r1 - rmid);
+  // c
+  c = (Pc * rmid - Pe * r1) / (r0 * (Pc - Pe));
```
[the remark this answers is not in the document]
dholladay00: I have written up the math for converting from p-alpha to sap ramp.

Yurlungur (approved):
Assuming all tests pass on Darwin, yep this looks good to me.

Yurlungur:
Thanks for all your hard work on this @dholladay00 can you submit a PR on Darwin to make sure tests pass there?

dholladay00:
gpu tests failing, but not in places I changed. I'll take a look next week, DO NOT MERGE before hand. I think it could be issues with doing emplacement of the ramp params in the builder test and the value category of the indexer bracket operator in the pte test.

dholladay00:
@Yurlungur @jhp-lanl gitlab ci fixes should work, I say once the latest ci finishes its good to merge.
Addenda:
@Yurlungur I added device markings to a couple spiner member functions to suppress warnings, may be worth checking out before clicking the button. 
@mauneyc-LANL maybe double-check my minor cmake changes are reasonable.

Yurlungur:
@dholladay00 is there a Spiner MR? Or do you mean SpinerEOS?

singularity-eos/eos/eos.hpp
```diff
@@ -528,9 +529,12 @@ class SpinerEOSDependsRhoT : public EosBase<SpinerEOSDependsRhoT> {
...
   PORTABLE_FORCEINLINE_FUNCTION Real MinimumDensity() const { return rhoMin(); }
   PORTABLE_FORCEINLINE_FUNCTION Real MinimumTemperature() const { return T_(lTMin_); }
+  PORTABLE_INLINE_FUNCTION
   int nlambda() const noexcept { return _n_lambda; }
-  inline RootFinding1D::Status rootStatus() const { return status_; }
-  inline TableStatus tableStatus() const { return whereAmI_; }
+  PORTABLE_INLINE_FUNCTION
+  RootFinding1D::Status rootStatus() const { return status_; }
+  PORTABLE_INLINE_FUNCTION
+  TableStatus tableStatus() const { return whereAmI_; }
```
Yurlungur, line 537: Making these `PORTABLE_INLINE_FUNCTION` won't do any harm. It's worth noting though that when these functions are called in a parallel context, it's undefined behavior.

test/test_eos_unit.cpp
```diff
@@ -64,6 +66,68 @@ constexpr Real ev2k = 1.160451812e4;
 #endif // SINGULARITY_TEST_SESAME
 #endif // SPINER_USE_HDF
 
+template <typename E1, typename E2>
+inline void compare_two_eoss(E1 &&test_e, E2 &&ref_e) {
```
Yurlungur, line 70: Not a problem, but worth noting that this will only work on CPU context.

Yurlungur (approved):
LGTM.

Yurlungur:
Happy to merge upon gitlab success and approval by @mauneyc-LANL .

dholladay00:
@Yurlungur I just remembered that @jhp-lanl had requested a name change `SAPRamp` -> `BilinearRamp` I'll start working on that.

dholladay00:
> @Yurlungur I just remembered that @jhp-lanl had requested a name change `SAPRamp` -> `BilinearRamp` I'll start working on that.

This change has been completed, CI running now.

PR #102, opened 2022-04-15
Fast, approximate and invertible logs

Yurlungur:
## PR Summary

This PR emerged from a discussion with @jdolence . @jdolence noticed that when using fast logs, the following was not true:
```C++
std::abs(pow(10., singularity::Math::Log10(x)) - x) < EPSILON
```
for some `EPSILON`. This was because we do fast logs, but we don't have a fast exponential to undo it in the exact same, approximate way. This may have implications for the accuracy of `P(rho,sie)` calls for `SpinerEOSDependsRhoT` though it should be a small effect.

To resolve this, @jdolence and I revisited the idea of log-linear interpolation. The reason we use log space is because logs are easy, and the grid spacing for a logarithmically spaced grid is roughly proportional to the size of the independent variable. *It doesn't actually matter that it be a logarithm exactly.* What matters is that we have a function to map into "grid space" which is approximately logarithmic and is fast and invertible.

This leads me to build off the fast log method I had previously implemented with @dholladay00 . `frexp` is invertible. Just as `frexp` picks apart the internal representation of a floating point number into mantissa and exponent (in base 2), there's an expression, `ldexp`, which reconstructs the floating point number from the mantissa and exponent. Thus we just need an invertible, continuous, differentiable function which approximately and quickly maps the mantissa `m` to `lg(m`). It turns out `2*(m - 1)` does nicely.

So that's what this PR introduces:
1. Replace the old fast `log10` with an even faster, more approximate `log10` which is easier to invert
2. Introduce a fast `pow10`, which exactly inverts the fast `log10`, up to almost machine precision.
3. Adds a test that proves this invertibility.
4. In `sesame2spiner` and `SpinerEOS`, use the fast `log10` and fast `pow10` everywhere when going to and from grid space, to ensure that we're consistent everywhere.
5. **Do not** use the fast math versions for `StellarCollapse`, as the stellar collapse tables are not re-interpolated, and we need to be faithful to the original log-log representation in the table. A future change would be to re-interpolate these tables so we can use fast math.
6. Document these changes thoroughly in comments and in the sphinx docs, because they're not intuitive. I also took the opportunity to fill in a little bit of missing documentation about models.

I am running the tests on Darwin, and I also plan to do a quick and dirty performance/accuracy comparison between the new data and the old. @dholladay00 and @annapiegra at your convenience (not necessarily before merge) it would be nice to run this through the spiner-eospac comparison pipeline.

I also ping @brryan and @lroberts36 because perhaps we should take this approach in the stellar collapse tables, as well as opacities in singularity-opac for Phoebus as well.

## PR Checklist

### DIFF
--- a/doc/sphinx/src/contributing.rst
+++ b/doc/sphinx/src/contributing.rst
@@ -2,3 +2,74 @@
 
 Contributing
 =============
+
+
+Fast Logs and Approximate Log Gridding
+---------------------------------------
+
+When spanning many orders of magnitude, Logarithmic grids are a
+natural choice. Even spacing in log space corresponds to exponential
+spacing in the original linear space. In other words, the grid spacing
+is proportional to the value of the independent variable.
+
+One can perform log-linear or log-log interpolation by simply
+converting to log space, interpolating as one normally would, and then
+converting back out. Unfortunately, logarithms and exponents are
+transcendental functions, meaning they are expensive to compute and it
+is thus expensive to transform in and out of log space.
+
+To avoid this issue, we construct a space that is *approximately*
+logarithmically spaced, but not quite exactly. The requirements for
+this space are that the transformation into and out of this space is
+fast to compute, continuous, differentiable, analytically invertible,
+and close to taking a logarithm or exponentiation (depending on which
+way you're going).
+
+To achieve this, we leverage the internal representation of a floating
+point number in the IEE standard. In particular, a floating point
+number :math:`x` is represented as a mantissa and an exponent in base
+2:
+
+.. math::
+
+   x = m 2^e
+
+for mantissa :math:`m` and exponent :math:`e`. The mantiss is
+guaranteed to be on the interval :math:`[1/2, 1)`. The standard
+library of most low-level languages provides a performant and portable
+routine to pick apart this represnetation, ``frexp``, which given a
+number :math:`x`, return :math:`m` and :math:`e`.
+
+The log in base 2 ``lg`` of :math:`x` is then given by the logarithm
+of the mantissa plus the exponent:
+
+.. math::
+
+   \lg(x) = \lg(m) + e
+
+Therefore, if we can find a fast, invertible approximation to
+:math:`\lg(m)`, we will have achieved our goal. It turns out the
+expression
+
+.. math::
+
+   2 (x - 1)
+
+works pretty well, so we use that. (To convince yourself of this note
+that for :math:`x=1/2` this expression returns -1 and for :math:`x=1`,
+it returns 0, which are the correct values of :math:`\lg(x)` at the
+bounds of the interval.) Thus our approximate, invertible expression
+for :math:`\lg` is just
+
+.. math::
+
+   2 (m - 1) + e
+
+for the mantissa and exponent extracted via ``frexp``. The absolute
+error of this approximation is about 0.1 at its largest, which
+translates to a relative error of at most 25 percent. This error is
+acceptable, for the reasoning stated above.
+
+To invert, we use the built in function that inverts ``frexp``,
+``ldexp``, which combines the mantissa and exponent into the original
+floating point representation.
--- a/doc/sphinx/src/contributing.rst
+++ b/doc/sphinx/src/contributing.rst
@@ -65,10 +65,11 @@ for :math:`\lg` is just
 
    2 (m - 1) + e
 
-for the mantissa and exponent extracted via ``frexp``. The absolute
-error of this approximation is about 0.1 at its largest, which
-translates to a relative error of at most 25 percent. This error is
-acceptable, for the reasoning stated above.
+for the mantissa and exponent extracted via ``frexp``. This differs
+from :math:`lg` by a maximum of about 0.1, which translates to at most
+a 25 percent difference. As discussed above, however, the function
+itself is an exact representation of itself and the difference from
+:math:`lg` is acceptable.
 
 To invert, we use the built in function that inverts ``frexp``,
 ``ldexp``, which combines the mantissa and exponent into the original
--- a/utils/fast-math/logs.hpp
+++ b/utils/fast-math/logs.hpp
@@ -31,7 +31,7 @@
  *
  * 2. The function must be invertible, meaning you can go back to
  * "linear space." Ideally it is also continuous.
- * 
+ *
  * 3. The function and its inverse must be fast.
  *
  * To meet these constraints, we approximate a logarithm by using
@@ -50,7 +50,7 @@
  *
  * So that mantissa = 0.5 returns log_2(0.5) = -1 and mantissa (1)
  * yields 0.
- * 
+ *
  * Then the approximate log of the whole thing is the approximation of
  * the log of the mantissa, plus the exponent:
  *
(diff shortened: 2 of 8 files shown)


Yurlungur:
Documentation on the fast log strategy can be found here: https://lanl.github.io/singularity-eos/jmm/fast-invertible-logs/src/contributing.html#fast-logs-and-approximate-log-gridding

Yurlungur:
I merged in #103 so that should be merged in before this one. Merging that one in and then rebasing this against main will also simplify the diff.

Yurlungur:
Timing comparison between the old logs and the new fast log/exponentials. Performed on a 512 squared grid, using 4272.

## old
```
                ...spiner(rho,T) time/point (microseconds) host = 0.358668
                ...spiner(rho,sie) time/point (microseconds) host = 0.212383
                ...spiner(rho,T) time/point (microseconds) device = 0.0015873
                ...spiner(rho,sie) time/point (microseconds) device = 0.000243379
```

## new
```
              ...spiner(rho,T) time/point (microseconds) host = 0.290972
                ...spiner(rho,sie) time/point (microseconds) host = 0.169928
                ...spiner(rho,T) time/point (microseconds) device = 0.00156937
                ...spiner(rho,sie) time/point (microseconds) device = 0.000245286
```

So the fast logs introduce a performance boost of 0.5 microseconds or so on CPU. On GPU, apparently there's no change. This probably implies the test at this scale is not flop limited on the GPU.

doc/sphinx/src/contributing.rst
```diff
@@ -2,3 +2,74 @@
...
+bounds of the interval.) Thus our approximate, invertible expression
+for :math:`\lg` is just
+
+.. math::
+
+   2 (m - 1) + e
+
+for the mantissa and exponent extracted via ``frexp``. The absolute
+error of this approximation is about 0.1 at its largest, which
+translates to a relative error of at most 25 percent. This error is
```
jdolence, line 70: This discussion of error could easily be misinterpreted.  There are, in fact, no errors of note in this except the typical precision issues.  Rather, there are differences between this function and the actual `log`, but that's not an error.  We're not really trying to approximate `log10` and introducing error in the approximation.  Instead we're just using a different function that looks a bit like `log10`, and the similarity is useful to point out to make this strange function easier to reason about.

singularity-eos/eos/eos_base.hpp
```diff
@@ -68,11 +68,11 @@ class EosBase {
                                        LambdaIndexer &&lambdas) const {
     static auto const name = SG_MEMBER_FUNC_NAME();
     static auto const cname = name.c_str();
+    CRTP copy = *(static_cast<CRTP const *>(this));
```
jdolence, line 71: good idea moving this out of the loop.

utils/fast-math/logs.hpp
```diff
@@ -19,50 +17,98 @@
...
+ * the log of the mantissa, plus the exponent:
+ *
+ * 2*(mantissa - 1) + exponent
+ *
+ * Which is a continuous, invertible function that approximates lg
+ * only relatively accurately. The absolute error of this
+ * approximation is about 0.1 at its largest. This translates to a
+ * relative error of at most 25%.
+ *
+ * However, we don't mind this error, as we're only interested in
```
jdolence, line 64: this is a bit better than above, but worth saying again that the error you're referring to here is not actually any error in what we're doing.

utils/fast-math/logs.hpp
```diff
@@ -19,50 +17,98 @@
...
-  // log10 x using frexp
-  return ILOG10 * expr + n * LOG2OLOG10;
+  return ldexpf(mantissa, exponent);
+#endif // SINGULARITY_USE_SINGLE_LOGS
+}
+
+PORTABLE_FORCEINLINE_FUNCTION
+double pow10(const double x) {
+  constexpr double LOG10OLOG2 = 3.321928094887362626;
+  return pow2(LOG10OLOG2 * x);
```
jdolence, line 108: if only we didn't have these extra flops...

jdolence (approved):
just a couple comments.  mostly about your descriptions, not your code.

singularity-eos/eos/eos_base.hpp
Yurlungur: This actually resolves #104 and comes from #105. We can't capture the `*this` pointer in a gpu kernel.

Yurlungur:
@dholladay00 gave me the go-ahead verbally last week, assuming tests pass. If no one else has opinions, I will merge shortly.

@jhp-lanl @chadmeyer @annapiegra @brryan @lroberts36 now's your chance to weigh in. I'll merge around mid-day tomorrow (Wednesday.)

jhp-lanl (approved):
Looks good. And thanks for increasing our documentation!

lroberts36:
LGTM. I think it should be pretty straightforward to rebuild the StellarCollapse tables with a different grid with Andre's SRO code.

annapiegra:
Looks good.

annapiegra (approved):

chadmeyer:
I'm wondering whether it makes sense to include some option to replace the fast exp/log code with exact exp/log code in some automated way (i.e. at build time) for testing purposes.  I can imagine wanting to do fp-precise kinds of things (even if as comparisons).

Yurlungur:
> I'm wondering whether it makes sense to include some option to replace the fast exp/log code with exact exp/log code in some automated way (i.e. at build time) for testing purposes. I can imagine wanting to do fp-precise kinds of things (even if as comparisons).

This is easy to add, and I'll go ahead and do that. Worth emphasizing though that everything *is* fp-precise. All that replacing the fast math with exact math does is change the grid spacing. But changing the grid spacing could be useful for comparisons.

Yurlungur:
@chadmeyer take a look now. I added a `SINGULARITY_USE_TRUE_LOG_GRIDDING` configure parameter. So now the user can select between single and double precision and gridding that's exactly logarithmic or the new faster gridding.

utils/fast-math/logs.hpp
```diff
@@ -19,50 +17,133 @@
...
+  // faster but less accurate
+  const float y = frexpf((float)x, &n);
+#endif // SINGULARITY_USE_SINGLE_LOGS
+  return 2 * (y - 1) + n;
+}
 
-  // const double ILOG10 = 1./std::log(10.0);
-  constexpr double ILOG10 = 0.43429448190325176;
+PORTABLE_FORCEINLINE_FUNCTION
+double fastpow2(const double x) {
```
lroberts36, line 88: Should there be some check that `x` is positive here and in `fastlg`?

Yurlungur: I can add an assert maybe, but I'd rather leave that to higher-level in the codes. This particular piece of code is supposed to be performant.

Yurlungur: Done.

Yurlungur:
Assuming no more comments, I will merge as soon as tests pass on Darwin, which will require Darwin being out of DST.

utils/fast-math/logs.hpp
lroberts36: I guess the concern is that it gives back a number that is nonsense but probably doesn't look super wrong. The issue is only in `fastlg` I think (I was wrong before about `fastpow2` I think, this should work fine for negative powers), so I guess the user could be assumed to know you shouldn't pass a negative number to a log function.

Yurlungur: > I guess the user could be assumed to know you shouldn't pass a negative number to a log function.

I think this is reasonable to assume, but also no harm in an assert, as they are pre-processed away in release builds. I've added an assert.

Yurlungur:
I have relative confidence everything in here is now correct. However, the Darwin CI is still not passing, as the recent darwin updates seem to have broken our spack buildout. I ran the tests myself, and they all pass, so I am going to merge so development can move forward. That said, we should pay special attention to the CI once it's repaired.

Issue #105, opened 2022-04-15
Vector interface currently does not work on x86 nodes with Volta GPUs


Yurlungur:
Just discovered. This is obviously a serious issue.

PR #107, opened 2022-04-16
Jmm/compare to eospac no ses2spiner

Yurlungur:
## PR Checklist

Quick bug fix for `compare_to_eospac.cpp`. Also cleans it up by not using `io_eospac.hpp` to provide the `Bounds` class.

### DIFF
--- a/test/compare_to_eospac.cpp
+++ b/test/compare_to_eospac.cpp
@@ -62,8 +62,7 @@ class Bounds {
   Bounds(Real min, Real max, int N, Real offset)
       : grid(RegularGrid1D(min, max, N)), offset(offset) {}
 
-  Bounds(Real min, Real max, int N)
-      : offset(0) {
+  Bounds(Real min, Real max, int N) : offset(0) {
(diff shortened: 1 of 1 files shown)


Yurlungur:
Depends on #105. Do not merge until that's in.

Yurlungur:
I don't think anyone cares about this PR but me. Merging by fiat.

dholladay00:
is the bounds class now defined in 2 places?

Yurlungur:
Yes, and the implementations are slightly different. The `sesame2spiner` one should get namespaced. For now however it's not an issue as `compare_to_eospac` no longer uses `sesame2spiner`.

dholladay00:
could we put the implementation in a header included in both places to prevent code duplication?

Yurlungur:
Yes technically, but it would require me abstracting out some functionality. The issue is the one in `sesame2spiner` uses the new fast logs, whereas the one in `compare_to_eospac` uses `std::log10`. I don't think it's worth worrying about.

PR #108, opened 2022-04-18
Fix kokkos-kernels depends condition

mauneyc-LANL:
The fix for older spack versions would append `kokkos-kernels` to specs that did not specify it. Here, a `when` is added to ensure it is only included when requested in the spec.

### DIFF
--- a/spack-repo/packages/singularity-eos/package.py
+++ b/spack-repo/packages/singularity-eos/package.py
@@ -62,7 +62,7 @@ class SingularityEos(CMakePackage, CudaPackage):
 
     # fix for older spacks
     if spack.version.Version(spack.main.get_version()) >= spack.version.Version("0.17"):
-        depends_on("kokkos-kernels ~shared")
+        depends_on("kokkos-kernels ~shared", when="+kokkos-kernels")
 
     for _flag in list(CudaPackage.cuda_arch_values):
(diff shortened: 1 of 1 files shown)


Yurlungur (approved):
I don't fully understand the logic here. Is it just "use the when syntax when appropriate and it's available in the sufficiently new version of spack"?

If so that seems good. Approved.

mauneyc-LANL:
> I don't fully understand the logic here. Is it just "use the when syntax when appropriate and it's available in the sufficiently new version of spack"?

The logic here was "if `spack` version is `>=17`, then depend on `kokkos-kernels ~shared`", when what it is supposed to be is "if `spack` version is `>=17`, AND we depend on `kokkos-kernels` then depend on `kokkos-kernels ~shared`"

What was happening for my case was that the default install spec (i.e. not using `cuda` + `kokkos`) was bringing in `kokkos-kernels` anyway

Yurlungur:
I see... And the versioning of >= 17 means the `when` syntax is supported?

mauneyc-LANL:
> I see... And the versioning of >= 17 means the `when` syntax is supported?

`when` is supported as a keyword arg in `depends_on` for a while now (since at least it was adopted by LANL). The specific versioning is due to the issue we encountered when trying to install/link with a `kokkos-kernels` as a shared library with newer spack versions

Yurlungur:
Got it. Thanks for the explanation.

PR #106, opened 2022-04-16
Fix vector interface on GPUs

Yurlungur:
## PR Summary

It turns out the vector interface on GPUs was only working on Power9s thanks to undefined behavior. This PR fixes it. The issue was that the `*this` pointer was being captured in the lambdas in the portable for in the base class.

To rectify, I dereference the `*this` pointer to make a copy of the derived class to pass into the `portableFor`.

To correctly test this, we need a CI build matrix for x86-volta nodes. I have added one, but some extra eyes from, e.g., @ktsai7, @junghans , or @dholladay00  would be welcome.

## PR Checklist

- [x] Adds a test for any bugs fixed. Adds tests for new features.
- [x] Format your changes by using the `make format` command after configuring with `cmake`.
- [x] Document any new features, update documentation for changes made.
- [x] Make sure the copyright notice on any files you modified is up to date.

PR #109, opened 2022-04-19
Report Minimum Density and Temeprature in an EOS

Yurlungur:
## PR Summary

@jdolence is currently working on solving issues with PTE solvers and reported that it would be useful to know the minimum values of the independent variables in a tabulated EOS.

This machinery provides this. We now report minimum density and temperature for all EOS's. For analytic EOS's, this is provided by a new method in the base class, which simply returns zero. (I assume here we will never be working with antimatter or the population inversions of a laser gain medium which might produce negative densities or temperatures respectively. ;) )

The tabulated EOS's and the modifiers implement their own versions as appropriate.

I believe there may be more to attack here, as far as what bounds to report. But this is a start.

As with all my pull requests as of late... Depends on #106 so that should go in first.

## PR Checklist

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -63,7 +63,7 @@ before_script:
     DO_GPU: "true"
 
 .x86volta: &x86volta
-    SCHEDULER_PARAMETERS: "--nodes=1 --partition=x96-volta"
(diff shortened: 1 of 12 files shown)


jhp-lanl (approved):
This looks good to me I think. Obviously there's a lot of changes that will go away with the vector fix MR but I think I see what's separate from that and it's pretty simple.

PR #114, opened 2022-04-21
changed defines to avoid collisions downstream

mauneyc-LANL:
Change the `#define`s in `utils/root-finding-1d/root_finding.hpp` to be prefixed with `SINGULARITY_`, to avoid possible collisions with names/symbols downstream

There is also a presently in the code I'm not sure about, see Summary

## PR Summary

As I understand it, this root-find shares a common origin in GSL; it is no only possible that other codes do so, but I in fact hit an issue when integrating `singularity-eos` into a code that does something similar. The results were lots of weird errors that ended up being these preprocessor directives over-writing variables names.

To prevent this in future, this PR changes the strings of these `#define`s to try and avoid possible future collisions.

There is also a possible unnecessary `#define` (originally `utils/root-finding-1d/root_finding.hpp:34`) :
```c++
#define ROOT_NAN_OK (0)
```
This isn't referenced in the file, and further isn't unset. Is this necessary for something outside the source tree? If we don't need it anywhere else, it would be good to remove it altogether.

NB: _Ideally_ these would be declared as `constexpr` and placed in namescope, but currently without `if constexpr` functionallity, this would introduce unwanted code branching. If we do ever decide to bump the C++ standard, this would be the preferred change. 

## PR Checklist

- [ x] Adds a test for any bugs fixed. Adds tests for new features.
- [ x] Format your changes by using the `make format` command after configuring with `cmake`.
- [ x] Document any new features, update documentation for changes made.
- [ x] Make sure the copyright notice on any files you modified is up to date.

### DIFF
--- a/utils/root-finding-1d/root_finding.hpp
+++ b/utils/root-finding-1d/root_finding.hpp
@@ -31,9 +31,9 @@
 
 #define SINGULARITY_ROOT_DEBUG (0)
 #define SINGULARITY_ROOT_VERBOSE (0)
-//TODO: mauneyc: this isn't used here, and I can't
-//find it used elsewhere. it also doesn't get unset.
-//should it be ditched?
+// TODO: mauneyc: this isn't used here, and I can't
+// find it used elsewhere. it also doesn't get unset.
+// should it be ditched?
 #define SINGULARITY_ROOT_NAN_OK (0)
 
 #define SINGULARITY_MY_SIGN(x) (x > 0) - (x < 0)
--- a/utils/root-finding-1d/root_finding.hpp
+++ b/utils/root-finding-1d/root_finding.hpp
@@ -29,11 +29,14 @@
 #include <math.h>
 #include <stdio.h>
 
-#define ROOT_DEBUG (0)
-#define ROOT_VERBOSE (0)
-#define ROOT_NAN_OK (0)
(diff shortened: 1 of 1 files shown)


utils/root-finding-1d/root_finding.hpp
```diff
@@ -29,11 +29,14 @@
...
 
-#define ROOT_DEBUG (0)
-#define ROOT_VERBOSE (0)
-#define ROOT_NAN_OK (0)
+#define SINGULARITY_ROOT_DEBUG (0)
+#define SINGULARITY_ROOT_VERBOSE (0)
+// TODO: mauneyc: this isn't used here, and I can't
+// find it used elsewhere. it also doesn't get unset.
+// should it be ditched?
+#define SINGULARITY_ROOT_NAN_OK (0)
```
Yurlungur, line 37: Yes let's ditch it. This is apparently old code.

Yurlungur (approved):
Thanks for this!

Yurlungur:
> NB: Ideally these would be declared as constexpr and placed in namescope, but currently without if constexpr functionallity, this would introduce unwanted code branching. If we do ever decide to bump the C++ standard, this would be the preferred change.

Agreed. I think it's only a matter of time until we upgrade C++ standards, but I'm not comfortable doing so yet.

PR #115, opened 2022-04-29
This commit introduces a new way to construct the variant list piece …

dholladay00:
…by piece. One can choose which eoss get specific modifiers and the ordering and nestedness of modifiers and combine all of the individual lists at the end to obtain the final model list.

## PR Checklist

### DIFF
--- a/singularity-eos/base/variadic_utils.hpp
+++ b/singularity-eos/base/variadic_utils.hpp
@@ -1,0 +1,193 @@
+//------------------------------------------------------------------------------
+// © 2021-2022. Triad National Security, LLC. All rights reserved.  This
+// program was produced under U.S. Government contract 89233218CNA000001
+// for Los Alamos National Laboratory (LANL), which is operated by Triad
+// National Security, LLC for the U.S.  Department of Energy/National
+// Nuclear Security Administration. All rights in the program are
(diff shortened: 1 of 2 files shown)


Yurlungur (approved):
I like this a lot. Is there a way we could test it? Just to see that it's all working? Maybe an optional "print the types" in the `eos_unit_tests` or something?

Also what's the effect on compile times?

dholladay00:
> I like this a lot. Is there a way we could test it? Just to see that it's all working? Maybe an optional "print the types" in the `eos_unit_tests` or something?
> 
> Also what's the effect on compile times?

@Yurlungur I have tested this on snow with intel and gcc, <5% hit on compile times. On power9 with all of the options toggled, I saw a 28s increase in compile time on a roughly 9-9.5 minute compile time (about 5%).

test/eos_unit_tests.cpp
```diff
@@ -203,6 +231,12 @@ SCENARIO("Rudimentary test of the root finder", "[RootFinding1D]") {
   }
 }
 
+SCENARIO("EOS Variant Type", "[Variant][EOS]") {
+  // print out the eos type
+  std::cout << demangle(typeid(EOS).name()) << std::endl;
+}
+
+
```
Yurlungur, line 239: :+1: I like it

PR #124, opened 2022-05-24
Add Python interface

rbberger:
## PR Summary

This PR adds a first implementation of Python bindings to Singularity-EOS using the [pybind11](https://pybind11.readthedocs.io/en/stable/) library.
It adds an example that showcases the currently supported functionality and sufficient annotations to name arguments.

```python
from singularity_eos import IdealGas
eos = IdealGas(gm1, Cv)
...
```

## PR Checklist


.gitmodules
```diff
@@ -10,3 +10,7 @@
 [submodule "utils/kokkos"]
 	path = utils/kokkos
 	url = https://github.com/kokkos/kokkos.git
+[submodule "utils/pybind11"]
+	path = utils/pybind11
+	url = https://github.com/pybind/pybind11
+	branch = stable
```
Yurlungur, line 16: Is PyBind as a submodule the preferred way to do this? Or should we use, e.g., fetch content through cmake?

python/module.cpp
```diff
@@ -0,0 +1,96 @@
...
+    .def("SpecificHeatFromDensityTemperature", &two_params<T, &T::SpecificHeatFromDensityTemperature>, py::arg("rho"), py::arg("temperature"), py::arg("lambda"))
+    .def("SpecificHeatFromDensityInternalEnergy", &two_params<T, &T::SpecificHeatFromDensityInternalEnergy>, py::arg("rho"), py::arg("sie"), py::arg("lambda"))
+    .def("BulkModulusFromDensityTemperature", &two_params<T, &T::BulkModulusFromDensityTemperature>, py::arg("rho"), py::arg("temperature"), py::arg("lambda"))
+    .def("BulkModulusFromDensityInternalEnergy", &two_params<T, &T::BulkModulusFromDensityInternalEnergy>, py::arg("rho"), py::arg("sie"), py::arg("lambda"))
+    .def("GruneisenParamFromDensityTemperature", &two_params<T, &T::GruneisenParamFromDensityTemperature>, py::arg("rho"), py::arg("temperature"), py::arg("lambda"))
+    .def("GruneisenParamFromDensityInternalEnergy", &two_params<T, &T::GruneisenParamFromDensityInternalEnergy>, py::arg("rho"), py::arg("sie"), py::arg("lambda"))
+
+    // TODO .def("GetOnDevice")
+    // TODO .def("FillEos")  what are the call semantics, since it uses references? return dict?
+    // TODO .def("ValuesAtReferenceState")
```
Yurlungur, line 46: For the functions with reference semantics, what about returning a struct? We can also skip them, as they are (mostly) just combining capability for the functions that return a single value.

Also the vector calls may be redundant with the functions with reference semantics at least for our purposes.

.gitmodules
rbberger: It was just the easiest to do for a quick prototype. I can definitely change it to use FetchContent instead.

Yurlungur: I think that's desirable, at least for now. We've been trying to move away from submodules when possible.

python/module.cpp
jhp-lanl: I assume a struct would map to python class of some sort?

If I were doing this in pure python, I might create a class called `EosState` that would instantiate with uninitialized values. I could set a couple of the values and then pass the whole thing to `FillEos` and detect the uninitialized values to be outputs.

jhp-lanl: Re the vector functions:
I don't know about with `pybind`, but I'm pretty sure there are ways to use some sort of python arrays with reference semantics. I've done it in cython at least.

Yurlungur: > I assume a struct would map to python class of some sort?

Yeah I was thinking that on the Python side the function would return a class containing just the values the function sets.

> Re the vector functions:

Yeah numpy arrays have reference semantics. Actually all python objects, except for primitive types like integers and floats, have reference semantics.

jhp-lanl: And I think there's a way to use the standard [python array](https://docs.python.org/3/library/array.html) and simply convert to a numpy array if numpy is available.

Yurlungur: Oh I wasn't even aware non-numpy python arrays existed.... Been living in numpy too long. Yes I'm sure you're right.

rbberger: As a side note, if you change the `get_sound_speed_press.py` example and use python's `array` instead of `numpy.array`, the pybind11 interface still works without changes.

rbberger: @Yurlungur added a first version of the vector overloads to the Python interface. These functions will manipulate the output arrays directly. The example now showcases their usage. So far I'm using the preprocessor to avoid redundant code. Couldn't think of a good way to deal with template member function pointers of the two overloads I'm using.

rbberger: > I assume a struct would map to python class of some sort?
> 
> If I were doing this in pure python, I might create a class called `EosState` that would instantiate with uninitialized values. I could set a couple of the values and then pass the whole thing to `FillEos` and detect the uninitialized values to be outputs.

With the same logic, we could define `FillEos` to use `kwargs` and treat not defined parameters as output and return a `py::dict` with the output values.

Yurlungur: > dded a first version of the vector overloads to the Python interface. These functions will manipulate the output arrays directly. The example now showcases their usage. So far I'm using the preprocessor to avoid redundant code. Couldn't think of a good way to deal with template member function pointers of the two overloads I'm using.

Cool! It strikes me a bit odd to provide vector overloads explicitly, given that numpy arrays work out of the box, but it's probably necessary to get good performance out of, e.g., the eospac backend.

> With the same logic, we could define FillEos to use kwargs and treat not defined parameters as output and return a py::dict with the output values.

This seems like a nice solution

jhp-lanl: Yeah that seems fine to me!

rbberger: Added `FillEos` for single (returns `EOSState`) and vectorized (manipulates arrays directly). Code will eventually need some clean up, but right now I'm still in the proof-of-concept phase.

rbberger:
> Should we add some tests for the Python infrastructure via, e.g., Nose?

* Added some basic CMake/CI infrastructure with vanilla Python `unittests`. Didn't want to add another dependency to `nose` or `pyunit` unless we actually need the more advanced features.
* Added both GH action and Gitlab CI changes: see https://github.com/lanl/singularity-eos/runs/6632563109?check_suite_focus=true

Yurlungur:
> > Should we add some tests for the Python infrastructure via, e.g., Nose?
> 
>     * Added some basic CMake/CI infrastructure with vanilla Python `unittests`. Didn't want to add another dependency to `nose` or `pyunit` unless we actually need the more advanced features.
> 
>     * Added both GH action and Gitlab CI changes: see https://github.com/lanl/singularity-eos/runs/6632563109?check_suite_focus=true

Sounds good

rbberger:
@Yurlungur Added basically everything I could see from the public interfaces now, including Spiner and EOSPAC. What's really missing the modifiers and proper tests.

rbberger:
@Yurlungur From my point of view this first iteration is ready. I've added more tests and docs in the past days. Note that I've added some autodoc documentation in Sphinx for the Python bindings, which requires the Python module to be compiled and available while building the documentation.

test/python_bindings.py
```diff
@@ -0,0 +1,435 @@
+import math
+import unittest
+import singularity_eos
+import numpy as np
+from numpy.testing import assert_allclose
+
+class EOSTestBase(object):
+    def assertIsClose(self, a, b, eps=5e-2):
+        rel = abs(b - a) / (abs(a + b) + 1e-20)
```
Yurlungur, line 9: This is fine for now, but we may want to convince ourselves at some later date that 1e-20 is the right number here.

test/python_bindings.py
```diff
@@ -0,0 +1,435 @@
...
+        self.eos.TemperatureFromDensityInternalEnergy(self.density, self.energy, self.temperature, self.num)
+        assert_allclose(self.temperature, self.temperature_true, rtol=1e-12)
+
+    def test_gamma(self):
+        """[Vector EOS][GruneisenEOS][Densities and Energies] A Gamma(rho, e) lookup is performed"""
+        self.eos.GruneisenParamFromDensityInternalEnergy(self.density, self.energy, self.gruneisen, self.num)
+        assert_allclose(self.gruneisen, self.gruneisen_true, rtol=1e-12)
+
+
+class AluminumGruneisenEOS_SoundSpeedAndPressureComp(unittest.TestCase, EOSTestBase):
```
Yurlungur, line 357: The test coverage is great. I am concerned though that we basically have the same test code duplicated in python and C++.

I wonder if we could maintain the same level of coverage by removing some of this code in the C++ unit tests since the python version tests the same underlying code path.

What do you think? That can be a project for another time of course. We can merge this in first.

Yurlungur (approved):
Overall this is great! And I'm really excited to have this in the code. A few very minor comments below, but they are non blocking.

@brryan I'm going to merge this in in a bit. I think you may find it useful from a Phoebus perspective.

Yurlungur:
@jhp-lanl @mauneyc-LANL @dholladay00 sot you want to take a look before I merge or should I just click the button?

test/python_bindings.py
rbberger: One of the downsides of the Python variants is the missing GPU versions. Other than that, they're much easier to write. So probably having them in first won't hurt, but yes long term it would be nice to remove redundancies.

test/python_bindings.py
rbberger: I originally tried to use some standard comparison functions, but due to failures then reverted to the same what the C++ tests were doing.

dholladay00:
> @jhp-lanl @mauneyc-LANL @dholladay00 sot you want to take a look before I merge or should I just click the button?

I see some tests in there and this isn't heavily used yet, so I'm good. If y'all are happy with the testing thus far I say merge.

CMakeLists.txt
```diff
@@ -354,12 +361,21 @@ if (SINGULARITY_USE_FORTRAN)
...
   PUBLIC
     $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>
     $<INSTALL_INTERFACE:include>
 )
 
+
+#------------------------------------------------------------------------------#
+# Python bindings
+#------------------------------------------------------------------------------#
+if (SINGULARITY_BUILD_PYTHON)
```
mauneyc-LANL, line 375: I'm don't want to insist on this, but with the `CMake` updates we do this as

```cmake
if(SINGULARITY_USE_HDF5)
  if(SINGULARITY_BUILD_STELLARCOLLAPSE2SPINER)
    list(APPEND _subdirs2add stellarcollapse2spiner)
  endif()
  if(SINGULARITY_USE_EOSPAC AND SINGULARITY_BUILD_SESAME2SPINER)
    list(APPEND _subdirs2add sesame2spiner)
  endif()
  if(SINGULARITY_BUILD_EXAMPLES)
    list(APPEND _subdirs2add example)
  endif()
endif()
# ... later on
foreach(_sdir ${_subdirs2add})
  singularity_msg(STATUS "Adding ${_sdir}...")
  add_subdirectory(${_sdir})
endforeach()
```

python/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
...
+  )
+  FetchContent_GetProperties(pybind11)
+  if(NOT pybind11_POPULATED)
+    FetchContent_Populate(pybind11)
+    add_subdirectory(${pybind11_SOURCE_DIR} ${pybind11_BINARY_DIR})
+  endif()
+endif()
+
+pybind11_add_module(singularity_eos module.cpp)
+target_link_libraries(singularity_eos PRIVATE singularity-eos)
```
mauneyc-LANL, line 19: What does this line do?

mauneyc-LANL (approved):

Issue #125, opened 2022-05-25
`#error "Eigen should not be used with Kokkos"`


mauneyc-LANL:
The following setup seems common enough:

```cmake
set(SINGULARITY_USE_KOKKOS ON)
set(SINGULARITY_USE_KOKKOSKERNELS OFF)
```

Without `kokkoskernels`, the library defaults to using `Eigen` as the LA backend. However, in `singularity-eos/closure/mixed_cell_models.hpp:123` there is

```c
#error "Eigen should not be used with Kokkos"
```

My understanding is that Eigen + Kokkos is okay, but Eigen + Kokkos + CUDA is what should be avoided; if so, then the check should be for Eigen + CUDA, without reference to kokkos. In any event, we should clear up the configuration of when Eigen is used, and when it is not used and do this in the CMake configuration. The preprocessor line is then redundant.

mauneyc-LANL:
This looks to be removed in `main`, closing.

mauneyc-LANL:
Addressed build failures in !126, but not clear to me if this has been resolved.

Yurlungur:
:+1: Let's close the issue for now then and re-open if needed.

PR #126, opened 2022-05-25
error to warning when using Kokkos and Eigen

mauneyc-LANL:
This changes the preprocessor directive when compiling with `kokkos` + `Eigen` to a `#warning`, so that the code will not halt compilation in this case.

This is a temporary fix for #125  . Ideally this (and other) directives will be superseded by the build configuration stage.


dholladay00:
I like this. Kokkos and Eigen aren't incompatible, only if enabling the gpu backend of kokkos is kokkos-kernels required and Eigen a problem. It also enables to test both the kokkos and non-kokkos portability strategies w/o the need to pull in kokkos kernels.

Yurlungur (approved):
I approve. Thanks for the change. And agreed it should be fixed in cmake in the long term.

PR #130, opened 2022-05-31
Add warning when provided rho_max is greater than the computed rho_max

jhp-lanl:
## PR Summary
See title

## PR Checklist

### DIFF
--- a/singularity-eos/eos/eos.hpp
+++ b/singularity-eos/eos/eos.hpp
@@ -134,10 +134,10 @@ class Gruneisen : public EosBase<Gruneisen> {
 #ifndef NDEBUG
     const Real computed_rho_max = ComputeRhoMax(s1, s2, s3, rho0);
     if (rho_max > RHOMAX_SAFETY * computed_rho_max) {
-      std::cout << "WARNING: Provided rho_max, " << rho_max << ", is greater than the computed "
-                << "rho_max of " << computed_rho_max << ". States beyond a density of "
-                << computed_rho_max << " are unphysical (i.e. imaginary sound speeds)."
-                << std::endl;
+      printf("WARNING: Provided rho_max, %g, is greater than the computed rho_max of %g.\n", rho_max,
+        computed_rho_max);
+      printf("         States beyond %g g/cm^3 are unphysical (i.e. imaginary sound speeds).\n",
+        computed_rho_max);
     }
 #endif
   }
--- a/singularity-eos/eos/eos.hpp
+++ b/singularity-eos/eos/eos.hpp
@@ -129,7 +129,18 @@ class Gruneisen : public EosBase<Gruneisen> {
             const Real b, const Real rho0, const Real T0, const Real P0, const Real Cv,
             const Real rho_max)
       : _C0(C0), _s1(s1), _s2(s2), _s3(s3), _G0(G0), _b(b), _rho0(rho0), _T0(T0), _P0(P0),
-        _Cv(Cv), _rho_max(rho_max) {}
+        _Cv(Cv), _rho_max(rho_max) {
+    // Warn user when provided rho_max is greater than the computed rho_max
+#ifndef NDEBUG
+    const Real computed_rho_max = ComputeRhoMax(s1, s2, s3, rho0);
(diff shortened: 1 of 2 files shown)


singularity-eos/eos/eos.hpp
```diff
@@ -129,7 +129,19 @@ class Gruneisen : public EosBase<Gruneisen> {
             const Real b, const Real rho0, const Real T0, const Real P0, const Real Cv,
             const Real rho_max)
       : _C0(C0), _s1(s1), _s2(s2), _s3(s3), _G0(G0), _b(b), _rho0(rho0), _T0(T0), _P0(P0),
-        _Cv(Cv), _rho_max(rho_max) {}
+        _Cv(Cv), _rho_max(rho_max) {
+    // Warn user when provided rho_max is greater than the computed rho_max
+#ifndef NDEBUG
+    const Real computed_rho_max = ComputeRhoMax(s1, s2, s3, rho0);
+    if (rho_max > RHOMAX_SAFETY * computed_rho_max) {
+      std::cout << "WARNING: Provided rho_max, " << rho_max
```
chadmeyer, line 137: This means that it can't be a `__device__` constructor -- need to use printf.

singularity-eos/eos/eos.hpp
```diff
@@ -129,7 +129,19 @@ class Gruneisen : public EosBase<Gruneisen> {
             const Real b, const Real rho0, const Real T0, const Real P0, const Real Cv,
             const Real rho_max)
       : _C0(C0), _s1(s1), _s2(s2), _s3(s3), _G0(G0), _b(b), _rho0(rho0), _T0(T0), _P0(P0),
-        _Cv(Cv), _rho_max(rho_max) {}
+        _Cv(Cv), _rho_max(rho_max) {
+    // Warn user when provided rho_max is greater than the computed rho_max
+#ifndef NDEBUG
+    const Real computed_rho_max = ComputeRhoMax(s1, s2, s3, rho0);
+    if (rho_max > RHOMAX_SAFETY * computed_rho_max) {
```
Yurlungur, line 136: Maybe worth also checking if `rho_max` is a NaN as `>` will return false in that case.

Yurlungur (approved):

singularity-eos/eos/eos.hpp
chadmeyer: There are no shortage of checks we could make.  Ensuring that `rho_max > rho0` seems promising.  But, there is a difference from a check that will ensure sane input (like this) or one that will ensure we don't sample regions that are mathematically unsound (`rho_max < computed_rho_max`).

singularity-eos/eos/eos.hpp
chadmeyer: I don't think we should assume construction on host, with the possible exception of tablular EOSs.  In case we want to do that, we should not label the constructor as a `PORTABLE_INLINE_FUNCTION`.  It could very much make sense to just call the constructor on the device for some applications.

Yurlungur: I agree but I'm not sure everything is marked appropriately. That may be something to look at later.

singularity-eos/eos/eos.hpp
jhp-lanl: I would propose that we tackle the issue of sanitizing inputs in a separate MR since that applies to other EOS as well. I don't think the behavior of the Gruneisen EOS will be particularly worse than any other EOS for NaN input parameters given that we can just expect undefined behavior across the board.

I'll make an issue.

jhp-lanl: https://github.com/lanl/singularity-eos/issues/131

chadmeyer (approved):
This looks good now, thanks @jhp-lanl for fixing this.

Yurlungur:
Seconded. Let's merge when tests pass.

jhp-lanl:
> Seconded. Let's merge when tests pass.

Correct me if I'm wrong, but I don't think there's an option to queue up the merge contingent upon the CI passing like there is in gitlab, right? That's the only reason I didn't merge this a while ago (and forgot it hadn't been merged until now).

Yurlungur:
> I think there might be a way but I don't know it.

PR #142, opened 2022-06-17
Change Gruneisen test to use a rational root instead of an irrational root

jhp-lanl:
## PR Summary

Changes the input parameters for one of the Gruneisen tests so that it uses an exact root rather than an irrational root whose solution is itself subject to root finding tolerances. The combination of these tolerances and those of our own root finder seems to be on the very edge of the test tolerance, causing it to fail on Power9 without debug flags.

## PR Checklist


jhp-lanl:
@mauneyc-LANL this may help your issues since I'm now a) using an exact root and b) using one that should propagate less error to the rho0 calculation

Yurlungur (approved):
Approved, assuming it passes on re-git.

PR #143, opened 2022-06-17
Make Gruneisen EOS root finding tolerance a function of machine epsilon

jhp-lanl:
## PR Summary

Changes the root finding tolerances used in the Gruneisen EOS to be related to machine epsilon. The function where this is used computes the maximum allowable density for the EOS given a fit to the Hugoniot.

I'm using a smaller tolerance for y than x in this case because it is possible that a local extremum will be the root (or will be very close to the root), causing the dy/dx derivative at the root to be very small. Previously the tolerances had been equal, in which case the small derivative would cause a small y tolerance to become a much larger (perhaps unacceptably so) tolerance in x.

## PR Checklist


Yurlungur (approved):
Nice improvement. Thanks!

PR #151, opened 2022-07-07
fix to install fortran mod files

mauneyc-LANL:
## PR Summary

CMake was not properly installing fortran module files when invoking `make install`. These changes explicitly install the module files.

I'm working on a more robust solution than hand-copying the files, but otherwise this addresses the prior flaw.

## PR Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Current develop
 
+### Fixed cmake code to properly install fortran module files in include path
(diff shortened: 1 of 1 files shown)


Yurlungur (approved):
Should we add some test for the fortran install so we know it works in the future?

mauneyc-LANL:
> Should we add some test for the fortran install so we know it works in the future?

No opinion; but would probably do it in a separate update

PR #156, opened 2022-07-19
Respect known inputs, constrained P and T solvers

dholladay00:
## PR Summary
The get_sg_eos function did not properly take into account the knowns and unknowns for all possible input conditions, nor could it initialize mixed material regions with P-T, rho-T, or rho-P, inputs. There is now a different kernel for each different input condition and 2 new PTE solvers that can take into account when P or T are known but the internal energy is not known. 

## PR Checklist


dholladay00:
@jhp-lanl, please take a look at the new kernels and make sure the inputs are treated as inputs and the outputs are treated as outputs.

@Yurlungur, @jdolence, @chadmeyer, @jhp-lanl I gave the solvers my best shot. For now I have just commented out initbase and replaced them with different init bases that I think should work. I think there is still an issue in initialization at how to get good initial volume fraction / component density guesses yet still conserve mass.

singularity-eos/closure/mixed_cell_models.hpp
```diff
@@ -231,6 +231,9 @@ class PTESolverBase {
     bool rho_fail;
     for (int i = 0; i < 3; i++) {
       Real vsum = 0.0;
+//      for (int m = 0; m < nmat; ++m) {
+//	printf("%i: r:%e rb:%e v:%e Tg:%e\n", m, rho[m], rhobar[m], vfrac[m], Tguess);
+//      }
```
Yurlungur, line 236: When the time comes, we should make sure to delete all debugging comments.

singularity-eos/closure/mixed_cell_models.hpp
```diff
@@ -265,6 +268,7 @@ class PTESolverBase {
     for (int m = 0; m < nmat; m++) {
       // scaled initial guess for temperature is just 1
       temp[m] = 1.0;
+      //printf("r[%i] = %e ; T = %e\n", m, rho[m], Tguess);
```
Yurlungur, line 271: see above

singularity-eos/closure/mixed_cell_models.hpp
```diff
@@ -647,6 +651,491 @@ class PTESolverRhoT : public mix_impl::PTESolverBase<EOSIndexer, RealIndexer> {
...
+         + 2 * nmat                // rhobar and u in base
+         + 2 * nmat                // nmat sized arrays in fixed T solver
+         + MAX_NUM_LAMBDAS * nmat; // the cache
+}
+inline size_t PTESolverFixedTRequiredScratchInBytes(const int nmat) {
+  return PTESolverFixedTRequiredScratch(nmat) * sizeof(Real);
+}
+
+template <typename EOSIndexer, typename RealIndexer, typename LambdaIndexer>
+class PTESolverFixedT : public mix_impl::PTESolverBase<EOSIndexer, RealIndexer> {
```
Yurlungur, line 668: I'm torn on the name `PTESolverFixedT`. But I can't think of a better name.

singularity-eos/closure/mixed_cell_models.hpp
```diff
@@ -647,6 +651,491 @@ class PTESolverRhoT : public mix_impl::PTESolverBase<EOSIndexer, RealIndexer> {
...
+                                                         press, scratch) {
+    dpdv = AssignIncrement(scratch, nmat);
+    vtemp = AssignIncrement(scratch, nmat);
+    Tequil = T_true;
+    Ttemp = T_true;
+    Tnorm = 1.0;
+    // TODO(JCD): use whatever lambdas are passed in
+    /*for (int m = 0; m < nmat; m++) {
+      if (lambda[m] != nullptr) Cache[m] = lambda[m];
+    }*/
```
Yurlungur, line 709: We should probably make an issue about this.

singularity-eos/closure/mixed_cell_models.hpp
```diff
@@ -647,6 +651,491 @@ class PTESolverRhoT : public mix_impl::PTESolverBase<EOSIndexer, RealIndexer> {
...
+  using mix_impl::PTESolverBase<EOSIndexer, RealIndexer>::residual;
+  using mix_impl::PTESolverBase<EOSIndexer, RealIndexer>::dx;
+  using mix_impl::PTESolverBase<EOSIndexer, RealIndexer>::u;
+  using mix_impl::PTESolverBase<EOSIndexer, RealIndexer>::rhobar;
+  using mix_impl::PTESolverBase<EOSIndexer, RealIndexer>::Cache;
+  using mix_impl::PTESolverBase<EOSIndexer, RealIndexer>::Tnorm;
+
+ public:
+  // template the ctor to get type deduction/universal references prior to c++17
+  template <typename EOS_t, typename Real_t, typename Lambda_t>
```
Yurlungur, line 693: Maybe this is also in the stuff Josh wrote, but why can't you use the `EOSIndexer` and `RealIndexer`s that the class is templated on?

singularity-eos/closure/mixed_cell_models.hpp
```diff
@@ -647,6 +651,491 @@ class PTESolverRhoT : public mix_impl::PTESolverBase<EOSIndexer, RealIndexer> {
...
+      jacobian[ind] = -dpdv[m];
+      jacobian[ind + 1] = dpdv[m + 1];
+    }
+//    printf("Jacobian:\n");
+//    for (int row = 0; row < nmat; row++) {
+//      for (int col = 0; col < nmat; ++col) {
+//	printf("%.3e ", jacobian[row*nmat+col]);
+//      }
+//      printf("| %.3e\n", residual[row]);
+//    }
```
Yurlungur, line 807: See above.

singularity-eos/closure/mixed_cell_models.hpp
```diff
@@ -647,6 +651,491 @@ class PTESolverRhoT : public mix_impl::PTESolverBase<EOSIndexer, RealIndexer> {
...
+  Real Init() {
+    //InitBase();
+    // fake init base
+    // rhobar is a fixed quantity: the average density of
+    // material m averaged over the full PTE volume
+    rho_total = 0.0;
+    for (int m = 0; m < nmat; ++m) {
+      rhobar[m] = rho[m] * vfrac[m];
+      rho_total += rhobar[m];
+    }
```
Yurlungur, line 935: Is there some reason this function wasn't factored out into the base class?

singularity-eos/closure/mixed_cell_models.hpp
```diff
@@ -647,6 +651,491 @@ class PTESolverRhoT : public mix_impl::PTESolverBase<EOSIndexer, RealIndexer> {
...
+      sie[m] = eos[m].InternalEnergyFromDensityTemperature(rho[m], Tequil, Cache[m]);
+      u_total += sie[m] * rho[m];
+      // note the scaling of pressure
+      press[m] =
+	eos[m].PressureFromDensityTemperature(rho[m], Tequil, Cache[m]);
+    }
+    for (int m = 0; m < nmat; ++m) {
+      press[m] /= u_total;
+      u[m] = sie[m] * rhobar[m] / u_total;
+    }
```
Yurlungur, line 737: We may want some check in this function that `rho` is greater than `rho` at Pmin.

singularity-eos/closure/mixed_cell_models.hpp
```diff
@@ -647,6 +651,491 @@ class PTESolverRhoT : public mix_impl::PTESolverBase<EOSIndexer, RealIndexer> {
...
+
+  PORTABLE_INLINE_FUNCTION
+  Real Init() {
+    //InitBase();
+    // fake init base
+    // rhobar is a fixed quantity: the average density of
+    // material m averaged over the full PTE volume
+    rho_total = 0.0;
+    for (int m = 0; m < nmat; ++m) {
+      rhobar[m] = rho[m] * vfrac[m];
```
Yurlungur, line 720: Looks like you're sitll using `vfrac` here. Is that definitely what we want to pass in to this function?

singularity-eos/eos/singularity_eos.cpp
```diff
@@ -230,6 +231,20 @@ static const std::map<const int, const unsigned long> EAPInputToBD = {
...
+struct EOSAccessor_ {
+  PORTABLE_INLINE_FUNCTION
+  EOSAccessor_(const Kokkos::View<EOS *, Llft> &eos_v, int *mats)
+    : eos_v_(eos_v), mats_(mats) {}
+  PORTABLE_INLINE_FUNCTION
+  EOS &operator[](const int m) const { return eos_v_(mats_[m]); }
+  Kokkos::View<EOS *, Llft> eos_v_;
+  int *mats_;
+};
+#endif // PORTABILITY_STRATEGY_KOKKOS
```
Yurlungur, line 245: Ah this let's you use the mat array to index into the EOS's?

singularity-eos/eos/singularity_eos.cpp
```diff
@@ -343,327 +351,512 @@ int get_sg_eos( // sizing information
...
-        // if failed, increment return value by 1
-        // use if rather than ternary to reduce contention
-        if (!res_) {
-          res() += 1;
-        }
-        // release token
-        tokens.release(token);
-      });
+  switch(input_int) {
+  case -3:
```
Yurlungur, line 375: Why is this a hardcoded number? Can we make an enum?

singularity-eos/eos/singularity_eos.cpp
```diff
@@ -343,327 +351,512 @@ int get_sg_eos( // sizing information
...
+				      pte_solver_scratch_size);
+    portableFor("PTE::solve (rho,T) input", 0, ncell,
+		PORTABLE_LAMBDA(const int &iloop)
+    {
+      // cell offset
+      const int i{offsets_v(iloop) - 1};
+      // get "thread-id" like thing with optimization
+      // for small loops
+      const int32_t token{tokens.acquire()};
+      const int32_t tid{small_loop ? iloop : token};
```
Yurlungur, line 392: Can the tokenization stuff for Kokkos be abstracted outside the switch statement?

Yurlungur (changes requested):
Great work, @dholladay00 ! A few comments:

- Looks like you mostly copy-pasted into the new solver definitions. I think there's room to abstract some of that functionality out so there's less repeat code.
- I didn't carefully cross check the equations against @jhp-lanl 's doc, but that should be done.
- I am not 100% confident about the initial guesses you're setting up for the solvers. I think that's worth thinking carefully about.
- We need some tests of these solvers. It should be pretty easy to set up a cross check We can use one of the full `PTESolverDependsRhoT` or `PTESolverDependsRhoSie` solvers to set up a system we know has an equilibrium solution, then pass in data to `PTESolverFixed*` to see if it can recover the same equilibrium.
- Nice cleanup of the  `singularity_eos*` file but I think it could still use some refactoring to clarify flow control and reduce repeated code.

singularity-eos/closure/mixed_cell_models.hpp
Yurlungur: That's fine. They're `RealIndexer`s but. We can document which one we want. If we want to support both, we could use tag dispatch or something to specify which we're doing.

singularity-eos/eos/singularity_eos.cpp
dholladay00: this is all slated to leave singularity-eos at some point or other and is very EAP specific. It can be enum-ized, but I'm not sure what it buys you.

singularity-eos/eos/singularity_eos.cpp
dholladay00: They can't, `iloop` is the lambda variable.

singularity-eos/eos/singularity_eos.cpp
Yurlungur: It buys clarity and costs one line of code. :P

singularity-eos/eos/singularity_eos.cpp
Yurlungur: ok :+1:

singularity-eos/closure/mixed_cell_models.hpp
jhp-lanl: I thought one thing we discussed was just overloading on the number of arguments so that you can specify `rhobar` in addition to `rho` and `vfrac`.

Another thing to consider is that `rhobar` and/or a mass fraction will be `const` whereas `rho` and `vfrac` are not. Could you overload on the `const`ness of these variables?

jhp-lanl (changes requested):
I'll have to finish reviewing `singularity_eos.cpp` next week unfortunately

singularity-eos/closure/mixed_cell_models.hpp
```diff
@@ -647,6 +651,491 @@ class PTESolverRhoT : public mix_impl::PTESolverBase<EOSIndexer, RealIndexer> {
...
+      rho_total += rhobar[m];
+    }
+    // utotal is fake but should act as a reasonable scale factor
```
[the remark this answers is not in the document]
dholladay00: That's probably a better name, but it would have to be change everywhere since the base rescales by this value at the end of the solve. @jdolence how would you feel about such a name change?

singularity-eos/closure/mixed_cell_models.hpp
```diff
@@ -647,6 +651,491 @@ class PTESolverRhoT : public mix_impl::PTESolverBase<EOSIndexer, RealIndexer> {
...
+  }
+
+  PORTABLE_INLINE_FUNCTION
+  void Residual() const {
+    Real vsum = 0.0;
+    for (int m = 0; m < nmat; ++m) {
+      vsum += vfrac[m];
+    }
+    residual[0] = 1.0 - vsum;
+    // the 1 here is the scaled total internal energy density
```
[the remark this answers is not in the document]
dholladay00: I mean whatever the sum of the volume fractions are for that iteration in the solver.

singularity-eos/closure/mixed_cell_models.hpp
```diff
@@ -647,6 +651,491 @@ class PTESolverRhoT : public mix_impl::PTESolverBase<EOSIndexer, RealIndexer> {
...
+      if (eos[m].PreferredInput() ==
+          (thermalqs::density | thermalqs::specific_internal_energy)) {
+        p_pert = eos[m].PressureFromDensityInternalEnergy(rho_pert, e_pert, Cache[m]) /
+                 u_total;
+      } else if (eos[m].PreferredInput() ==
+                 (thermalqs::density | thermalqs::temperature)) {
+        p_pert =
+            eos[m].PressureFromDensityTemperature(rho_pert, Tnorm * Tequil, Cache[m]) /
+            u_total;
+      }
```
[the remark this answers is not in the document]
dholladay00: In the fixed T solver, I enforced the using T since it is known to be correct. In this case, I left the logic to use the preferred eos method for the lookup.

singularity-eos/closure/mixed_cell_models.hpp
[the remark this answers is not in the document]
jhp-lanl: I just meant that it's not internal energy in the comment

singularity-eos/closure/mixed_cell_models.hpp
Yurlungur: I thought based on our in-person discussion that the changes were more extensive than that? Regardless, if a functionality is shared betweeen two `Init` methods, seems like we should put it into the base class so both can use it?

singularity-eos/closure/mixed_cell_models.hpp
Yurlungur: > Figuring out constness of the underlying data of an indexer would not be trivial.

What I have done in the past is templated on both `ConstRealIndexer` and a `RealIndexer`. This doesn't enforce constness automatically, but if a const type is passed in, type inferrence notices that and treats it accordingly.

> I thought one thing we discussed was just overloading on the number of arguments so that you can specify rhobar in addition to rho and vfrac.

Right---that's probably the way to go, as we need to output `vfrac`. But if `rhobar` is available then that implies different initialization.

singularity-eos/closure/mixed_cell_models.hpp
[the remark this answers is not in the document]
Yurlungur: My 2c: it's best to rename it `uscale` everywhere, but with comments explaining "here it's physically meaningful and here it is not."

singularity-eos/closure/mixed_cell_models.hpp
dholladay00: Have we created a const only indexer, one that entirely disallows lvalues? We can add that they could potentially be of different types, but I'm not it buys much.

singularity-eos/closure/mixed_cell_models.hpp
dholladay00: sie_total isn't known, initbase always assumes it is known and that isn't the case in either of the new solvers. So it must necessarily differ from init base as the ordering varies significantly between the 2.

singularity-eos/closure/mixed_cell_models.hpp
Yurlungur: Here's what I mean
```C++
template<typename RealIndexer, typename ConstRealIndexer>
void Foo(RealIndexer &&outs, ConstRealIndexer &&ins, int num) {
  portableFor(num, PORTABLE_LAMBDA(int i) { outs[i] = foo(ins[i]); });
}
```
what this buys us is that the input and output arrays can be different types. i.e.,
```C++
Real *my_outs = something;
const Real *my_ins = somethingelse;
Foo(my_outs, my_ins, N);
```
will compile. On the other hand, if `Foo` only templated on `RealIndexer` but didn't allow for a `ConstRealIndexer`, the discarding of the `const` qualifier would mean type inferrence would fail.

singularity-eos/closure/mixed_cell_models.hpp
Yurlungur: but the subroutine for `rho_total` is essentially the same in both cases, yeah? Can we not factor out this routine?

singularity-eos/closure/mixed_cell_models.hpp
dholladay00: oh duh, silly me :)

singularity-eos/closure/mixed_cell_models.hpp
dholladay00: We can factor out that routine.

singularity-eos/eos/singularity_eos.cpp
```diff
@@ -343,327 +350,575 @@ int get_sg_eos( // sizing information
...
-            // vfrac true
-            int midx = eos_offsets_v(m) - 1;
-            pte_mats(tid, npte) = midx;
-            pte_idxs(tid, npte) = m;
-            mass_pte(tid, npte) = frac_mass_v(i, m);
-            sie_pte(tid, npte) = frac_sie_v(i, m) / frac_mass_v(i, m);
-            vfrac_pte(tid, npte) = frac_vol_v(i, m);
-            npte += 1;
+          // since rho_ptes are correct, now do rho-T lookups for remaining quantities
+          // first zero out quantities to be averaged
```
jhp-lanl, line 460: As I see it, you're using rho-T lookups to fill out all the quantities no matter what the input conditions were right? Can this be put in a function to avoid code duplication?

singularity-eos/eos/singularity_eos.cpp
```diff
@@ -343,327 +350,575 @@ int get_sg_eos( // sizing information
...
+          double mass_sum{0.0};
+          // normalize mass fractions
+          // first find the mass sum
+          // also set idxs as the decrement of the eos offsets
+          // to take into account 1 based indexing in fortran
+          for (int m = 0; m < nmat; ++m) {
+            mass_sum += frac_mass_v(i, m);
+            pte_idxs(tid, m) = eos_offsets_v(m) - 1;
+            frac_vol_v(i, m) = 0.0;
+          }
```
jhp-lanl, line 404: This can be a function I think since it's used many times

singularity-eos/eos/singularity_eos.cpp
```diff
@@ -343,327 +350,575 @@ int get_sg_eos( // sizing information
...
+            if (frac_mass_v(i, m) > 1.e-12) {
+              pte_idxs(tid, npte) = eos_offsets_v(m) - 1;
+              pte_mats(tid, npte) = m;
+              npte += 1;
+            }
+            vfrac_pte(tid, m) = 0.0;
+            sie_pte(tid, m) = 0.0;
+            temp_pte(tid, m) = 0.0;
+            press_pte(tid, m) = 0.0;
+          }
```
jhp-lanl, line 420: Similarly, this can be put in a function too I think

singularity-eos/eos/singularity_eos.cpp
```diff
@@ -343,327 +350,575 @@ int get_sg_eos( // sizing information
...
+
+          for (int mp = 0; mp < npte; ++mp) {
+            const int m = pte_mats(tid, mp);
+            rho_pte(tid, mp) = npte / spvol_v(i) * frac_mass_v(i, m);
+            vfrac_pte(tid, mp) = 1.0 / npte;
```
jhp-lanl, line 425: Another function

singularity-eos/eos/singularity_eos.cpp
dholladay00: Before we get too carried away with code making everything a function. There is probably a much better way of handling this entire function as a whole and we have discussed that very thing and I just don't think this is the PR to do that. First correctness, then we can think more about what the correct function is and how to go about reducing duplication.

Yurlungur: I'd rather keep things clean here---though I'm willing to do a cleanup PR after this one, assuming that cleanup actually happens.

jhp-lanl: My main motivation in bringing this up is because I was trying to read through the code to help in debugging the xRAGE integration. I was thinking that making the code more DRY would aid in the debugging process. In speaking from my own experience, I've added entirely too many bugs by not putting code in a function when I wanted to reuse it.

At this point in the debugging process, I think you've eliminated this code from being the source of the error we're tracking down, but I think it may be helpful in the future for readability and future debugging.

singularity-eos/closure/mixed_cell_models.hpp
dholladay00: I am now calculating the offset from the scratch memory and using it in the calling code. We can still create an issue if you think it necessary.

singularity-eos/closure/mixed_cell_models.hpp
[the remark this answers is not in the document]
dholladay00: done

singularity-eos/eos/singularity_eos.cpp
dholladay00: done

dholladay00:
@jhp-lanl @jdolence I just saw that new Tguess machinery was merged in. Now that I have function in the base class for obtaining a Tguess, should it be used there?

singularity-eos/eos/singularity_eos.cpp
dholladay00: different values need to be set based on the input condition.

singularity-eos/eos/singularity_eos.cpp
dholladay00: Yep that's correct.

singularity-eos/closure/mixed_cell_models.hpp
dholladay00: It is, and I tried this but it resulted in compilation errors. I assume the comment is the reason.

singularity-eos/closure/mixed_cell_models.hpp
dholladay00: a different template parameter has been added in case one passes in const real *'s

singularity-eos/closure/mixed_cell_models.hpp
dholladay00: resolving for now.

singularity-eos/closure/mixed_cell_models.hpp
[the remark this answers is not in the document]
dholladay00: done

singularity-eos/closure/mixed_cell_models.hpp
dholladay00: done.

jhp-lanl:
> @jhp-lanl @jdolence I just saw that new Tguess machinery was merged in. Now that I have function in the base class for obtaining a Tguess, should it be used there?

Are you asking about your `GetTguess()` function? I don't think that's necessary. It looks like @jdolence intended `ApproxTemperatureFromRhoMatU()` to be used by the host code when an approximation for the PTE volume fractions and densities is already known (i.e. the host code decides when this is the case).

Your stuff is mainly for initialization where we wouldn't expect the material volumes or volume fractions to be known before the PTE solve, so your volume fraction guess is probably as good as any. I think your functions need to assume that only the bare minimum is known about a state. You still allow a temperature guess to be passed through right? I think that should be sufficient.

But I haven't thought about this code in a while so I could be forgetting key details.

singularity-eos/eos/singularity_eos.cpp
dholladay00: done

singularity-eos/eos/singularity_eos.cpp
dholladay00: done

singularity-eos/eos/singularity_eos.cpp
dholladay00: done

singularity-eos/eos/singularity_eos.cpp
dholladay00: done

jhp-lanl (approved):
This looks good! I think you addressed all of my issues

dholladay00:
@Yurlungur I think this is ready, give this another look when you get a chance. Hoping to get this merged soon.

Yurlungur (approved):
You've addressed all my comments and this looks good. But before we merge, I asked @jdolence to run this branch in riot to make sure the changes to PTE didn't break anything.

Yurlungur:
Oh also @dholladay00 and/or @jhp-lanl if we could add documentation regarding the two new PTE solvers, that would be helpful. Jeff's pdf should be enough, just dump it into sphinx in the right file.

singularity-eos/closure/mixed_cell_models.hpp
```diff
@@ -711,6 +710,422 @@ class PTESolverRhoT : public mix_impl::PTESolverBase<EOSIndexer, RealIndexer> {
...
+
+  PORTABLE_INLINE_FUNCTION
+  Real Init() {
+    // InitBase();
+    // fake init base
+    // rhobar is a fixed quantity: the average density of
+    // material m averaged over the full PTE volume
+    Tnorm = 1.0;
+    this->InitRhoBarandRho();
+    this->SetVfracFromT(Tequil);
```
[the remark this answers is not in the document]
dholladay00: if you use `this->`, you don't have to add the using statement in the class declaration

singularity-eos/eos/singularity_eos.cpp
```diff
@@ -329,12 +350,6 @@ int get_sg_eos( // sizing information
   const auto eos_nmat{*std::max_element(eos_offsets, eos_offsets + nmat)};
   Kokkos::View<EOS *, Llft, HS, Unmgd> eos_hv(eos, eos_nmat);
   Kokkos::View<EOS *, Llft> eos_v{create_mirror_view_and_copy(DMS(), eos_hv)};
-  double **lambda_map = (double **)PORTABLE_MALLOC(sizeof(double *) * nmat);
-  portableFor(
-      "filling lambda ptrs", 0, nmat, PORTABLE_LAMBDA(const int &i) {
-        // const int lambda_size = eos_v(i).nlambda();
-        lambda_map[i] = nullptr; //(double*)malloc(sizeof(double)*lambda_size);
-      });
```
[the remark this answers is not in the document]
dholladay00: The cache is now allocated as part of the scratch space.

dholladay00:
@Yurlungur, in terms of using the cache, I do the offset calculation to figure out the cache location within the solver and create an accessor to it outside of the solve. It's not ideal but it doesn't mess with the current implementation of caching in the solvers and it allows me to take advantage of it when doing lookups after the PTE solve.

jhp-lanl:
> Oh also @dholladay00 and/or @jhp-lanl if we could add documentation regarding the two new PTE solvers, that would be helpful. Jeff's pdf should be enough, just dump it into sphinx in the right file.

I can do this @dholladay00 since you've been slogging through debugging and writing this code. It will come in a separate MR and I'll create an issue

singularity-eos/closure/mixed_cell_models.hpp
dholladay00: The PTE solver uses a portion of the solver scratch memory to store an internal cache. The address of the first location of the cache, as well as the cache size is known and can be calculated anywhere. I perform this calculation in the scope outside the PTE solve and thus have access to the cache used internally by the PTE solver. Therefore I do not need to allocate additional memory for the cache nor do I need to pass it in to the PTE solve, and even if I did, it would not be used as can be seen in the implementation. If you need more context it may be easier to communicate over the phone.

test/test_pte.cpp
```diff
@@ -88,15 +89,131 @@ void set_state(RealIndexer &&rho, RealIndexer &&vfrac, RealIndexer &&sie,
     vfrac[i] *= 1.0 / vsum;
 }
 
+#ifdef PORTABILITY_STRATEGY_KOKKOS
+int run_sg_get_eos_tests() {
```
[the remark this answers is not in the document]
dholladay00: I have added a todo.

singularity-eos/eos/singularity_eos.cpp
```diff
@@ -259,6 +274,13 @@ int get_sg_eos( // sizing information
   bool do_frac_cv{true};
   if (frac_dpde == NULL || frac_dpde == nullptr) do_frac_cv = false;
   // get inputs
+  enum input_condition {
```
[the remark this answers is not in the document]
dholladay00: changed to an enum class.

singularity-eos/eos/singularity_eos.cpp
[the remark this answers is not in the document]
dholladay00: see comments above.

singularity-eos/closure/mixed_cell_models.hpp
Yurlungur: Nah I get it now. Allocating the lambdas in scratch memory is nice, but we should still make an issue because in the future, there could be lambda parameters that are not cache parameters but some actual physical values. (For example a mix fraction for a hybrid or multiphase eos.)

singularity-eos/eos/singularity_eos.cpp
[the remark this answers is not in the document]
Yurlungur: :+1:

singularity-eos/closure/mixed_cell_models.hpp
dholladay00: I agree, but I think it should be part of a different MR. If I made an issue would that suffice for now?

Yurlungur: An issue suffices for now. We don't currently have a use-case to test against---let's not waste developer effort. I just want to make sure we keep this in mind going forward.

dholladay00: addressed in issue #173

Yurlungur:
> @Yurlungur, in terms of using the cache, I do the offset calculation to figure out the cache location within the solver and create an accessor to it outside of the solve. It's not ideal but it doesn't mess with the current implementation of caching in the solvers and it allows me to take advantage of it when doing lookups after the PTE solve.

:+1: I think this good enough for now. With the issue we can resolve later.

> I can do this @dholladay00 since you've been slogging through debugging and writing this code. It will come in a separate MR and I'll create an issue

Thanks, @jhp-lanl ! That works for me.

As soon as @jdolence confirms nothing breaks downstream in riot, I'm ready to merge.

dholladay00:
@Yurlungur @jdolence is there a timeline for the downstream testing? I'd like to get this merged ASAP.

dholladay00:
@Yurlungur and @jdolence I reiterate that this should be merged ASAP.

Yurlungur:
Just spoke to @jdolence let's just merge it. Wil lmerge as soon as tests pass.

PR #157, opened 2022-07-28
Replace secant method with regula_falsi

chadmeyer:
## PR Summary
Replaces the original hand-written solvers using a naive secant method (with a bug!) with the now present `regula_falsi` solver. 

The buggy versions were in the Gruneisen EOS and the two Davis EOSs, and there was a slightly less buggy version (which had the missing abs) in JWL, but it makes sense for these to be put together for consistency.

## PR Checklist

### DIFF
--- a/singularity-eos/eos/eos_davis.cpp
+++ b/singularity-eos/eos/eos_davis.cpp
@@ -14,6 +14,7 @@
 
 #include <cmath>
 #include <singularity-eos/eos/eos.hpp>
+#include <singularity-eos/base/root-finding-1d/root_finding.hpp>
 
 namespace singularity {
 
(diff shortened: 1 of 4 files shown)


Yurlungur (approved):
:+1: Great improvement. I ran clang-format and triggered the CI on re-git. Once that passes, let's get this in.

Yurlungur:
At some point we should add tests to JWL/Davis, but that can be for another day.

singularity-eos/eos/eos_davis.cpp
```diff
@@ -139,28 +140,20 @@ PORTABLE_FUNCTION void DavisReactants::DensityEnergyFromPressureTemperature(
...
-    if (resm / press < 1e-8) break;
-    rho1 = rho2;
-    res1 = res2;
-    rho2 = rhom;
-    res2 = resm;
+  using RootFinding1D::regula_falsi;
+  using RootFinding1D::Status;
+  RootFinding1D::RootCounts counts;
+  auto status =
+      regula_falsi(PofRatT, press, _rho0, 1.0e-5, 1.0e3, 1.0e-8, 1.0e-8, rho, counts);
```
dholladay00, line 151: Is there a need to have better bounding guesses than hardcoded ones? Should we base them off of rho0?

Yurlungur: This is a good point. We could make an issue for that and address it later, though, when we're actually working more closely with davis.

chadmeyer: Possibly.  I know that the Regula Falsi solver gave me some trouble with wider bounds on a Gruneisen EOS (1e-12 -- 1e12) however, with an initial guess of `_rho0` across the board, we're likely to find the solution we desire.  There is logic (that pre-existed this MR) in the JWL version that checks to see if the incoming rho is set (which itself isn't a guarantee of anything good either).

I'll re-echo my previous comments that this sort of routine is likely limited to initialization and therefore initial-like states.  If that turns out to be insufficient, we could re-visit.

Yurlungur: Issue #158. Let's put a pin in it and re-visit when we're working more closely with this EOS.

PR #164, opened 2022-08-15
Provide a function to give initial temperature guesses for PTE

jdolence:
## PR Summary
This PR introduces a function that solves $u_0 = \sum_m \rho_m \alpha_m \epsilon_m(\rho_m, T)$ for the temperature $T$, for fixed input $u_0$ (total material energy density, i.e. $erg/cm^{3}$), $\alpha_m$ (material volume fractions), and $\rho_m$ (material physical densities).  The intent of the function is to be used to provide an initial guess for the PTE solver, so the hard coded tolerance of the solver is quite loose.  Bisection is used because it's better behaved for EOSs that have only a weak dependence of $\epsilon$ on $T$ at cold temperatures.

To go along with this, the PTE solvers have been augmented to allow an initial temperature guess to be (optionally) provided.  If a guess is not provided, or the guess is zero, the solver tries to guess a temperature as before.

## PR Checklist

### DIFF
--- a/singularity-eos/closure/mixed_cell_models.hpp
+++ b/singularity-eos/closure/mixed_cell_models.hpp
@@ -49,6 +49,8 @@ constexpr Real line_search_alpha = 1.e-2;
 constexpr int line_search_max_iter = 6;
 constexpr Real line_search_fac = 0.5;
 constexpr Real vfrac_safety_fac = 0.95;
+constexpr Real minimum_temperature = 1.e-9;
+constexpr Real maximum_temperature = 1.e9;
 } // namespace mix_params
 
 namespace mix_impl {
--- a/singularity-eos/closure/mixed_cell_models.hpp
+++ b/singularity-eos/closure/mixed_cell_models.hpp
@@ -16,6 +16,7 @@
 #define _SINGULARITY_EOS_CLOSURE_MIXED_CELL_MODELS_
 
 #include <ports-of-call/portability.hpp>
+#include <singularity-eos/base/fast-math/logs.hpp>
 #include <singularity-eos/eos/eos.hpp>
 
 #include <cmath>
--- a/singularity-eos/closure/mixed_cell_models.hpp
+++ b/singularity-eos/closure/mixed_cell_models.hpp
@@ -420,12 +420,10 @@ class PTESolverBase {
 
 } // namespace mix_impl
 
-
 template <typename EOSIndexer, typename RealIndexer>
-PORTABLE_INLINE_FUNCTION
-Real ApproxTemperatureFromRhoMatU(const int nmat, EOSIndexer &&eos, const Real u_tot,
-                                  RealIndexer &&rho, RealIndexer &&vfrac,
-                                  const Real Tguess = 0.0) {
+PORTABLE_INLINE_FUNCTION Real ApproxTemperatureFromRhoMatU(
+    const int nmat, EOSIndexer &&eos, const Real u_tot, RealIndexer &&rho,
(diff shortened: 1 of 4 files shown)


jhp-lanl:
The changes look good to me, but can we add a dead simple unit test to cover the behavior?

singularity-eos/closure/mixed_cell_models.hpp
```diff
@@ -221,9 +224,15 @@ class PTESolverBase {
     u_total = rho_total * sie_total;
 
     // guess some non-zero temperature to start
-    Real Tguess = 300.0;
-    for (int m = 0; m < nmat; ++m)
-      Tguess = std::max(Tguess, temp[m]);
+    // If a guess was passed in, it's saved in Tnorm
+    Real Tguess;
+    if (Tnorm > 0.0) {
```
Yurlungur, line 229: Should we allow zero temperature? I assume not?

singularity-eos/closure/mixed_cell_models.hpp
```diff
@@ -412,6 +421,60 @@ class PTESolverBase {
...
+    const Real uT = ufunc(FastMath::pow2(lT));
+    if (uT < u_tot) {
+      lTlo = lT;
+      ulo = uT;
+    } else {
+      lThi = lT;
+      uhi = uT;
+    }
+    iter++;
+  }
```
Yurlungur, line 472: Don't think this is necessary, but don't we have a pre-packaged bisection in the root finders file you could use?

singularity-eos/closure/mixed_cell_models.hpp
```diff
@@ -412,6 +421,60 @@ class PTESolverBase {
...
+      Tguess < mix_params::maximum_temperature) {
+    const Real ug = ufunc(Tguess);
+    if (ug < u_tot) {
+      lTlo = FastMath::lg(Tguess);
+      ulo = ug;
+    } else {
+      lThi = FastMath::lg(Tguess);
+      uhi = ug;
+    }
+  }
```
Yurlungur, line 456: Nice use of the fast math. I assume the fast logs here don't need to be precise, so long as they're exactly invertible?

Yurlungur (approved):
I like this. Also nice use of not-quite-transcendental functions. :D 

I do think we should have some kind of test. What about just using this initial guess machinery in `test_pte.cpp` to generate initial guesses for the solver?

singularity-eos/closure/mixed_cell_models.hpp
jdolence: maybe.  just figured it was like 10 lines of code.  also, can `max_iter` be easily controlled via the function interface because that needs to be small here, but shouldn't typically be small for a bisection based root find.

singularity-eos/closure/mixed_cell_models.hpp
jdolence: correct.  doesn't matter that they're logs at all, just need to be log like and invertible.

singularity-eos/closure/mixed_cell_models.hpp
Yurlungur: yeah fair enough.

jdolence:
thanks for the quick review @jhp-lanl and @Yurlungur.  I've added a call to this function in the test_pte code, as @Yurlungur suggested.  If the tests pass, is this good to go?

Yurlungur:
Yep, I'd say so. I'll trigger tests on Darwin and then let's get it merged in.

Yurlungur:
Ah, actually Darwin CI currently in-repair. I'll just merge when github tests pass.

PR #165, opened 2022-08-16
Draft: Updated target install paths, updates to C/F90 interface

mauneyc-LANL:
The public `include` dirs made finding the fortran module a headache, so the exported `include` path now has the direct path to it. 
```
<include_install> #<- this is what `singularity-eos::singularity-eos` has for the include dir
|_ base
|_ closure
|_ eos
__|_ singularity_eos.mod #<- the compiler couldn't "find" this without forcing it to look down a directory. (CMake has some blindspots with Fortran)
```
Another possibility is to install the modfile up in the top directory. 

Also made a change to bring in the `HDF5` fortran libraries if `SINGULARITY_USE_FORTRAN` is enabled; this should make slotting into other fortran codes easier.

In tandem, a "skeleton" fortran test is now implemented. This was based on what looked to be an old fortran test. Right now it just `init` and `finalize`, without evaluating anything.

I had to update the `singularity-eos.cpp` (and there other cases of this) where code is compliation guarded with preprocessor directives

```c
#ifdef SINGULARITY_USE_EOSPAC
// declare/define function
#endif
```
However, the fortran interface doesn't have this compiler guards, so there could be undefined references.
The straightforward thing would be to clone the `#ifdef` guards into the fortran, but that opens a potential can of worms (compilers all do this slightly differently). Instead, I fixed the `EOSPAC` bit but always declaring the function (no `#ifdef` guard in the `.hpp` file), but in the definition the `#ifdef` is used to conditional compile the "real" function or a stub function.

## PR Checklist


CMakeLists.txt
```diff
@@ -275,9 +275,13 @@ if (SINGULARITY_USE_HDF5)
   # hdf5, for data interopability
   # type: external
   enable_language(C)
+  set(_hdf5_comp "C;HL")
+  if(SINGULARITY_USE_FORTRAN)
+    set(_hdf5_comp "${_hdf5_comp};Fortran")
+  endif()
```
Yurlungur, line 281: Is this needed? I don't think the fortran HDF5 functionality is ever required.

singularity-eos/eos/singularity_eos.cpp
```diff
@@ -208,9 +208,21 @@ int init_sg_eospac(const int matindex, EOS *eos, const int id, int const *const
...
+// CMM: A more robust solution may be to preprocess the fortran,
+//      however that is slightly more involved than you'd like to think.
+#else // stubs if not built, just to satisfy linking.
+int init_sg_eospac(const int matindex, EOS *eos, const int id, int const *const enabled,
+                   double *const vals) {
+  return 0;
+}
+
+int init_sg_eospac(const int matindex, EOS *eos, const int id) { return 0; }
+
```
Yurlungur, line 225: this seems fine for now

test/CMakeLists.txt
```diff
@@ -81,3 +81,9 @@ if (SINGULARITY_BUILD_PYTHON AND SINGULARITY_TEST_PYTHON)
            COMMAND ${Python_EXECUTABLE} ${CMAKE_CURRENT_SOURCE_DIR}/python_bindings.py -v)
   set_tests_properties(PythonBindings PROPERTIES ENVIRONMENT "${PYTHON_TEST_ENVIRONMENT}")
 endif()
+
+if(SINGULARITY_USE_FORTRAN)
+  add_executable(test_ftniface test_ftniface.f90)
+  target_link_libraries(test_ftniface singularity-eos::singularity-eos)
+  add_test(ftniface test_ftniface)  
+endif()
```
Yurlungur, line 89: :+1:

Yurlungur (approved):
Overall looks great, pending the one substantive comment below, regarding HDF5.

CMakeLists.txt
dholladay00: Perhaps it would be ok, but it seems that hdf5 through cmake is very fragile and subtle differences cause linking issues in host codes. When hdf5 is less fragile, perhaps we can get away with using this machinery as intended, but until then, I'm hesitant to allow this line to be different from the host codes.

mauneyc-LANL: This actually is looking maybe a bit complicated: when building for EAP, there are some undefined symbols related to `libhdf5_fortran`. It's not clear to me if this is an issue with EAP build system. SAP is using hand-spun makefiles, so this may also be an issue with CMake handling libs. 

I know @dholladay00 had to do something with the `singularity-eos` hdf5 libs to build with EAP correctly, my intinct was it was the Fortran libs. My thinking here was to do what we're doing with MPI/HDF5, though there are differences that I know think have some deeper issue.

If there isn't objection, can we bring this into `main` for now?

mauneyc-LANL: > When hdf5 is less fragile, perhaps we can get away with using this machinery as intended, but until then, I'm hesitant to allow this line to be different from the host codes.

I'm actually replicating the EAP HDF5 components here, (i.e. `find_package(HDF5 COMPONENTS C Fortran HL)`)

singularity-eos/eos/singularity_eos.cpp
dholladay00: This may need some amending for GPU builds. I'm not sure the Darwin CI will pass. Lets get that passing and then merge in. But I agree that eventually we should use the preprocessor to block out unneeded functions. @mauneyc-LANL could you make an issue so that I don't forget to modify the fortran interface to use the preprocessor in the future?

mauneyc-LANL: Just to make a note of it: this would apply to other "dependent" routines that are conditionally compiled (e.g. stuff with `SINGULARITY_USE_HDF5/SPINER_USE_HDF5`). For the moment I'm not aware of host-codes dependent on `HDF5` that need a Fortran interface, but likely will encounter a similar issue to the above if that every happens.

mauneyc-LANL: > This may need some amending for GPU builds. I'm not sure the Darwin CI will pass. Lets get that passing and then merge in. But I agree that eventually we should use the preprocessor to block out unneeded functions. @mauneyc-LANL could you make an issue so that I don't forget to modify the fortran interface to use the preprocessor in the future?

I should probably make it clear here: the reason I didn't use preprocessor guards in the Fortran source is because this compilers all handle fortran preprocessing in slightly different ways (and AFAIK none do it by default), and this can lead to frustrations down the road (especially with CMake, with it's somewhat tepid Fortran support). I would argue against putting preprocessor directives into the Fortran source, if we can help it.

As for tests, this would only effect a build with `SINGULARITY_USE_FORTRAN=ON` & `SINGULARITY_USE_EOSPAC=OFF`, and as best as I can tell the only test where that is an issue would be the test I added with this PR.

CMakeLists.txt
Yurlungur: Ok---if that's what it takes for the host codes, that's what it takes for the host codes.

singularity-eos/eos/singularity_eos.cpp
Yurlungur: Why would this effect GPU builds at all @dholladay00 ?

dholladay00: Because it will by default use fortran as the linker language and resolving relocatable device code at link time requires extra steps that the fortran linker won't know about unless to you tell cmake to tell the fortran linker to worry about it(it will require enabling cuda as one of the cmake project languages and then a special command for the exec). I now know how to make this work so it's not a big deal, but I'd be very surprised if this test compiles and links for GPU builds.

Yurlungur: @mauneyc-LANL can you give this a try?

mauneyc-LANL:
Wait for #168 to merge to main

singularity-eos/eos/singularity_eos.cpp
dholladay00: @Yurlungur @mauneyc-LANL we no longer need rdc

CMakeLists.txt
```diff
@@ -275,9 +275,13 @@ if (SINGULARITY_USE_HDF5)
   # hdf5, for data interopability
   # type: external
   enable_language(C)
+  set(_hdf5_comp "C;HL")
+  if(SINGULARITY_USE_FORTRAN)
+    set(_hdf5_comp "${_hdf5_comp};Fortran")
```
dholladay00, line 280: I don't think we require fortran hdf5 interface at all do we?

mauneyc-LANL: Correct; I thought this fixed an issue downstream but I was likely incorrect. Let me sort that out first and I'll remove this once I'm sure it's erroneous

singularity-eos/eos/singularity_eos.cpp
Yurlungur: :+1:

CMakeLists.txt
dholladay00: Ooohhhh yes this is all coming back to me. If its needed even for silly reasons we can leave it in for now. 

@Yurlungur the context is this may be needed due to hdf5/cmake combo is buggy and the fortran is needed in EAP.

Yurlungur: ok I see.

mauneyc-LANL:
Closing as I'm putting changes into #181

PR #167, opened 2022-08-16
let CMake fail to infer version of Kokkos

Yurlungur:
## PR Summary

I recently tried to compile `singularity-eos` against a Kokkos release candidate, rather than a firm version, and I discovered that the version is not appropriately populated in `cmake`. This adds guard rails for this use-case, where `cmake` warns the user that it could not correctly infer `Kokkos`'s version number, but then chugs along anyway.

@mauneyc-LANL please take a look at this. It's somewhat blocking downstream.

## PR Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Added (new features/APIs/variables/...)
 
 ### Changed (changing behavior/API/variables/...)
+- [[PR67]](https://github.com/lanl/singularity-eos/pull/167) allow for the possiblity Kokkos version can't be inferred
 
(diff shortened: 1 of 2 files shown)


mauneyc-LANL (approved):

dholladay00:
Yes I had noticed this at some point to. I think the band-aid I used to was slightly different but as long as this works then I'm ok with it.

Yurlungur:
> Yes I had noticed this at some point to. I think the band-aid I used to was slightly different but as long as this works then I'm ok with it.

I found grosser band-aids downstream. ;) But this one seems relatively robust, I think.

PR #168, opened 2022-08-22
Move EOS files to header only

Yurlungur:
## PR Summary

This is moving us towards supporting `OpenMP` target offload. It also is just more performant because it doesn't require link-time-optimization for CUDA builds.

@ktsai7 I'm pinging you because some tests on Darwin started failing late last week---without any change to the code as far as I can tell---and it would be good to have that resolved so we know these tests pass on GPU, as this is a big change.

## PR Checklist


singularity-eos/eos/eos_davis.hpp
```diff
@@ -12,18 +12,159 @@
...
+ public:
+  DavisReactants() = default;
+  PORTABLE_INLINE_FUNCTION
+  DavisReactants(const Real rho0, const Real e0, const Real P0, const Real T0,
+                 const Real A, const Real B, const Real C, const Real G0, const Real Z,
+                 const Real alpha, const Real Cv0)
+      : _rho0(rho0), _e0(e0), _P0(P0), _T0(T0), _A(A), _B(B), _C(C), _G0(G0), _Z(Z),
+        _alpha(alpha), _Cv0(Cv0) {}
+  DavisReactants GetOnDevice() { return *this; }
+  PORTABLE_INLINE_FUNCTION Real TemperatureFromDensityInternalEnergy(
```
dholladay00, line 39: Maybe this is work for a different PR, but I think collocating declarations and implementations would be preferable.

singularity-eos/eos/eos_davis.hpp
```diff
@@ -12,18 +12,159 @@
...
+  static constexpr unsigned long _preferred_input =
+      thermalqs::density | thermalqs::specific_internal_energy;
+  PORTABLE_INLINE_FUNCTION Real Ps(const Real rho) const;
+  PORTABLE_INLINE_FUNCTION Real Es(const Real rho) const;
+  PORTABLE_INLINE_FUNCTION Real Ts(const Real rho) const;
+  PORTABLE_INLINE_FUNCTION Real Gamma(const Real rho) const;
+  PORTABLE_INLINE_FUNCTION Real square(const Real x) const { return x * x; }
+  PORTABLE_INLINE_FUNCTION Real cube(const Real x) const { return x * x * x; }
+  PORTABLE_INLINE_FUNCTION Real fourth(const Real x) const { return x * x * x * x; }
+  PORTABLE_INLINE_FUNCTION Real fifth(const Real x) const { return x * x * x * x * x; }
```
dholladay00, line 97: These should be `FORCEINLINE` and they are common to both Davis implementations, can we move them to be free functions or something like that to reduce duplication?

singularity-eos/eos/eos_davis.hpp
Yurlungur: Yeah my preference is short declarations/implementations should be collocated, and longer ones separate. I did that for `IdealGas` and then quickly decided it was too tedious to do for all of them. I say we make that change piecewise as we work on individual EOS's.

singularity-eos/eos/eos_davis.hpp
```diff
@@ -186,7 +327,7 @@ void DavisReactants::ValuesAtReferenceState(Real &rho, Real &temp, Real &sie, Re
 }
 
 #if 0
```
dholladay00, line 329: This section of code is not compiled, should we remove it?

singularity-eos/eos/eos_davis.hpp
Yurlungur: I'll add the `FORCRINLINE`. 

Combining the implementations is a bit annoying---I moved them into the classes to avoid exposing them to global namespace. I can make a davis_utils` or `davis_impl` namespace to hide them again, it just involves peppering the code with `using namespace davis_utils` everywhere in all the functions. When I made the changes, I felt duplicating the functionality in both classes was the simpler option. I'm willing to be convinced otherwise.

Yurlungur: I created an issue: #170

singularity-eos/eos/eos_davis.hpp
Yurlungur: I created an issue: #170

singularity-eos/eos/eos_spiner.hpp
```diff
@@ -26,19 +28,418 @@
...
+
+  static constexpr unsigned long PreferredInput() { return _preferred_input; }
+  int matid() const { return matid_; }
+  PORTABLE_INLINE_FUNCTION Real lRhoOffset() const { return lRhoOffset_; }
+  PORTABLE_INLINE_FUNCTION Real lTOffset() const { return lTOffset_; }
+  PORTABLE_INLINE_FUNCTION Real lEOffset() const { return lEOffset_; }
+  PORTABLE_INLINE_FUNCTION Real rhoMin() const { return fromLog_(lRhoMin_, lRhoOffset_); }
+  PORTABLE_INLINE_FUNCTION Real rhoMax() const { return fromLog_(lRhoMax_, lRhoOffset_); }
+  PORTABLE_INLINE_FUNCTION Real TMin() const {
+    return fromLog_(T_.range(0).min(), lTOffset_);
```
dholladay00, line 366: could these be initialized and stored at construction? I'm not sure how often these calls are made. I know this is unrelated to this PR, so maybe make an issue?

singularity-eos/eos/eos_stellar_collapse.hpp
```diff
@@ -25,18 +27,306 @@
...
+  inline void medianFilter_(Spiner::DataBox &db);
+  inline void medianFilter_(const Spiner::DataBox &in, Spiner::DataBox &out);
+  inline void fillMedianBuffer_(Real buffer[], int width, int iY, int iT, int irho,
+                                const Spiner::DataBox &tab) const;
+  inline Real findMedian_(Real buffer[], int size) const;
+  inline void computeBulkModulus_();
+  inline void computeColdAndHotCurves_();
+  inline void setNormalValues_();
+
+  PORTABLE_INLINE_FUNCTION __attribute__((always_inline)) void
```
dholladay00, line 188: `FORCEINLINE`?

singularity-eos/eos/eos_spiner.hpp
Yurlungur: yes they can and this is an easy fix. I can do it here.

singularity-eos/eos/eos_davis.hpp
chadmeyer: I mean, these aren't *davis* utilities; they are generic, it's just that nobody else needs them.

singularity-eos/eos/eos_davis.hpp
chadmeyer: As discussed, if we don't need this anymore for the PTE solver formulation (we don't) then it can go.

singularity-eos/eos/eos_davis.hpp
dholladay00: @chadmeyer, these functions are reimplemented in starting in line 163

singularity-eos/eos/eos_gruneisen.hpp
```diff
@@ -12,14 +12,116 @@
...
+  Real dPres_drho_e(const Real rho, const Real sie) const;
+  static constexpr const unsigned long _preferred_input =
+      thermalqs::density | thermalqs::specific_internal_energy;
+  // Scaling factor for density singularity (reference pressure blows up). Consistent with
+  // Pagosa implementation
+  static constexpr Real RHOMAX_SAFETY = 0.99;
+};
+
+namespace gruneisen_utils {
 PORTABLE_INLINE_FUNCTION Real square(const Real x) { return x * x; }
```
dholladay00, line 125: @Yurlungur, @chadmeyer, here are more duplicates.

singularity-eos/eos/eos_davis.hpp
dholladay00: I also think these should be implemented force inline and that they build on each other (cube = square*x, fourth=square*square, fifth = square*cube). This will produce faster code at all opt levels.

mauneyc-LANL: This optimizes quite well

```c++
template<unsigned N>
constexpr auto sg_pow(const Real value)
{
  return value * sg_pow<N-1>(value);
}

template<>
constexpr auto sg_pow<0>(const Real value)
{
  return 1;
}
```

jhp-lanl: I always enjoy a _good_ recursive function 😄

dholladay00: I am looking at it in compiler explorer and it does not appear to optimize well, even at O3, only with O3 -ffast-math (or intel equivalent) will it optimize a fourth power to 2 multiplies. I wouldn't mind have a mode where we enable fast math, but I know it's suspect from a reproducibility standpoint.

mauneyc-LANL: > appear to optimize well, even at O3, only with O3 -ffast-math (or intel equivalent) will it optimize a fourth power to 2 multiplie

Is this similar to what you see? https://godbolt.org/z/qzMbedE8T

dholladay00: I modified slightly, you can see the individual function this way, and you'll see the recursive has 3 multiplies, the non-recursive has 2:
https://godbolt.org/z/5Tda4s5GY

mauneyc-LANL: > I modified slightly, you can see the individual function this way, and you'll see the recursive has 3 multiplies, the non-recursive has 2: https://godbolt.org/z/5Tda4s5GY

After spending *way* too much time figuring this out, the issue here is getting the compiler do "reduction"; with `fast-math` enabled, the compiler would be allowed to "flatten" the recursive stack, and give the optimal sequence of multiplies.

The solution is interesting, but rather bloated (at least in my attempts) https://godbolt.org/z/roqdMxKc3 (with side-by-side comparison of two methods)

```c++
template <unsigned N, typename std::enable_if<N == 1>::type* = nullptr>
constexpr auto sg_pow(const Real& value) {
  return value;
}

template <unsigned N, typename std::enable_if<N % 2 == 0>::type* = nullptr>
constexpr auto sg_pow(const Real& value) {
  return sg_pow<N / 2>(value) * sg_pow<N / 2>(value);
}

template <unsigned N, typename std::enable_if<N % 2 != 0>::type* = nullptr>
constexpr auto sg_pow(const Real& value) {
  return value * sg_pow<N - 1>(value);
}
```

Yurlungur: This is neat, but I think it's probably overkill here. Agreed that we could move these simple utilities into a header in base. @mauneyc-LANL @dholladay00 what about just specializing the template to all the powers we actually use? That way we have a clean interface but don't have to fight with the compiler to get it to optimize.

mauneyc-LANL: > This is neat, but I think it's probably overkill here. Agreed that we could move these simple utilities into a header in base. @mauneyc-LANL @dholladay00 what about just specializing the template to all the powers we actually use? That way we have a clean interface but don't have to fight with the compiler to get it to optimize.

what @dholladay00 suggested is easiest and correct, I just got carried away in godbolt

singularity-eos/eos/eos_davis.hpp
Yurlungur: :+1:

singularity-eos/eos/eos_davis.hpp
Yurlungur: I ended up resolving this now.

mauneyc-LANL (approved):

singularity-eos/eos/eos_stellar_collapse.hpp
Yurlungur: done

singularity-eos/eos/eos_spiner.hpp
Yurlungur: actually I'm gonna put this one off and make an issue.

singularity-eos/eos/eos_davis.hpp
Yurlungur: Resolved as discussed in the standup today: Now `singularity::math_utils::pow<N>` with explicit specializations for the even expansions

jhp-lanl (approved):

jhp-lanl:
Is this ready to be merged?

ktsai7:
> Tests pass and we're ready for merge. I'm going to merge.
> 
> > gitlab check isn't showing up in the CI? I see your branch on gitlab and it passes, but it's nice to have the check appear here too. Does the MR need to have the same name in addition to having the same branch name?
> 
> No I don't know. I didn't change anything. @ktsai7 and/or @mauneyc-LANL ?

No, I haven't changed anything related to github for a while...

PR #177, opened 2022-09-01
EOSPAC vector implementations

rbberger:
## PR Checklist


singularity-eos/eos/eos_eospac.hpp
```diff
@@ -72,103 +72,1156 @@ class EOSPAC : public EosBase<EOSPAC> {
...
+            // We need a different DPDE call apparently????
+            // In xRAGE, they call out to P(rho,e) in this case to get the
+            // derivative directly from the half-inverted table.
+            // But upon further review,
+            // I think it will end up evaluating to BMOD_T in any case because cv will
+            // be zero! See eos_eospac.f90 line 1261 BMOD =
+            // BMOD_T+DPDE*DPDE*std::max(DEDT,0.0)*T[0]/rho;
+            // BMOD_T = std::max(rho * DPDR[i], 0.0);
+            // BMOD = BMOD_T; //+DPDE*DPDE*std::max(DEDT,0.0)*T[0]/rho;
+            BMOD = std::max(rho * DPDR[i], 0.0);
```
jhp-lanl, line 312: Honestly there's not much we can do when the heat capacity (DEDT) is zero. If it's negative then the EOS is fundamentally flawed.

I'd be in favor of an assert here so that if somebody gets crazy results they can run in debug mode and see that their EOS has serious issues, but since this code was here before your modifications I'll create an issue for this so we can handle it later.

singularity-eos/eos/eos_eospac.hpp
```diff
@@ -72,103 +72,1156 @@ class EOSPAC : public EosBase<EOSPAC> {
...
+  inline void
+  TemperatureFromDensityInternalEnergy(ConstRealIndexer &&rhos, ConstRealIndexer &&sies,
+                                       RealIndexer &&temperatures, Real *scratch,
+                                       const int num, LambdaIndexer &&lambdas) const {
+    using namespace EospacWrapper;
+    EOS_REAL *R = scratch;
+    EOS_REAL *E = scratch + num;
+    EOS_REAL *T = scratch + 2 * num;
+    EOS_REAL *dTdr = scratch + 3 * num;
+    EOS_REAL *dTde = scratch + 4 * num;
```
jhp-lanl, line 118: We should probably have a way to compute the required scratch space and hand that to the user. I'm not sure the best way to do this though.

singularity-eos/eos/eos_eospac.hpp
```diff
@@ -72,103 +72,1156 @@ class EOSPAC : public EosBase<EOSPAC> {
...
-  //                                              LambdaIndexer &&lambdas) const;
-
-  // template<typename RealIndexer, typename LambdaIndexer>
-  // inline
-  // void FillEos(RealIndexer &&rhos, RealIndexer &&temps, RealIndexer &&energies,
-  //              RealIndexer &&presses, RealIndexer &&cvs, RealIndexer &&bmods,
-  //              const int num, const unsigned long output,
-  //              LambdaIndexer &&lambdas) const;
+  // EOSPAC vector implementations
+  template <int CHUNK_SIZE = 64, typename RealIndexer, typename ConstRealIndexer,
```
jhp-lanl, line 76: I assume that if you want to use the `CHUNK_SIZE` template parameter, you need to use the `EOSPAC` class directly since this isn't exposed through the `Variant` class right? I don't think I'd advocate for any changes to the `Variant` class to accommodate this yet since it's a requirement specific to EOSPAC and unlikely to be needed by other EOS.

singularity-eos/eos/eos_eospac.hpp
```diff
@@ -72,103 +72,1156 @@ class EOSPAC : public EosBase<EOSPAC> {
...
+      }
+
+      EOS_INTEGER table = EofRT_table_;
+      eosSafeInterpolate(&table, nelements, R.data(), T.data(), E.data(), DEDR.data(),
+                         DEDT.data(), "EofRT", Verbosity::Quiet);
+
+      for (int i = 0; i < nelements; ++i) {
+        int j = istart + i;
+        cvs[j] =
+            cvFromSesame(std::max(DEDT[i], 0.0)); // Here we do something to the data!
```
jhp-lanl, line 596: I think we should get rid of this `max` but again since this was here before, I don't think it needs to be changed in this MR. Stuff like this should be handled in a production hardening layer I think.

jhp-lanl (changes requested):
I think this looks really good! I don't think the underlying math and thermodynamics you copied over from the existing code has been thoroughly vetted, but that's beyond the scope of this MR. There may be additional performance improvements we can add, but those can be done later.

The only changes I would ask for would be a simple function to provide a required scratch size and a simple test. There are multiple scratch sizes needed though so this requires some discussion I think. For the test, I would just pick one of these methods and make a test for the chunk version and the full array version and make sure the result agrees with a for loop over scalar lookups. We can expand the test later as part of vetting the underlying logic and math.

singularity-eos/eos/eos_eospac.hpp
jhp-lanl: Actually, can't we just use the `lambdas` variable for scratch? EOSPAC currently says it doesn't need any lambdas, but we could set that requirement to 8 to account for the largest possible scratch usage.

In fact, we could also overload `nlambdas()` to take an integer corresponding to the number of points to ease the calculation of lambda arrays for spiner as well.

singularity-eos/eos/eos_eospac.hpp
rbberger: Yes, that's the intention. It's a tradeoff between using more on-demand stack vs preallocating a huge heap scratch space based of `nscratch() * nelements * sizeof(Real)`. Which of the two implementations or if both should stay is still up for debate. I can imagine use cases for either one of them.

singularity-eos/eos/eos_eospac.hpp
rbberger: The required scratch space is `nscratch() * elements * sizeof(Real)`. We could have a helper function that just takes `nelements` and returns the number of bytes.

rbberger: One possible solution for returning the exact scratch space requirement per call is having a function that returns the right value for each member function pointer.

singularity-eos/eos/eos_eospac.hpp
```diff
@@ -72,103 +72,1156 @@ class EOSPAC : public EosBase<EOSPAC> {
...
+      EOS_INTEGER table = TofRE_table_;
+      eosSafeInterpolate(&table, nelements, R.data(), E.data(), T.data(), dTdr.data(),
+                         dTde.data(), "TofRE", Verbosity::Quiet);
+
+      for (int i = 0; i < nelements; ++i) {
+        int j = istart + i;
+        temperatures[j] = temperatureFromSesame(T[i]);
+      }
+    }
+  }
```
Yurlungur, line 106: Have you tested the chunk size version of the code on GPUs? I'm pretty sure as written, this will not work, as the chunk are filled on CPU, whereas the data may live on GPU. This might be fixable with synchronized memory or with explicit deep copies, but it's something to be aware/concerned about.

singularity-eos/eos/eos_eospac.hpp
```diff
@@ -72,103 +72,1156 @@ class EOSPAC : public EosBase<EOSPAC> {
...
+      pressures[i] = pressureFromSesame(P[i]);
+    }
+  }
+
+  template <int CHUNK_SIZE = 64, typename RealIndexer, typename LambdaIndexer>
+  inline void FillEos(RealIndexer &&rhos, RealIndexer &&temps, RealIndexer &&energies,
+                      RealIndexer &&presses, RealIndexer &&cvs, RealIndexer &&bmods,
+                      const int num, const unsigned long output,
+                      LambdaIndexer &&lambdas) const {
+    using namespace EospacWrapper;
```
Yurlungur, line 195: It's pretty painful how much duplicated code lives in `FillEos`. I wonder if there's a way to reduce this.

singularity-eos/eos/eos_eospac.hpp
Yurlungur: I think this probably merits a larger discussion.

Yurlungur (commented):
This is a draft, so not approving or requesting changes. A little bit of feedback below.

Yurlungur:
Also thanks for putting in the work to prototype this! I like both the "scatch" and "chunk" approaches. But I think we should be a little thoughtful of GPU performance in both cases. I don't think any of this code will run on GPU as-is.

singularity-eos/eos/eos_eospac.hpp
jhp-lanl: Again like you said, larger discussion. But one last comment and I'll leave it to another time.

My changes to the Gruneisen EOS were mainly targeted at sanitizing inputs to produce realistic outputs. Since there's no way to _a priori_ know what inputs will lead to bad inputs for sesame tables, I feel that it's not necessarily our place to modify what the EOS returns _without the user explicitly asking for it_. That's why I argue for a production hardening modifier to handle some of these things. The behavior of the modifier can be documented so that the user knows what they are guaranteed to be handed. I feel that without asking for a modifier like this, the user should be guaranteed the best representation of the EOS itself even if that data is garbage.

But again, this needs a larger discussion beyond the scope of this MR.

singularity-eos/eos/eos_eospac.hpp
jhp-lanl: Discussion: https://github.com/lanl/singularity-eos/issues/180

singularity-eos/eos/eos_eospac.hpp
jhp-lanl: Could the chunk version call the regular version but with chunked data?

singularity-eos/eos/eos_eospac.hpp
rbberger: @Yurlungur Added a `max_scratch_size(nelements)` method

singularity-eos/eos/eos_eospac.hpp
```diff
@@ -72,103 +72,1156 @@ class EOSPAC : public EosBase<EOSPAC> {
...
+    EOS_REAL *R = scratch;
+    EOS_REAL *E = scratch + num;
+    EOS_REAL *T = scratch + 2 * num;
+    EOS_REAL *dTdr = scratch + 3 * num;
+    EOS_REAL *dTde = scratch + 4 * num;
+
+    for (int i = 0; i < num; ++i) {
+      R[i] = rhos[i];
+      E[i] = sieToSesame(sies[i]);
+    }
```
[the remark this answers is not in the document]
rbberger: Updated all loops to portableFor in those cases.

rbberger:
@Yurlungur @jhp-lanl I've updated all scratch and chunk versions to use EOSPAC options to minimize copies and do unit conversions. I've also expanded the profiling to include more methods and made it part of CTest. Below are timings I've generated per method on `skylake-platinum` nodes on Darwin:
EOSPAC_vectorization_time

EOSPAC_vectorization_throughput

jhp-lanl:
Great work @rbberger ! I'm excited especially that this work is part of `CTest` so that we can have a baseline and look for ways to improve. We should discuss this more at the meeting tomorrow!

rbberger:
@jhp-lanl yeah, having it part of CTest already pays off. I'm having weird failures on power9 that need further investigation. :smile:

Yurlungur:
Nice work @rbberger ! I second @jhp-lanl 's comments.

Yurlungur:
Peanut gallery says we go to scratch and drop lambdas and chunks

rbberger:
@Yurlungur @jhp-lanl Issue #187 blocks me from doing any meaningful GPU tests on the CI.

jhp-lanl:
> @Yurlungur @jhp-lanl Issue #187 blocks me from doing any meaningful GPU tests on the CI.

I seem to recall that the accessible EOSPAC spackage isn't up-to-date and I'm not quite sure what the best way to get 6.5 is. I'll ping Daniel Sheppard.

rbberger:
@Yurlungur ready for review.

singularity-eos/eos/eos_eospac.hpp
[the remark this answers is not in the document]
jhp-lanl: Since we decided to punt on getting EOSPAC working on GPUs, I'm resolving

python/module.cpp
```diff
@@ -113,9 +113,12 @@ PYBIND11_MODULE(singularity_eos, m) {
 #endif
 
 #ifdef SINGULARITY_USE_EOSPAC
-  eos_class<EOSPAC>(m, "EOSPAC")
+  eos_class<EOSPAC, true>(m, "EOSPAC")
```
jhp-lanl, line 116: I assume it's the `use_scratch` template parameter below, right? Would it make sense to add a comment or just introduce a `bool eospac_use_scratch = true` variable to explain this since it deviates from the construction of the other classes?

python/module.hpp
```diff
@@ -68,13 +68,32 @@ void func(const T & self, py::array_t<Real> a, py::array_t<Real> b,
...
+  py::buffer_info lambdas_info = lambdas.request();                               \
+  if (lambdas_info.ndim != 2)                                                     \
+      throw std::runtime_error("lambdas dimension must be 2!");                   \
+                                                                                  \
+  if(lambdas_info.shape[1] > 0) {                                                 \
+    self.func(a.data(), b.data(), out.mutable_data(), scratch.mutable_data(), num, LambdaHelper(lambdas)); \
+  } else {                                                                        \
+    self.func(a.data(), b.data(), out.mutable_data(), scratch.mutable_data(), num, NoLambdaHelper());      \
+  }                                                                               \
+}                                                                                 \
```
jhp-lanl, line 85: Is the "WithScratch" variant of the functions tested somewhere?

singularity-eos/eos/eos_eospac.hpp
```diff
@@ -89,86 +88,640 @@ class EOSPAC : public EosBase<EOSPAC> {
...
+
+    table = PofRT_table_;
+    {
+      EOS_INTEGER options[]{EOS_XY_PASSTHRU};
+      EOS_REAL values[]{1.0};
+      EOS_INTEGER nopts = 1;
+      eosSafeInterpolate(&table, num, R, T, P, DPDR, DPDT, "PofRT", Verbosity::Quiet,
+                         options, values, nopts);
+    }
```
jhp-lanl, line 418: Just a comment for the future for all these derivative quantities that require two lookups: we'll want to work with the EOSPAC team to find more performant ways to do this. If you _just_ want the bulk modulus and nothing else, we can probably accomplish this with a single lookup and let EOSPAC handle the thermodynamics to figure out how to do it.

Again... no changes needed as this is a future consideration and I'll make an issue.

test/CMakeLists.txt
```diff
@@ -81,3 +81,30 @@ if (SINGULARITY_BUILD_PYTHON AND SINGULARITY_TEST_PYTHON)
...
+
+  foreach(MUT IN LISTS METHODS_TO_TEST)
+    add_executable(profile_eospac_${MUT} profile_eospac.cpp)
+    target_compile_definitions(profile_eospac_${MUT} PUBLIC METHOD_UNDER_TEST=${MUT})
+    target_link_libraries(profile_eospac_${MUT} singularity-eos::singularity-eos)
+
+    add_test(NAME Profile_EOSPAC_${MUT}
+             COMMAND profile_eospac_${MUT} 100 16)
+  endforeach()
+endif()
```
jhp-lanl, line 110: Thanks again for adding this profiling test! This is excellent and we need to do more of this!

test/profile_eospac.cpp
```diff
@@ -0,0 +1,360 @@
...
+//#define METHOD_UNDER_TEST TemperatureFromDensityInternalEnergy
+//#define METHOD_UNDER_TEST PressureFromDensityTemperature
+//#define METHOD_UNDER_TEST InternalEnergyFromDensityTemperature
+//#define METHOD_UNDER_TEST PressureFromDensityInternalEnergy
+//#define METHOD_UNDER_TEST SpecificHeatFromDensityTemperature
+//#define METHOD_UNDER_TEST SpecificHeatFromDensityInternalEnergy
+//#define METHOD_UNDER_TEST BulkModulusFromDensityTemperature
+//#define METHOD_UNDER_TEST BulkModulusFromDensityInternalEnergy
+//#define METHOD_UNDER_TEST GruneisenParamFromDensityTemperature
+//#define METHOD_UNDER_TEST GruneisenParamFromDensityInternalEnergy
```
jhp-lanl, line 79: Were these for debugging or do we need to leave them in?

test/CMakeLists.txt
```diff
@@ -81,3 +81,30 @@ if (SINGULARITY_BUILD_PYTHON AND SINGULARITY_TEST_PYTHON)
            COMMAND ${Python_EXECUTABLE} ${CMAKE_CURRENT_SOURCE_DIR}/python_bindings.py -v)
   set_tests_properties(PythonBindings PROPERTIES ENVIRONMENT "${PYTHON_TEST_ENVIRONMENT}")
 endif()
+
+add_executable(profile_eos profile_eos.cpp)
+target_link_libraries(profile_eos singularity-eos::singularity-eos)
+
+if(SINGULARITY_USE_EOSPAC AND SINGULARITY_TEST_SESAME AND NOT SINGULARITY_USE_CUDA)
```
jhp-lanl, line 88: This may be my own ignorance, but do we need to exempt CUDA or can it operate on the non-Kokkos code for the test?

singularity-eos/eos/eos_eospac.hpp
```diff
@@ -89,86 +88,640 @@ class EOSPAC : public EosBase<EOSPAC> {
...
+        {"SpecificHeatFromDensityTemperature", 2},
+        {"SpecificHeatFromDensityInternalEnergy", 3},
+        {"BulkModulusFromDensityTemperature", 6},
+        {"BulkModulusFromDensityInternalEnergy", 6},
+        {"GruneisenParamFromDensityTemperature", 4},
+        {"GruneisenParamFromDensityInternalEnergy", 5},
+        {"PTofRE", 11}};
+    return nbuffers;
+  }
+
```
jhp-lanl, line 724: I like this!

jhp-lanl (approved):
I'm nominally approving this because most of my comments are questions rather than critiques and I can be convinced out of any code changes.

@rbberger feel free to resolve any of my questions as you answer them or make code changes to address them.

singularity-eos/eos/eos_eospac.hpp
```diff
@@ -89,86 +88,640 @@ class EOSPAC : public EosBase<EOSPAC> {
...
+        {"InternalEnergyFromDensityTemperature", 2},
+        {"PressureFromDensityInternalEnergy", 3},
+        {"SpecificHeatFromDensityTemperature", 2},
+        {"SpecificHeatFromDensityInternalEnergy", 3},
+        {"BulkModulusFromDensityTemperature", 6},
+        {"BulkModulusFromDensityInternalEnergy", 6},
+        {"GruneisenParamFromDensityTemperature", 4},
+        {"GruneisenParamFromDensityInternalEnergy", 5},
+        {"PTofRE", 11}};
+    return nbuffers;
```
Yurlungur, line 722: Since the scratch methods are available in the variant, should the scratch size methods be available too? We can implement the "null" versions in the base class and just return 0?

test/CMakeLists.txt
```diff
@@ -83,3 +83,30 @@ if (SINGULARITY_BUILD_PYTHON AND SINGULARITY_TEST_PYTHON)
...
+  )
+
+  foreach(MUT IN LISTS METHODS_TO_TEST)
+    add_executable(profile_eospac_${MUT} profile_eospac.cpp)
+    target_compile_definitions(profile_eospac_${MUT} PUBLIC METHOD_UNDER_TEST=${MUT})
+    target_link_libraries(profile_eospac_${MUT} singularity-eos::singularity-eos)
+
+    add_test(NAME Profile_EOSPAC_${MUT}
+             COMMAND profile_eospac_${MUT} 100 16)
+  endforeach()
```
Yurlungur, line 111: This is clever.

Yurlungur (changes requested):
Sorry I left this so long @rbberger it slipped my mind. Overall this looks very good. A few nitpicks below.

singularity-eos/eos/eos_base.hpp
```diff
@@ -79,6 +79,13 @@ class EosBase {
         });
   }
   template <typename RealIndexer, typename ConstRealIndexer, typename LambdaIndexer>
+  inline void
+  TemperatureFromDensityInternalEnergy(ConstRealIndexer &&rhos, ConstRealIndexer &&sies,
+                                       RealIndexer &&temperatures, Real *, const int num,
```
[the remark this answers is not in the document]
rbberger: In C++ 17 I could use `[[maybe_unused]]` in this context to still give it a name, while silencing the compiler warning. As a compromise I could do `/* scratch */` for now.

test/profile_eospac.cpp
rbberger: It was left as a comment to document how to test different methods. The CMake logic generates executables with each one of these defines. I have no preferences either way.

test/CMakeLists.txt
rbberger: I remember I had to do this because it was failing the regit CI otherwise.

test/profile_eospac.cpp
jhp-lanl: We can leave them, but can you add a comment above to just explain what's going on here?

singularity-eos/eos/eos_base.hpp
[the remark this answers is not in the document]
jhp-lanl: I'll defer to your judgement @rbberger

singularity-eos/eos/eos_eospac.hpp
jhp-lanl: I remember discussing this before and I thought we decided that it was specific enough to EOSPAC that a host code would need to differentiate anyway. However, I'm less convinced now that that's the right way to go, especially if we use EOSPAC in the PTE solvers. I'd support including null versions in the variant.

python/module.cpp
rbberger: Fixed.

singularity-eos/eos/eos_eospac.hpp
```diff
@@ -89,86 +88,640 @@ class EOSPAC : public EosBase<EOSPAC> {
...
-  //              LambdaIndexer &&lambdas) const;
+  // EOSPAC vector implementations
+  template <typename RealIndexer, typename ConstRealIndexer, typename LambdaIndexer>
+  inline void
+  TemperatureFromDensityInternalEnergy(ConstRealIndexer &&rhos, ConstRealIndexer &&sies,
+                                       RealIndexer &&temperatures, Real *scratch,
+                                       const int num, LambdaIndexer &&lambdas) const {
+    static auto const name =
+        singularity::mfuncname::member_func_name(typeid(EOSPAC).name(), __func__);
+    static auto const cname = name.c_str();
```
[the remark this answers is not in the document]
rbberger: Removed the ones that weren't used.

test/profile_eospac.cpp
rbberger: Added a comment above.

eospac-wrapper/eospac_wrapper.cpp
```diff
@@ -184,7 +196,17 @@ bool eosSafeInterpolate(EOS_INTEGER *table, EOS_INTEGER nxypairs, EOS_REAL xVals
...
+
+  for (int i = 0; i < nopts; i++) {
+    eos_ResetOption(table, &options[i], &errorCode);
+    if (errorCode != EOS_OK) {
+      eos_GetErrorMessage(&errorCode, errorMessage);
+      std::cerr << "Table " << tablename << ":" << std::endl;
+      std::cerr << "eos_ResetOption ERROR " << errorCode << ": " << errorMessage
+                << std::endl;
+    }
+  }
```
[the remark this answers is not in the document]
rbberger: Didn't measure that, but my guess it is smaller than setting ALL options to a defined state before each call, which would be necessary if you don't reset them right away. Otherwise, you can't rely on the option values, given that you don't know what other EOSPAC calls have happened before each function.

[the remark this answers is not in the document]
rbberger: While the table handle is associated with the EOSPAC class, how do we know im which order the available functions are called? At a minimum we need to set all options that potentially could have been set by any member function to avoid interference.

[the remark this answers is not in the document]
rbberger: Ok, as an example: if you look at the usage of the `PofRT_table_`, the calls happen sometimes with `F_CONVERT` and sometimes with `X_Y_PASSTHROUGH` options passed. I can not predict in what order the different EOS functions are called, so I have to make sure the options are in a defined state.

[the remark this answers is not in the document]
Yurlungur: Ok I see. Let's leave this for now then. Maybe worth putting a TODO to talk with the EOSPAC folks about though. If setting options implies a performance hit we might want to redesign the class to avoid having to do so.

singularity-eos/eos/eos_base.hpp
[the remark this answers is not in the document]
Yurlungur: Let's do `/* scratch */` as we can't get `C++17` yet.

singularity-eos/eos/eos_eospac.hpp
rbberger: Done.

singularity-eos/eos/eos_base.hpp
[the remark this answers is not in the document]
rbberger: This is already done.

singularity-eos/eos/eos_ideal.hpp
```diff
@@ -106,6 +106,10 @@ class IdealGas : public EosBase<IdealGas> {
   PORTABLE_INLINE_FUNCTION
   int nlambda() const noexcept { return 0; }
   static constexpr unsigned long PreferredInput() { return _preferred_input; }
+  static inline unsigned long scratch_size(std::string method, unsigned int nelements) {
+    return 0;
+  }
+  static inline unsigned long max_scratch_size(unsigned int nelements) { return 0; }
```
Yurlungur, line 112: These "0 size" funcs could probably go in the base class. Either way though. :)

singularity-eos/eos/eos_variant.hpp
```diff
@@ -267,6 +267,28 @@ class Variant {
...
+  TemperatureFromDensityInternalEnergy(ConstRealIndexer &&rhos, ConstRealIndexer &&sies,
+                                       RealIndexer &&temperatures, Real *scratch,
+                                       const int num, LambdaIndexer &&lambdas) const {
+    return mpark::visit(
+        [&rhos, &sies, &temperatures, &scratch, &num, &lambdas](const auto &eos) {
+          return eos.TemperatureFromDensityInternalEnergy(rhos, sies, temperatures,
+                                                          scratch, num, lambdas);
+        },
+        eos_);
+  }
```
Yurlungur, line 291: Should the non-scratch vector calls be removed? Or perhaps just default scratch to a nullptr?

Yurlungur (approved):
I approve and am ready to merge, pending the items on the checklist being checked off.

A few things on my mind. But we don't have to resolve these now, we can merge first and tackle in a second PR.
- Should we have a correctness test? Perhaps use the vector calls and the scalar calls for the EOSPAC backend and compare? Or alternatively compare EOSPAC and spiner, as is currently done, but convert the EOSPAC calls to vector calls? We could make it work for CPU I imagine?
- Still need to get GPU working, but that's for later as discussed previously.
- Make sure to trigger tests on Darwin and make sure they pass. :)
- Make sure to do changelog, CC, etc.
- Perhaps the scratch API and the particulars of the EOSPAC passthrough should be documented
- We'll have to figure out how to do the passthrough for modifiers... I think something is needed there. Probably we can leverage the scratch machinery.

singularity-eos/eos/eos_variant.hpp
jhp-lanl: Doesn't the python interface benefit from the non-scratch vector calls?

singularity-eos/eos/eos_eospac.hpp
```diff
@@ -89,86 +88,640 @@ class EOSPAC : public EosBase<EOSPAC> {
...
-
-  // template<typename RealIndexer, typename LambdaIndexer>
-  // inline
-  // void FillEos(RealIndexer &&rhos, RealIndexer &&temps, RealIndexer &&energies,
-  //              RealIndexer &&presses, RealIndexer &&cvs, RealIndexer &&bmods,
-  //              const int num, const unsigned long output,
-  //              LambdaIndexer &&lambdas) const;
+  // EOSPAC vector implementations
+  template <typename RealIndexer, typename ConstRealIndexer, typename LambdaIndexer>
+  inline void
```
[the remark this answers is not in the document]
rbberger: I opted for removing the template parameters and just forcing pointer types. A `static_assert` or SFINAE with `enable_if_t` would have been a bit over-engineered in my opinion.

singularity-eos/eos/eos_ideal.hpp
rbberger: I followed the patterns I saw in PreferredInput and other functions.

rbberger:
> * Should we have a correctness test? Perhaps use the vector calls and the scalar calls for the EOSPAC backend and compare? Or alternatively compare EOSPAC and spiner, as is currently done, but convert the EOSPAC calls to vector calls? We could make it work for CPU I imagine?

Actually, the `profile_eospac` test does include a comparison with the scalar versions.

singularity-eos/eos/eos_variant.hpp
Yurlungur: Ah good point. :+1:

python/module.hpp
rbberger: @jhp-lanl just added Python tests that do so: https://github.com/lanl/singularity-eos/blob/59a2e74f784d6b1ce36304331893a74f62512c89/test/python_bindings.py#L528

jhp-lanl (approved):
This looks really good @rbberger ! Thanks for all the effort on this!

jhp-lanl:
I'll wait for the Darwin tests to pass before we merge.

Yurlungur (approved):
Looks good. :+1: Let's merge as soon as Darwin tests pass

jhp-lanl:
@rbberger is there anything holding this up? To me it looks like you can merge in master (and resolve conflicts) and we can merge this.

rbberger:
> @rbberger is there anything holding this up? To me it looks like you can merge in master (and resolve conflicts) and we can merge this.

No. It kind of got stuck because we were looking at why GPU builds being so slow (in general, not just this), but I'll fix this one up again so we can merge it.

rbberger:
@Yurlungur re-git pipeline successful. feel free to merge.

Yurlungur:
Thanks for all your hard work on this @rbberger ! At some point we should check to see what is necessary to pull this through for GPU implementations and modifiers. But for now I think we can celebrate speeding up the EOSPAC calls.

jhp-lanl:
Yes indeed! Thanks for all the hard work @rbberger !

PR #183, opened 2022-09-28
Checks for found and target in singularity-eosConfig.cmake

mauneyc-LANL:
## PR Summary

The exported configuration checked for `Eigen3_FOUND` and passed, but the concurrent processing wasn't able to find the interface target `Eigen3::Eigen`. This updates the configuration check to check both.

## PR Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - [[PR181]](https://github.com/lanl/singularity-eos/pull/181) change from manual dependecy handling to using hdf5 interface targets
 
 ### Fixed (Repair bugs, etc)
+- [[PR183]](https://github.com/lanl/singularity-eos/pull/183) fortify cmake export config to always have interface targets of dependencies that need them
 - [[PR174]](https://github.com/lanl/singularity-eos/pull/174) fix build configuration when closures are disabled
 - [[PR157]](https://github.com/lanl/singularity-eos/pull/157) fix root finders in Gruneisen, Davis, and JWL
 - [[PR151]](https://github.com/lanl/singularity-eos/pull/151) fix module install
--- a/cmake/dependency.cmake
+++ b/cmake/dependency.cmake
@@ -131,6 +131,9 @@ macro(singularity_import_dependency)
   if(dep_COMPONENTS)
     list(APPEND SINGULARITY_DEP_PKGS_${dep_PKG} "${dep_COMPONENTS}")
   endif()
+  if(dep_TARGET)
+    set(SINGULARITY_DEP_TARGET_${dep_PKG} ${dep_TARGET})
+  endif()
   unset(_SMSG_PREFIX)
 endmacro() # singularity_import_dependency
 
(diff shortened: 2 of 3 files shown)


Yurlungur:
@mauneyc-LANL the minimal test is failing because eigen is being included somewhere, despite closures being disabled. Might be a missing `#ifdef` in `singularity_eos.cpp`.

Yurlungur (commented):
I don't understand these changes or their impact. But I'll trust you that they're correct. I'll approve once the boxes are all checked.

mauneyc-LANL:
> I don't understand these changes or their impact. But I'll trust you that they're correct. I'll approve once the boxes are all checked.

In the `singularity-eosConfig.cmake` file, what was happening was

```cmake
if(NOT Eigen3_FOUND)
        find_package(Eigen3  REQUIRED)
endif()
```

What was happening when building downstream was `Eigen3_FOUND` was set by an earlier part of the build, but for some reason the target `Eigen3::Eigen` was not finding it's was down to `singularity-eos`. That is our target `singularity-eos::singlarity-eos` had an link dependency on `Eigen3::Eigen`, but at the end of the configuration step, CMake was complaining that `singularity-eos depends on target Eigen3::Eigen, but that target was not found`.

This is a bit brutish, but now `singularity-eosConfig.cmake` will have

```cmake
if(NOT Eigen3_FOUND OR NOT TARGET Eigen3::Eigen)
        find_package(Eigen3  REQUIRED)
endif()
```

So that, if `Eigen3::Eigen` isn't a defined target, CMake will engage the `find_package()` tools to explicitly make it.

dholladay00:
If I am understanding correctly, this change will be in effect for all dependencies @mauneyc-LANL? It seems like a good addition if we want to maintain in-tree and externally built capabilities for our dependencies.

mauneyc-LANL:
> If I am understanding correctly, this change will be in effect for all dependencies @mauneyc-LANL? It seems like a good addition if we want to maintain in-tree and externally built capabilities for our dependencies.

This effects any dependencies that _define an interface target that we adhere to_, e.g. `Eigen3::Eigen`, `ports-of-call`.

A counter-example is when we import `HDF5`, we no longer rely on an interface target and add the paths explicitly to the export config. This is b/c `HDF5` doesn't have a consistent interface target for a range of `cmake` & `hdf5` installs - that is, `HDF5::HDF5` could resolve to different configurations downstream (or not be defined at all), and I threw my hands up and just rely on exporting explicit paths.

I should stress that this is a sloppy attempt to fix an issue I haven't figured out - specifically, when I was importing `singularity-eos` into a host code, that `Eigen3_FOUND` was set, but CMake was complaining that `Eigen3::Eigen` was not a defined target. This is likely due to some conflict in how a separate dependency was importing `Eigen`, but I didn't try and track that down. This change will ensure that the target `Eigen3::Eigen` is set by `find_package()`, regardless of what prior configuration has done.

Yurlungur (approved):

PR #198, opened 2022-11-30
More robust hdf5, along with more refined I/O from cmake

mauneyc-LANL:
## PR Summary

Addressing issues @jdolence encountered with HDF5 resolution. Some small rearranging of dependency logic, and a less chatty cmake (by default, can be adjusted)

I had attempted to use `module` mode to resolve `find_package(HDF5)` with a distributed `FindHDF5.cmake`, but I think I made a mess of it and Josh hit an issue that arose from that. The large change here is to revert to a more "standard" approach that relies on the implementation of "official" Kitware CMake module (https://gitlab.kitware.com/cmake/cmake/-/blob/master/Modules/FindHDF5.cmake).

For reference, it can be informative to expose some trace information on `find_package` when configuring

```bash
cmake --debug-find-pkg=HDF5 .. -DSINGULARITY_USE_HDF5=ON #...
```

Other notable changes:
- Output has been reduced in the default log level. A more chatty output can be restored using `--log-level=VERBOSE` option: `cmake --log-level=VERBOSE .. #...`
- some cleanup of `singularity_import_*` macros

@jonahm-LANL @dholladay00 @jhp-lanl 

## PR Checklist


CMakeLists.txt
```diff
@@ -208,7 +208,7 @@ if(SINGULARITY_USE_KOKKOS)
   # type: soft submodule 
   singularity_import_dependency(
     PKG Kokkos
-    TARGET Kokkos::kokkos
+    TARGETS Kokkos::kokkos
```
Yurlungur, line 211: what is the difference between these?

Yurlungur (commented):
Don't forget to update changelog

CMakeLists.txt
mauneyc-LANL: An `S`

the macro took as input `TARGET` and `TARGETS`, These did the same thing, but just bifurcated into singular and plural. I changed this to only one input `TARGETS`

Yurlungur (approved):
Looks good to me. @jdolence can you double check this resolves the issues you were seeing?

dholladay00:
@jdolence what is your timeline on testing this? I'd like to get this in soon.

dholladay00:
@mauneyc-LANL can you resolve all incomplete items in the checklist? When completed, I'll merge. 

I suspect downstream testing will be necessary.

Yurlungur:
Yeah I think @jdolence is not going to test this. Assuming tests pass I'm fine with merging without waiting for him.

CMakeLists.txt
```diff
@@ -285,22 +285,12 @@ if (SINGULARITY_USE_HDF5)
   enable_language(C)
```
junghans, line 285: @mauneyc-LANL We can drop this for cmake>=3.26, see: https://gitlab.kitware.com/cmake/cmake/-/merge_requests/8015

mauneyc-LANL: yes; Indeed, we don't require to do this for `spiner`, and I'm working on a PR for singularity to follow that approach.

PR #201, opened 2022-12-02
WIP: Tilted Maxwell construction EoS modifier

lroberts36:
## PR Summary
Starting work on implementing an EoS modifier that will replace spinodal regions in an EoS with a "tilted" Maxwell construction in those regions in a thermodynamically consistent way. Within the coexistence region, we do not assume exact pressure equality between the two phases but allow for a difference in pressure so that the pressure is not constant across the coexistence region but it is still continuous across the phase boundaries. Although this is to some extent physically inconsistent, constant pressure regions potentially pose some problems for hydro codes and multi-material PTE solves.

## PR Checklist

### DIFF
--- a/singularity-eos/base/math_utils.hpp
+++ b/singularity-eos/base/math_utils.hpp
@@ -50,6 +50,104 @@ PORTABLE_FORCEINLINE_FUNCTION constexpr auto pow<4>(const Real x) {
   return pow<2>(x) * pow<2>(x);
 }
 
+template<class F>
+double GLQuad8(F&& f, double a = -1.0, double b = 1.0) {
+  static constexpr double u[4]{0.183434642495649,
+                               0.525532409916329,
+                               0.796666477413626,
+                               0.960289856497536};
+  
+  static constexpr double w[4]{0.362683783378362,
+                               0.313706645877887,
+                               0.222381034453374,
+                               0.101228536290376};
+  double val = 0.0;     
+  const double delta = 0.5 * (b - a);
+  const double offset = 0.5 * (b + a);       
+  for (int i = 0; i < 4; ++i) {
+    const double xm = offset - delta * u[i];
+    const double xp = offset + delta * u[i];
+    val += w[i] * (f(xp) + f(-xm));  
+  }
+  return delta * val;
+}
+
+template<class F>
+double GLQuad16(F&& f, double a = -1.0, double b = 1.0) {
+  static constexpr double u[8]{ 0.0950125098376374,
+                                0.2816035507792589,
+                                0.4580167776572274,
+                                0.6178762444026438,
+                                0.7554044083550030,
+                                0.8656312023878318,
+                                0.9445750230732326,
+                                0.9894009349916499};
+  
+  static constexpr double w[8]{0.1894506104550685,
+                               0.1826034150449236,
+                               0.1691565193950025,
+                               0.1495959888165767,
+                               0.1246289712555339,
+                               0.0951585116824928,
+                               0.0622535239386479,
+                               0.0271524594117541};
+  double val = 0.0;     
+  const double delta = 0.5 * (b - a);
+  const double offset = 0.5 * (b + a);       
+  for (int i=0; i < 8; ++i) {
+    const double xm = offset - delta * u[i];
+    const double xp = offset + delta * u[i];
+    val += w[i] * (f(xp) + f(xm));  
+  }
+  return delta * val;
+}
+
+template<class F>
+double GLQuad32(F&& f, double a = -1.0, double b = 1.0) {
+  static constexpr double u[16]{0.04830766568773834,
+                                0.14447196158279654,
+                                0.23928736225213719,
+                                0.33186860228212779,
+                                0.42135127613063534,
+                                0.50689990893222944,
+                                0.58771575724076238,
+                                0.66304426693021528,
+                                0.73218211874028970,
+                                0.79448379596794240,
+                                0.84936761373257008,
+                                0.89632115576605218,
+                                0.93490607593773976,
+                                0.96476225558750646,
+                                0.98561151154526849,
+                                0.99726386184948169};
+  
+  static constexpr double w[16]{0.09654008851472785,
+                                0.09563872007927495,
+                                0.09384439908080466,
+                                0.09117387869576396,
+                                0.08765209300440385,
+                                0.08331192422694675,
+                                0.07819389578707037,
+                                0.07234579410884857,
+                                0.06582222277636189,
+                                0.05868409347853559,
+                                0.05099805926237628,
+                                0.04283589802222678,
+                                0.03427386291302149,
+                                0.02539206530926219,
+                                0.01627439473090571,
+                                0.00701861000947011}; 
+  double val = 0.0;     
+  const double delta = 0.5 * (b - a);
+  const double offset = 0.5 * (b + a);       
+  for (int i=0; i < 16; ++i) {
+    const double xm = offset - delta * u[i];
+    const double xp = offset + delta * u[i];
+    val += w[i] * (f(xp) + f(xm));  
+  }
+  return delta * val;            
+}         
+
 } // namespace math_utils
 } // namespace singularity
 
--- a/singularity-eos/eos/modifiers/tilted_maxwell_eos.hpp
+++ b/singularity-eos/eos/modifiers/tilted_maxwell_eos.hpp
@@ -1,0 +1,314 @@
+//------------------------------------------------------------------------------
+// © 2021-2022. Triad National Security, LLC. All rights reserved.  This
+// program was produced under U.S. Government contract 89233218CNA000001
+// for Los Alamos National Laboratory (LANL), which is operated by Triad
+// National Security, LLC for the U.S.  Department of Energy/National
+// Nuclear Security Administration. All rights in the program are
+// reserved by Triad National Security, LLC, and the U.S. Department of
+// Energy/National Nuclear Security Administration. The Government is
+// granted for itself and others acting on its behalf a nonexclusive,
+// paid-up, irrevocable worldwide license in this material to reproduce,
+// prepare derivative works, distribute copies to the public, perform
+// publicly and display publicly, and to permit others to do so.
+//------------------------------------------------------------------------------
+
+#ifndef _SINGULARITY_EOS_EOS_TILTED_MAXWELL_EOS_
+#define _SINGULARITY_EOS_EOS_TILTED_MAXWELL_EOS_
+
+#include "stdio.h"
+#include <cassert>
+#include <cmath>
+#include <cstdlib>
+#include <iostream>
+#include <limits>
+#include <tuple>
+#include <utility>
+
+#include <ports-of-call/portability.hpp>
+#include <singularity-eos/base/constants.hpp>
+#include <singularity-eos/base/eos_error.hpp>
+#include <singularity-eos/base/math_utils.hpp>
+#include <singularity-eos/base/root-finding-1d/root_finding.hpp>
+#include <singularity-eos/eos/eos_base.hpp>
+
+namespace singularity {
+
+using namespace eos_base;
+using namespace RootFinding1D;
+
+template <typename T>
+class TiltedMaxwellEOS : public EosBase<TiltedMaxwellEOS<T>> {
+ public:
+  // Generic functions provided by the base class. These contain
+  // e.g. the vector overloads that use the scalar versions declared
+  // here We explicitly list, rather than using the macro because we
+  // overload some methods.
+
+  // TODO(JMM): The modifier EOS's should probably call the specific
+  // sub-functions of the class they modify so that they can leverage,
+  // e.g., an especially performant or special version of these
+  using EosBase<TiltedMaxwellEOS<T>>::TemperatureFromDensityInternalEnergy;
+  using EosBase<TiltedMaxwellEOS<T>>::InternalEnergyFromDensityTemperature;
+  using EosBase<TiltedMaxwellEOS<T>>::PressureFromDensityTemperature;
+  using EosBase<TiltedMaxwellEOS<T>>::PressureFromDensityInternalEnergy;
+  using EosBase<TiltedMaxwellEOS<T>>::SpecificHeatFromDensityTemperature;
+  using EosBase<TiltedMaxwellEOS<T>>::SpecificHeatFromDensityInternalEnergy;
+  using EosBase<TiltedMaxwellEOS<T>>::BulkModulusFromDensityTemperature;
+  using EosBase<TiltedMaxwellEOS<T>>::BulkModulusFromDensityInternalEnergy;
+  using EosBase<TiltedMaxwellEOS<T>>::GruneisenParamFromDensityTemperature;
+  using EosBase<TiltedMaxwellEOS<T>>::GruneisenParamFromDensityInternalEnergy;
+  using EosBase<TiltedMaxwellEOS<T>>::PTofRE;
+  using EosBase<TiltedMaxwellEOS<T>>::FillEos;
+
+  using BaseType = T;
+
+  // give me std::format or fmt::format...
+  static std::string EosType() {
+    return std::string("TiltedMaxwellEOS<") + T::EosType() + std::string(">");
+  }
+
+  static std::string EosPyType() { return std::string("Shifted") + T::EosPyType(); }
+
+  // move semantics ensures dynamic memory comes along for the ride
+  TiltedMaxwellEOS(T &&t, Real tilt_fac) : t_(std::forward<T>(t)), tilt_(tilt_fac) {}
+  TiltedMaxwellEOS() = default;
+
+  Real g(Real u) const { return (1.0 - tilt_) * u + tilt_ * u * u;} 
+  Real gp(Real u) const { return 1.0 - tilt_ + 2 * tilt_ * u;} 
+  
+  const Real tilt_;
+  static constexpr Real tol = 1.e-10;
+  static constexpr Real eos_rho_min = 1.e-8;
+  static constexpr Real eos_rho_max = 1.e2;
+
+  auto TestMaxwellConditions(Real pin, Real temp, Real rho_p_min, Real rho_p_max) const { 
+    // Find densities above and below spinodal region with p(rho_l) = pin and p(rho_h) = pin  
+    Real rho_l, rho_h;
+    RootCounts counts_l, counts_h;
+    auto stat_l = regula_falsi([this, temp](Real rho){return this->t_.PressureFromDensityTemperature(rho, temp);}, 
+                               pin, 0.8 * rho_p_max, 
+                               eos_rho_min, rho_p_max,
+                               tol, tol,
+                               rho_l, counts_l);
+    auto stat_h = regula_falsi([this, temp](Real rho){return this->t_.PressureFromDensityTemperature(rho, temp);}, 
+                               pin, 1.1 * rho_p_min, 
+                               rho_p_min, eos_rho_max,
+                               tol, tol,
+                               rho_h, counts_h);
+    // Integrate Maxwell construction integral (i.e. the chemical potential difference) 
+    Real maxwell_int = math_utils::GLQuad32([this, temp, pin](Real rho){return (this->t_.PressureFromDensityTemperature(rho, temp) - pin)/(rho*rho);}, rho_l, rho_h);
+    return std::make_tuple(maxwell_int, rho_l, rho_h);
+  }
+  
+  auto GetMaxwellConstruction(Real temp) const { 
+    constexpr double eps = std::sqrt(std::numeric_limits<Real>::epsilon());
+    
+    // Minimize dPdrho to split isotherm into two regions, if minimum is greater than zero 
+    // there is no spinodal region
+    auto dpdrho = [this, eps, temp](double rho) { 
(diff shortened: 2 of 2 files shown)


singularity-eos/base/math_utils.hpp
```diff
@@ -50,6 +50,104 @@ PORTABLE_FORCEINLINE_FUNCTION constexpr auto pow<4>(const Real x) {
   return pow<2>(x) * pow<2>(x);
 }
 
+template<class F>
+double GLQuad8(F&& f, double a = -1.0, double b = 1.0) {
```
Yurlungur, line 54: Nice.

singularity-eos/eos/modifiers/tilted_maxwell_eos.hpp
```diff
@@ -0,0 +1,314 @@
...
+  static constexpr Real tol = 1.e-10;
+  static constexpr Real eos_rho_min = 1.e-8;
+  static constexpr Real eos_rho_max = 1.e2;
+
+  auto TestMaxwellConditions(Real pin, Real temp, Real rho_p_min, Real rho_p_max) const { 
+    // Find densities above and below spinodal region with p(rho_l) = pin and p(rho_h) = pin  
+    Real rho_l, rho_h;
+    RootCounts counts_l, counts_h;
+    auto stat_l = regula_falsi([this, temp](Real rho){return this->t_.PressureFromDensityTemperature(rho, temp);}, 
+                               pin, 0.8 * rho_p_max,
```
Yurlungur, line 89: why this value? And not, say `0.990 * rho_p_max`?

singularity-eos/eos/modifiers/tilted_maxwell_eos.hpp
```diff
@@ -0,0 +1,314 @@
...
+    // Find densities above and below spinodal region with p(rho_l) = pin and p(rho_h) = pin  
+    Real rho_l, rho_h;
+    RootCounts counts_l, counts_h;
+    auto stat_l = regula_falsi([this, temp](Real rho){return this->t_.PressureFromDensityTemperature(rho, temp);}, 
+                               pin, 0.8 * rho_p_max, 
+                               eos_rho_min, rho_p_max,
+                               tol, tol,
+                               rho_l, counts_l);
+    auto stat_h = regula_falsi([this, temp](Real rho){return this->t_.PressureFromDensityTemperature(rho, temp);}, 
+                               pin, 1.1 * rho_p_min,
```
Yurlungur, line 94: and why this value?

singularity-eos/eos/modifiers/tilted_maxwell_eos.hpp
```diff
@@ -0,0 +1,314 @@
...
+    auto d2pdrho2 = [this, eps, temp](double rho) { 
+      return (this->t_.PressureFromDensityTemperature(rho * (1.0 + eps), temp) 
+            - 2*this->t_.PressureFromDensityTemperature(rho, temp)
+            + this->t_.PressureFromDensityTemperature(rho * (1.0 - eps), temp)) / (eps * eps * rho * rho);
+    };
+    Real rho_dpdrho_min = 0.0; 
+    RootCounts counts;
+    auto stat = regula_falsi(d2pdrho2, 
+                             0.0, 1.0, 
+                             0.1, 3.0,
```
Yurlungur, line 121: I notice lots of hardcoded numbers floating around. It would be helpful to know why these were chosen to be what they are and maybe have the variables stored as fields in a central location with descriptive names

singularity-eos/eos/modifiers/tilted_maxwell_eos.hpp
```diff
@@ -0,0 +1,314 @@
...
+                        rho_p_max, counts);
+    stat = regula_falsi(dpdrho, 
+                        0.0, 1.2 * rho_dpdrho_min, 
+                        rho_dpdrho_min, eos_rho_max, 
+                        tol, tol,
+                        rho_p_min, counts);
+    Real p_max = std::min(t_.PressureFromDensityTemperature(rho_p_max, temp), 
+                          t_.PressureFromDensityTemperature(eos_rho_max, temp));
+    Real p_min = std::max(t_.PressureFromDensityTemperature(rho_p_min, temp),
+                          t_.PressureFromDensityTemperature(eos_rho_min, temp));
```
Yurlungur, line 142: this is almost identical code to in the test. Is there some way to factor out the logic to find for the edges of the spinodal region so this code doesn't need to be duplicated?

singularity-eos/eos/modifiers/tilted_maxwell_eos.hpp
```diff
@@ -0,0 +1,314 @@
...
+                        p_min, p_max,
+                        tol, tol, 
+                        p_maxwell, counts);
+    const auto [f, rho_maxwell_low, rho_maxwell_hi] = TestMaxwellConditions(p_maxwell, temp, rho_p_min, rho_p_max); 
+
+    return std::make_tuple(rho_maxwell_low, rho_maxwell_hi, p_maxwell, true);
+  }
+
+  auto GetTiltedMaxwellConstruction(Real temp) const {
+    const auto [rho_maxwell_low, rho_maxwell_hi, p_maxwell, has_spinodal] = this->GetMaxwellConstruction(temp);
```
Yurlungur, line 159: I think singularity-eos is not yet C++17 although that's coming soon.

singularity-eos/eos/modifiers/tilted_maxwell_eos.hpp
```diff
@@ -0,0 +1,314 @@
...
+    if (!has_spinodal) return std::make_tuple(0.0, 0.0, 0.0, false); 
+    
+    auto func = [this, temp, p_maxwell](std::array<Real, 2> x){
+      const Real Pl = this->t_.PressureFromDensityTemperature(x[0], temp); 
+      const Real Ph = this->t_.PressureFromDensityTemperature(x[1], temp);
+      const Real PD = math_utils::GLQuad32([this, temp](Real rho){return this->t_.PressureFromDensityTemperature(rho, temp)/(rho * rho);}, x[0], x[1])/(1/x[0] - 1/x[1]);
+      return std::array<Real, 2>{(Pl - PD * this->gp(1.0)) / p_maxwell, (Ph - PD * this->gp(0.0)) / p_maxwell};
+    };
+    
+    // NR Iteration
```
Yurlungur, line 169: Should we have a 2D NR somewhere global?

Yurlungur (commented):
Assume Maxwell constructed switching logic not yet included?

singularity-eos/eos/modifiers/tilted_maxwell_eos.hpp
lroberts36: In answer to most of your questions below, right now all of these values are hard coded because I was working on getting it to work for a particular EoS. I have to think a bit about what parameters we need to set in the class constructor. There are many factors that show up for initializing guesses, etc., so it would be nice if we didn't have to expose all of them and could relate them all to a few user definable parameters.

singularity-eos/eos/modifiers/tilted_maxwell_eos.hpp
lroberts36: It is close, but different. It is finding the densities corresponding to the local minimum and maximum of the pressure (i.e. the boundaries of the spinodal region), while the code in `TestMaxwellConditions` is finding a densities in a particular region that correspond to a given pressure. I am not sure that there is enough code duplication that we can factor it out.

singularity-eos/eos/modifiers/tilted_maxwell_eos.hpp
lroberts36: yeah, I was worried about that. What is the timescale for "soon"?

singularity-eos/eos/modifiers/tilted_maxwell_eos.hpp
lroberts36: Maybe, I have not robustified this yet and it may be that there are some problem specific things that can be done.

lroberts36:
@jonahm: Right now, I am just working on writing code that will find the required boundaries. I think this is the trickiest part. Once the boundaries of the fixed up region can be found robustly, any (rho, T) call to the eos (e.g. for pressure, internal energy) requires a few lookups in one-dimensional tables and then evaluation of an explicit expression. What sort of array should I be storing these values in? I guess there is some type within singularity that can be copied from host to device?

singularity-eos/eos/modifiers/tilted_maxwell_eos.hpp
Yurlungur: :+1:

singularity-eos/eos/modifiers/tilted_maxwell_eos.hpp
Yurlungur: Months, likely? We are waiting on xrage to move to C++17.

Yurlungur:
What sort of array should I be storing these values in? 

How big will this array be? If it's only a few numbers, a `std::array` is probably fine... If larger... yeah probably not.

> I guess there is some type within singularity that can be copied from host to device?

Elsewhere in the code we use `Spiner::DataBox` for this. `DataBox` is what's behind the `SpinerEOS` and `StellarCollapseEOS` objects. 
https://lanl.github.io/spiner/main/index.html

Here's an example of using `DataBox` just to store data rather than interpolate:
```C++
Spiner::DataBox db; // There's also a constructor to set the size
db.resize(N);
for (int i = 0; i < db.size(); ++db) {
  db(i) = 2.2*i;
}
Spiner::DataBox dbOnDevice = db.getOnDevice();
// do something in a kernel

// you must call free when you're done. Databoxes are not reference counted.
free(db);
free(dbOnDevice);
```
There are more features---you can put them in smart pointers. They're multi-D. You can interpolate. But that's the basics.

jhp-lanl (commented):
Overall, my concern is that if you intend for this modifier to be applied to tabular data, there can be a lot of issues and I wouldn't expect an on-the-fly Maxwell construction to be very robust. Here are a few of the issues with tabular data:

1. Tabular EOS often do not contain enough density points to fully-resolve the behavior between the binodal and spinodal in the vapor phase, meaning that there is a very large uncertainty in that part of the integral. The net effect of this is that the uncertainty in the vapor pressure increases as the vapor pressure decreases.
2. You often have no guarantee here that the vapor pressure you get for one temperature will be reasonable compared to another temperature. In other words, you could find yourself in a situation where the vapor pressure is increasing with decreasing temperature? This could result in a very strange material response should you find yourself in this region of the EOS and I think is a bigger problem than a constant pressure region of the EOS.

singularity-eos/eos/modifiers/tilted_maxwell_eos.hpp
```diff
@@ -0,0 +1,314 @@
...
+                               eos_rho_min, rho_p_max,
+                               tol, tol,
+                               rho_l, counts_l);
+    auto stat_h = regula_falsi([this, temp](Real rho){return this->t_.PressureFromDensityTemperature(rho, temp);}, 
+                               pin, 1.1 * rho_p_min, 
+                               rho_p_min, eos_rho_max,
+                               tol, tol,
+                               rho_h, counts_h);
+    // Integrate Maxwell construction integral (i.e. the chemical potential difference) 
+    Real maxwell_int = math_utils::GLQuad32([this, temp, pin](Real rho){return (this->t_.PressureFromDensityTemperature(rho, temp) - pin)/(rho*rho);}, rho_l, rho_h);
```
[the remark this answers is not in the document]
lroberts36: $\int_{\nu_L}^{\nu_H} (P - P_{L/H}) d\nu = \int_{\rho_H}^{\rho_L} (P(\rho) - P_{L/H})/\rho^2 d\rho$ since $\nu = 1/\rho$

jhp-lanl:
What do you mean by this?

> Within the coexistence region, we do not assume exact pressure equality between the two phases but allow for a difference in pressure so that the pressure is not constant across the coexistence region but it is still continuous across the phase boundaries.

First, I just want to make sure we're on the same page so forgive me if this is a bit pedantic. The spinodal region represents the set of EOS states that are thermodynamically unstable between the local minimum and maximum in the EOS where dP/dV is positive. By contrast, the binodal region represents the region where two phases will be in equilibrium with each other. The states between the binodal and spinodal are metastable and represent compressing or expanding a phase beyond the point where it _could_ transition to the other phase (could being the operative word since this is now a nucleation issue).

The Maxwell construction pressure is by definition the binodal pressure because it links the two phases _at equilibirum_, but this will necessarily be different from the spinodal states since those are basically the local extrema in the EOS. So there will be a jump up or down from the extremum to the Maxwell construction pressure.

lroberts36:
@jhp-lanl: 
> What do you mean by this?
In a standard Maxwell construction, the pressure is constant across the binodal region and both the liquid and gas phase are in pressure equilibrium with one another. A constant pressure across a density region is not ideal for doing PTE solves in hydro codes (in this case Riot). Therefore, it would be useful to have Maxwell like fixup for thermodynamically unstable EoSs (the specific EoS I have in mind right now is the Davis reactants EoS), but where the pressure varies across the phase coexistence region so root finds don't run into a region with $(\partial P / \partial \rho)_T = 0$. 

In a Maxwell construction, the Helmholtz free energy is given by $f = f(\nu_L, T) u + (1 - u) f(\nu_H, T)$ with $ u = (\nu - \nu_H)/(\nu_L - \nu_H)$ where $\nu_L(T)$ and $\nu_H(T)$ define the bounds of the binodal region. Here, we replace that with $f = f(\tilde \nu_L, T) g(\tilde u) + (1 - g(\tilde u)) f(\tilde \nu_H, T)$ where $g$ is an interpolation function with $g(0) = 0$ and $g(1) = 1$ and $ \tilde u = (\nu - \tilde \nu_H(T))/(\tilde \nu_L(T) - \tilde \nu_H(T))$. We can then ensure $(\partial P/ \partial \rho)_T > 0$ in the coexistence region and ensure pressure continuity at the the edges of the new coexistence region by choosing $\tilde \nu_L$ and $\tilde \nu_H$ correctly. This does mean that $P(\nu_L, T) \neq P(\nu_H, T)$, which is what I was referring to in the comment above. This is mostly a practical fixup, but this could be interpreted as introducing a surface tension between the two phases I think. With this model for the free energy, it is straightforward to right down thermodynamically consistent expressions for all required quantities in the coexistence region ($\epsilon, s, \mu$, etc.)

I agree that for general tabulated EoSs it may not be possible/wise to fix up thermodynamically unstable regions with a Maxwell construction (or something close), but this is an issue with those particular tables. Right now, I am mostly thinking of this for analytic EOSs.

Also, the (pseudo-)Maxwell construction will not be done on the fly. The idea is that the coexistence region will be calculated on initialization and tabulated. Then when the modified EOS is called there will be a check to see if the required point is inside or outside the coexistence region and if inside the EOS will be evaluated in terms of the tabulated quantities.

singularity-eos/eos/modifiers/tilted_maxwell_eos.hpp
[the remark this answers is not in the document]
jhp-lanl: Ah I missed that denominator. Are you protecting against divide by zero somewhere else?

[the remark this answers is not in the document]
lroberts36: Not explicitly right now, but that can be enforced by ensuring the limits of integration are both positive

jhp-lanl:
> the specific EoS I have in mind right now is the Davis reactants EoS)

Ah okay this makes a lot more sense then. There are just too many issues with tabular data. Most of our analytic EOS are monotonic but I forgot about the Davis reactants EOS.

> In a Maxwell construction...

Github quoting seems to not want to use the original markdown so the whole quote doesn't render correctly.

Anyway, I was not arguing about how or whether a Maxwell construction works. Your PR description specifically says you are replacing the **_spinodal_** region without mentioning the **_binodal_** region. Since your implementation isn't complete yet, I couldn't figure out whether you were actually going to only substitute the **_spinodal_** region as it appeared you indicated or whether you were going to substitute the entire **_binodal_** region.

I hadn't heard of anybody proposing to just replace the spinodal region, but it's an intriguing idea since you confine the regions of the EOS where $(\partial P/\partial \rho)_S < 0$ to basically two discontinuities in the pressure and free energy without sacrificing the metastable states. Based off of your response though, I think I can now assume that you are actually intending to replace the entire binodal region in addition to just the spinodal region.

> Also, the (pseudo-)Maxwell construction will not be done on the fly. The idea is that the coexistence region will be calculated on initialization and tabulated. Then when the modified EOS is called there will be a check to see if the required point is inside or outside the coexistence region and if inside, the EOS will be evaluated in terms of the tabulated quantities.

Ah okay and this hasn't been implemented yet then? Can you add to the PR description more about where this work is headed (a checklist is very useful) so that those of us without prior knowledge of what you're working on can understand the context better? I'd like to see how you intend to interpolate on the tabulated VLE data since at least P(T) obeys a specific relationship along a phase change.

lroberts36:
> Anyway, I was not arguing about how or whether a Maxwell construction works. Your PR description specifically says you are replacing the **_spinodal_** region without mentioning the **_binodal_** region. Since your implementation isn't complete yet, I couldn't figure out whether you were actually going to only substitute the **_spinodal_** region as it appeared you indicated or whether you were going to substitute the entire **_binodal_** region.

I think some of the confusion stems from the fact I use *coexistence* region, which is mentioned in the PR description, rather than the synonymous *binodal* region.

> Ah okay and this hasn't been implemented yet then? 

Yes, that part hasn't been implemented yet (see the discussion above with Jonah). 

> Can you add to the PR description more about where this work is headed (a checklist is very useful) so that those of us without prior knowledge of what you're working on can understand the context better? 

Once I have things working and figure out if this is a useful path to follow, I will put up a more detailed description of what the modifier is supposed to do. I originally put up the WIP PR mainly to ask Jonah a couple of questions about the implementation. 

> I'd like to see how you intend to interpolate on the tabulated VLE data since at least P(T) obeys a specific relationship along a phase change.

At least initially I was going to try interpolation with cubic Hermite polynomials for $\tilde \nu_L(T)$, $\tilde \nu_H(T)$ and $P_D(T)$ (where $P(\rho, T) = g'(u) P_D(T)$ inside the fixed-up region) since we require continuous derivatives of the boundary specific volumes. I think the trick will be choosing how densely to sample these points. Note that for the construction above the transition is spread out in P-T space, so for instance the Clausius-Clapeyron relation does not hold.

singularity-eos/eos/modifiers/tilted_maxwell_eos.hpp
```diff
@@ -0,0 +1,314 @@
...
+    const auto [f, rho_maxwell_low, rho_maxwell_hi] = TestMaxwellConditions(p_maxwell, temp, rho_p_min, rho_p_max); 
+
+    return std::make_tuple(rho_maxwell_low, rho_maxwell_hi, p_maxwell, true);
+  }
+
+  auto GetTiltedMaxwellConstruction(Real temp) const {
+    const auto [rho_maxwell_low, rho_maxwell_hi, p_maxwell, has_spinodal] = this->GetMaxwellConstruction(temp);
+    if (!has_spinodal) return std::make_tuple(0.0, 0.0, 0.0, false); 
+    
+    auto func = [this, temp, p_maxwell](std::array<Real, 2> x){
```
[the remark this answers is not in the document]
Yurlungur: I recall having issues with this in C++11, but I think in C++14 it works. Also the function this lives in is host-only, so it's fine.

Yurlungur:
Going to assume this PR can wait for further progress before continuing this discussion?

singularity-eos/eos/modifiers/tilted_maxwell_eos.hpp
dholladay00: That timeline is also subject to change, my recommendation would be to not rely on a move to c++17 at this time. Would `std::tie` work for your purposes?

lroberts36: Ok, thanks. I think `std::tie` would certainly work here (or just returning via pointers in the function call). I was also using `constexpr`, but that should be easy to get around as well.

singularity-eos/eos/modifiers/tilted_maxwell_eos.hpp
```diff
@@ -0,0 +1,314 @@
...
+  using EosBase<TiltedMaxwellEOS<T>>::FillEos;
+
+  using BaseType = T;
+
+  // give me std::format or fmt::format...
+  static std::string EosType() {
+    return std::string("TiltedMaxwellEOS<") + T::EosType() + std::string(">");
+  }
+
+  static std::string EosPyType() { return std::string("Shifted") + T::EosPyType(); }
```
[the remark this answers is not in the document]
lroberts36: Yes, that is a copy and paste error I think.

lroberts36:
@Yurlungur et al.: Yeah, sorry that this got shelved for now. I initially started this because we thought some of the issues we were seeing in RIOT were related to thermodynamically unstable portions of the EOS, but it turned out to be a different issue. I would like to get it finished at some point though because this could still become a problem in the future.

Yurlungur:
I'm closing this for now. @lroberts36 let me know if you want to re-open at some point.

lroberts36:
@Yurlungur: Yeah, someday...

Yurlungur:
Some investigations in riot suggest this may be worth revisiting.

PR #202, opened 2022-12-06
Vinet EOS implemented

aematts:
## PR Summary
The coding is done in eos_vinet.hpp and it is verified to get the correct numbers out.
Unit test cases are in test_eos_vinet.cpp, testing all Vinet related functions and some more. A little more work is needed here to finalize the unit tests.
No documentation is provided yet.

## PR Checklist

- [x ] Adds a test for any bugs fixed. Adds tests for new features.
- [x ] Format your changes by using the `make format` command after configuring with `cmake`.
- [x ] Document any new features, update documentation for changes made.
- [ ] Make sure the copyright notice on any files you modified is up to date.
- [ ] After creating a pull request, note it in the CHANGELOG.md file

PR #215, opened 2023-01-10
fix epsilon

Yurlungur:
## PR Summary

In PR #202 @aematts noticed that `robust_utils.hpp` defines `EPS` and `constants` defines `EPS` and these two definitions can conflict. Here I remove the version in `constants.hpp`. I also noticed that tests are failing as of yesterday, I think due to some change in the docker instance used for github actions. This PR includes an attempted fix for that as well.

## PR Checklist

### DIFF
--- a/singularity-eos/eos/eos_gruneisen.hpp
+++ b/singularity-eos/eos/eos_gruneisen.hpp
@@ -175,6 +175,8 @@ PORTABLE_INLINE_FUNCTION Real Gruneisen::ComputeRhoMax(const Real s1, const Real
   };
 
   // Comparisons to zero are actually made to
+  // TODO(JMM): Should this become robust::EPS()? They differ by a
+  // factor of 10.
   static constexpr Real EPS = std::numeric_limits<Real>::epsilon();
 
   // First find the eta root. A negative root indicates that there is no maximum density.
--- a/singularity-eos/eos/eos_gruneisen.hpp
+++ b/singularity-eos/eos/eos_gruneisen.hpp
@@ -22,6 +22,7 @@
 
 #include <singularity-eos/base/constants.hpp>
 #include <singularity-eos/base/math_utils.hpp>
+#include <singularity-eos/base/robust_utils.hpp>
 #include <singularity-eos/base/root-finding-1d/root_finding.hpp>
 #include <singularity-eos/eos/eos_base.hpp>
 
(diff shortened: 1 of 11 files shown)


singularity-eos/eos/eos_gruneisen.hpp
```diff
@@ -175,6 +175,8 @@ PORTABLE_INLINE_FUNCTION Real Gruneisen::ComputeRhoMax(const Real s1, const Real
   };
 
   // Comparisons to zero are actually made to
+  // TODO(JMM): Should this become robust::EPS()? They differ by a
+  // factor of 10.
   static constexpr Real EPS = std::numeric_limits<Real>::epsilon();
```
jhp-lanl, line 180: I don't think the factor of 10 will make a difference. It looks like I used the EPS parameter just for zero-valued polynomial coefficients. Since the behavior of the polynomial will be very similar for a non-zero coefficient if that coefficient is very small, I think the analysis is still valid.

jhp-lanl (approved):
Looks good to me. Feel free to change the Gruneisen code or not

singularity-eos/eos/eos_gruneisen.hpp
Yurlungur: :+1: ok. It's simpler to use robust::eps so I'll try and make that change and see if tests till pass.

aematts (approved):
I see nothing in there that needs changing.

dholladay00:
What is the plan for updating all EOS implementations to use `robust::ratio` @Yurlungur ? Should it be done in this PR, create an issue? I'm afraid if we do it ad hoc it may not get accomplished.

chadmeyer (approved):

Yurlungur:
> What is the plan for updating all EOS implementations to use `robust::ratio` @Yurlungur ? Should it be done in this PR, create an issue? I'm afraid if we do it ad hoc it may not get accomplished.

A valid concern. I tried to do it wherever I saw an issue in this PR. But I may not have caught them all.

Yurlungur:
> What is the plan for updating all EOS implementations to use `robust::ratio` @Yurlungur ? Should it be done in this PR, create an issue? I'm afraid if we do it ad hoc it may not get accomplished.

I think it's fine to just do it ad hoc and commit to using `robust::ratio` going forward. I grepped for `EPS` in the code to find the spots where we had `a / (b + EPS)` and cleaned those spots up.

dholladay00:
> > What is the plan for updating all EOS implementations to use `robust::ratio` @Yurlungur ? Should it be done in this PR, create an issue? I'm afraid if we do it ad hoc it may not get accomplished.
> 
> I think it's fine to just do it ad hoc and commit to using `robust::ratio` going forward. I grepped for `EPS` in the code to find the spots where we had `a / (b + EPS)` and cleaned those spots up.

I guess the underlying question is this: do we want to replace every division in every implementation with a divide? If so, I don't think adding them ad hoc is sufficient, it should at least be documented in an issue as a goal if it in fact is a goal.

Yurlungur:
Ah I see. I will add something to the docs and create an issue.

PR #221, opened 2023-01-20
Draft: Up-to-date with upstream

mauneyc-LANL:
## PR Summary

Collect updates from spiner and ports-of-call.

The `FetchContent` code for `spiner` was slightly modified and used to replace the `singularity_import_*` code; this should make dependencies handle more consistently across codes. 

@jhp-lanl your fixes in #220 will be overwritten, but I still need to clear up some build issues on my end, so that isn't going to interfere.

2023-02-03: depends on https://github.com/lanl/spiner/pull/58

## One Must Imagine Sisyphus Happy

- [x] ~~`spiner` builds tests even with `BUILD_TESTING=OFF` cache set. why?~~ a3405a4
- [ ] with `kokkos+kokkoskernels` (openmp backend) segfaults on `pte` test
- [ ] some CMake lists have duplicate entries - these shouldn't break anything but why take the chance
- [ ] Check if "all-`Fetch`" build works (no spack, except hdf5)
- [ ] Check some mixed builds
- [ ] GPU build works
- [ ] make sure github/gitlab CI is up2date

## PR Checklist

- [ ] Adds a test for any bugs fixed. Adds tests for new features.
- [ ] Format your changes by using the `make format` command after configuring with `cmake`.
- [ ] Document any new features, update documentation for changes made.
- [ ] Make sure the copyright notice on any files you modified is up to date.
- [ ] After creating a pull request, note it in the CHANGELOG.md file
- [ ] If preparing for a new release, update the version in cmake.

PR #223, opened 2023-01-23
Update ports-of-call and introduce error handling

Yurlungur:
## PR Summary

This pulls my error handling machinery through `singularity-eos`, mostly motivated by @aematt's request for testing. I make a few changes of note:
1. I make `ports-of-call` its own submodule. This is unfortunately required for now. Once @mauneyc-LANL 's PR #221 is through, this won't be necessary and the submodule should be removed. Note that I **do not** update `spiner`. I think @mauneyc-LANL 's changes to `spiner` make it incompatible with the submodule path for now. So spiner stays on the old commit until #221 goes through.
2. I create a new macro, `REQUIRE_MAYBE_THROWS` which is to be used in units tests with catch. It either resolves to a `REQUIRE_THROWS` if `PORTABILITY_STRATEGY_NONE` or a no-op. This is what @aematts should use.
3. As a temp hack, I replace `EOS_ERROR` with `PORTABLE_ABORT`. We should move to `PORTABLE_*` everywhere with time. This is just for backwards compatibility.

## PR Checklist

### DIFF
--- a/spack-repo/packages/singularity-eos/package.py
+++ b/spack-repo/packages/singularity-eos/package.py
@@ -63,7 +63,7 @@ class SingularityEos(CMakePackage, CudaPackage):
 
     depends_on("eospac", when="+eospac")
     depends_on("spiner")
-    depends_on("ports-of-call@main")
+    depends_on("ports-of-call@1.4.0")
     depends_on("spiner +kokkos", when="+kokkos")
 
     depends_on("mpark-variant")
--- a/spack-repo/packages/singularity-eos/package.py
+++ b/spack-repo/packages/singularity-eos/package.py
@@ -63,6 +63,7 @@ class SingularityEos(CMakePackage, CudaPackage):
 
     depends_on("eospac", when="+eospac")
     depends_on("spiner")
+    depends_on("ports-of-call@main")
     depends_on("spiner +kokkos", when="+kokkos")
 
     depends_on("mpark-variant")
--- a/spack-repo/packages/singularity-eos/package.py
+++ b/spack-repo/packages/singularity-eos/package.py
@@ -63,7 +63,7 @@ class SingularityEos(CMakePackage, CudaPackage):
 
     depends_on("eospac", when="+eospac")
     depends_on("spiner")
-    depends_on("ports-of-call@1.4.0:")
+    depends_on("ports-of-call@1.4.1:")
     depends_on("spiner +kokkos", when="+kokkos")
 
     depends_on("mpark-variant")
--- a/spack-repo/packages/singularity-eos/package.py
+++ b/spack-repo/packages/singularity-eos/package.py
@@ -63,7 +63,7 @@ class SingularityEos(CMakePackage, CudaPackage):
 
     depends_on("eospac", when="+eospac")
     depends_on("spiner")
-    depends_on("ports-of-call@1.4.0")
+    depends_on("ports-of-call@1.4.0:")
     depends_on("spiner +kokkos", when="+kokkos")
 
     depends_on("mpark-variant")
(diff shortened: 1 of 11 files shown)


Yurlungur:
@aematts once this is merged in the way to get your tests to work is to use `PORTABLE_ALWAYS_THROW_OR_ABORT` in the places you want caught, then use `PORTABLE_REQUIRE_MAYBE_THROWS` in the test to catch it.

jhp-lanl (approved):
Looks pretty straight-forward. I approve

Yurlungur:
Ok tests now pass on github and re-git. To make tests pass, I had to change the spack spec to point to the main branch of `ports-of-call`. This is temporary and can be removed when `ports-of-call` has a new updated spackage in spack upstream for the [newest release](https://github.com/lanl/ports-of-call/releases/tag/v1.4.0).

Yurlungur:
@rbberger please take a look at the spack trickery I added here and let me know whether you're ok with it

aematts (approved):
Looks fine to me even though I am not sure how to use it yet. Once I start using it I might come with more comments.

spack-repo/packages/singularity-eos/package.py
```diff
@@ -63,6 +63,7 @@ class SingularityEos(CMakePackage, CudaPackage):
 
     depends_on("eospac", when="+eospac")
     depends_on("spiner")
+    depends_on("ports-of-call@1.4.0")
```
rbberger, line 66: Should be
```suggestion
    depends_on("ports-of-call@1.4.0:")
```
You also want the same in `spiner`.

Yurlungur: Spiner isn't using the new code. I'll add the colon in singularity.

rbberger (approved):
Approving this, assuming the re-git CI passes.

test/test_eos_unit.cpp
```diff
@@ -131,6 +132,13 @@ inline void compare_two_eoss(E1 &&test_e, E2 &&ref_e) {
   return;
 }
 
+SCENARIO("Test that we can either throw an error on host or do nothing on device",
+         "[RequireMaybe]") {
+  // TODO(JMM): For whatever reason, the preprocessor does not like it if I
+  // call `PORTABLE_ALWAYS_THROW_OR_ABORT
+  REQUIRE_MAYBE_THROWS(PORTABLE_ALWAYS_THROW_OR_ABORT("Error message"));
```
dholladay00, line 139: Is catch ok when this scenario is empty as is the case if portability strategy none is not defined @Yurlungur ?

Yurlungur: Catch is ok with empty scenarios I think. The scenario just auto-passes with zero conditions I believe.

dholladay00: Is the empty scenario scenario tested in our CI @Yurlungur ?

Yurlungur: Yes if the CI passes on GPU it passed for the empty scenario.

Yurlungur:
Tests pass on re-git. Pulling the trigger.

aematts:
> @aematts once this is merged in the way to get your tests to work is to use `PORTABLE_ALWAYS_THROW_OR_ABORT` in the places you want caught, then use `REQUIRE_MAYBE_THROWS` in the test to catch it.

@Yurlungur , did you get something in for the necessary warning when T<0? You did not like to have a plain write statement so deep in the code.

Yurlungur:
@aematts ah yes I did. You can use `PORTABLE_WARN`, which is a no-op by default, but when compiled in debug mode, it will turn into an actual warning.

PR #240, opened 2023-03-17
WIP: fix warnings

Yurlungur:
## PR Summary

I'm trying to chase down some bug in the stellar collapse table, but while I'm at it I figured I'd clean up the warnings as they've started to get extensive. 

## PR Checklist

- [ ] Adds a test for any bugs fixed. Adds tests for new features.
- [ ] Format your changes by using the `make format` command after configuring with `cmake`.
- [ ] Document any new features, update documentation for changes made.
- [ ] Make sure the copyright notice on any files you modified is up to date.
- [ ] After creating a pull request, note it in the CHANGELOG.md file
- [ ] If preparing for a new release, update the version in cmake.

PR #243, opened 2023-03-21
Limit undefined behaviour and fix some wharnings too

Yurlungur:
## PR Summary

In a downstream code, @AstroBarker experienced a cuda "unspecified launch failure" for some applications and on some architectures. After some effort, we narrowed it down to the `RootFinding1D::RootCounts` and `RootFinding1D::RootStatus` variables used in the tabulated EOS's. Each EOS had an internal instants of `RootCounts` and `Status` used for diagnostics when the code was run on host in serial. 

Writing to these variables in multiple threads is undefined behavior. However,  I intentionally put the logic in the code anyway. My reasoning was that since they wouldn't be *read* except for diagnostic purposes in serial, this undefined behavior shouldn't cause any problems. This reasoning was apparently incorrect.

I still think these variables are useful, so I don't want to eliminate this capability completely. However, I want to avoid them being written to in a "device" context (whatever that means going forward), to prevent us from hitting this fail case in the future. So I have made the following changes:
- The root counts object is now an *optional* parameter of the root finding routines. You pass in a pointer, if that pointer isn't null, the histogram is recorded.
- The root counts and root status fields owned by each tabulated EOS object are only written to if the EOS believes itself to be "on host."

Note that this strategy doesn't prevent someone from sticking a host EOS in an `openmp for` loop, so this doesn't entirely escape the issue, but it seems to be a safe enough workaround. A more complete solution would be to hide the histogram/status logic behind a compile-time flag or remove it entirely. I'd be open to that if people feel strongly, but I think this is probably good enough for now.

@AstroBarker please try this branch and see if it resolves the issues you were seeing.

@dholladay00 @mauneyc-LANL @rbberger @jhp-lanl  I'd like your feedback on this solution. Do you agree this feels safe enough for now? Or should we more aggressively eliminate any possibility of UB here?

While I'm at it, I cleaned up some warnings I was seeing when compiling the code which were getting on my nerves:
- A few unused variable warnings
- I re-implemented `strcat` with device decorations (needed for the error message for entropy not implemented), as the standard library version is not implemented on device
- I made some device/inline annotations consistent in a few spots.

There are more warnings in the code than this... and I'd like to clean them up when we have the time. But one thing at a time.

## PR Checklist

### DIFF
--- a/singularity-eos/base/root-finding-1d/root_finding.hpp
+++ b/singularity-eos/base/root-finding-1d/root_finding.hpp
@@ -148,7 +148,8 @@ template <typename T>
 PORTABLE_INLINE_FUNCTION Status regula_falsi(const T &f, const Real ytarget,
                                              const Real guess, Real a, Real b,
                                              const Real xtol, const Real ytol,
-                                             Real &xroot, const RootCounts &counts) {
+                                             Real &xroot,
+                                             const RootCounts *counts=nullptr) {
   constexpr int max_iter = SECANT_NITER_MAX;
   auto func = [&](const Real x) { return f(x) - ytarget; };
   Real ya = func(a);
@@ -215,10 +216,12 @@ PORTABLE_INLINE_FUNCTION Status regula_falsi(const T &f, const Real ytarget,
         "root finding reached the maximum number of iterations.  likely not converged\n");
     status = Status::FAIL;
   }
-  if (iteration_count < counts.nBins()) {
-    counts.increment(iteration_count);
-  } else {
-    counts.increment(counts.more());
+  if (counts != nullptr) {
+    if (iteration_count < counts->nBins()) {
+      counts->increment(iteration_count);
+    } else {
+      counts->increment(counts->more());
+    }
   }
   xroot = 0.5 * (a + b);
   return status;
@@ -228,7 +231,7 @@ template <typename T>
 PORTABLE_INLINE_FUNCTION Status findRoot(const T &f, const Real ytarget, Real xguess,
                                          const Real xmin, const Real xmax,
                                          const Real xtol, const Real ytol, Real &xroot,
-                                         const RootCounts &counts) {
+                                         const RootCounts *counts=nullptr) {
   Status status;
 
   // first check if we're at the max or min values
@@ -236,14 +239,18 @@ PORTABLE_INLINE_FUNCTION Status findRoot(const T &f, const Real ytarget, Real xg
   const Real errmax = fabs(fmax - ytarget) / (fabs(fmax) + ytol);
   if (errmax < ytol) {
     xroot = xmax;
-    counts.increment(0);
+    if (counts != nullptr) {
+      counts->increment(0);
+    }
     return Status::SUCCESS;
   }
   const Real fmin = f(xmin);
   const Real errmin = fabs(fmin - ytarget) / (fabs(fmin) + ytol);
   if (errmin < ytol) {
     xroot = xmin;
-    counts.increment(0);
+    if (counts != nullptr) {
+      counts->increment(0);
+    }
     return Status::SUCCESS;
   }
 
@@ -289,7 +296,7 @@ template <typename T>
 PORTABLE_INLINE_FUNCTION Status secant(const T &f, const Real ytarget, const Real xguess,
                                        const Real xmin, const Real xmax, const Real xtol,
                                        const Real ytol, Real &xroot,
-                                       const RootCounts &counts) {
+                                       const RootCounts *counts=nullptr) {
   Real dx;
   Real x_last, y, yp, ym, dyNum, dyDen, dy;
 
@@ -330,17 +337,21 @@ PORTABLE_INLINE_FUNCTION Status secant(const T &f, const Real ytarget, const Rea
               xguess, ytarget, x, x_last, xmin, xmax, y, dx, yp, ym, dyNum, dyDen, dy,
               iter, (int)SINGULARITY_MY_SIGN(x));
 #endif
-      counts.increment(counts.more());
+      if (counts != nullptr) {
+        counts->increment(counts->more());
+      }
       return Status::FAIL;
     }
     if (fabs(x - x_last) / (fabs(x) + xtol) < xtol) break;
   }
   // increment iter
   ++iter;
-  if (iter < counts.nBins()) {
-    counts.increment(iter);
-  } else {
-    counts.increment(counts.more());
+  if (counts != nullptr) {
+    if (iter < counts->nBins()) {
+      counts->increment(iter);
+    } else {
+      counts->increment(counts->more());
+    }
   }
   xroot = x;
 
@@ -388,7 +399,7 @@ template <typename T>
 PORTABLE_INLINE_FUNCTION Status bisect(const T &f, const Real ytarget, const Real xguess,
                                        const Real xmin, const Real xmax, const Real xtol,
                                        const Real ytol, Real &xroot,
-                                       const RootCounts &counts) {
+                                       const RootCounts *counts=nullptr) {
   Real xl, xr, fl, fr, dx;
 
   Real grow = 0.01;
--- a/singularity-eos/base/root-finding-1d/root_finding.hpp
+++ b/singularity-eos/base/root-finding-1d/root_finding.hpp
@@ -149,7 +149,7 @@ PORTABLE_INLINE_FUNCTION Status regula_falsi(const T &f, const Real ytarget,
                                              const Real guess, Real a, Real b,
                                              const Real xtol, const Real ytol,
                                              Real &xroot,
-                                             const RootCounts *counts=nullptr) {
+                                             const RootCounts *counts = nullptr) {
   constexpr int max_iter = SECANT_NITER_MAX;
   auto func = [&](const Real x) { return f(x) - ytarget; };
   Real ya = func(a);
@@ -231,7 +231,7 @@ template <typename T>
 PORTABLE_INLINE_FUNCTION Status findRoot(const T &f, const Real ytarget, Real xguess,
                                          const Real xmin, const Real xmax,
                                          const Real xtol, const Real ytol, Real &xroot,
-                                         const RootCounts *counts=nullptr) {
+                                         const RootCounts *counts = nullptr) {
   Status status;
 
   // first check if we're at the max or min values
@@ -296,7 +296,7 @@ template <typename T>
 PORTABLE_INLINE_FUNCTION Status secant(const T &f, const Real ytarget, const Real xguess,
                                        const Real xmin, const Real xmax, const Real xtol,
                                        const Real ytol, Real &xroot,
-                                       const RootCounts *counts=nullptr) {
+                                       const RootCounts *counts = nullptr) {
   Real dx;
   Real x_last, y, yp, ym, dyNum, dyDen, dy;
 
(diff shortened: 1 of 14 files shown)


Yurlungur:
This branch is the same one as for the closed PR #240 . I re-used the branch.

AstroBarker:
Thanks for this @jonahm-LANL -- this does solve my downstream problems. The `pcounts` implementation looks solid to me.

singularity-eos/eos/eos_base.hpp
```diff
@@ -391,8 +414,8 @@ class EosBase {
   void EntropyIsNotEnabled(const char *eosname) const {
     // Construct the error message using char* so it works on device
     char msg[120] = "Entropy is not enabled for the '";
```
dholladay00, line 416: I know this isn't part of this PR, but we should really consider shrinking this to as small as we can make it.

Yurlungur: Let's see some rough estimates here... the first half of the message is 32 chars. The longest EOS name is `SpinerEOSDependsRhoSie` which is 22 characters long. Plus the null terminating character. So currently this message could be as small as 55 characters at the moment. But that doesn't include future potential EOS's. 

I'm not comfortable being too too close to the margin there. So maybe we could do something like 81 chars?

Yurlungur: I forgot the 5 chars for the end of the message. Anyway I updated it here. Message length is now 81 chars, which is still 22 characters longer than it needs to be currently, but I think this is a reasonable margin for error for now.

test/profile_eos.cpp
```diff
@@ -222,8 +222,8 @@ int main(int argc, char *argv[]) {
 
     EOSBuilder::EOSType type = EOSBuilder::EOSType::IdealGas;
     EOSBuilder::params_t params;
-    params["Cv"] = 1e7 * 0.716; // specific heat in ergs/(g K)
-    params["gm1"] = 0.4;        // gamma - 1
+    params["Cv"].emplace<Real>(1e7 * 0.716); // specific heat in ergs/(g K)
+    params["gm1"].emplace<Real>(0.4);        // gamma - 1
```
Yurlungur, line 226: This was actually an error having to do with variant.

singularity-eos/eos/eos_base.hpp
```diff
@@ -34,6 +34,29 @@ static inline auto member_func_name(const char *type_name, const char *func_name
...
+namespace impl {
+// Cuda doesn't have strcat, so we implement it ourselves
+PORTABLE_FORCEINLINE_FUNCTION
+char *StrCat(char *destination, const char *source) {
+  int i, j;
+
+  // move to the end of the destination string
+  // specifically avoid strlen, which isn't on GPU
+  for (i = 0; destination[i] != '\0'; i++)
+    ;
```
Yurlungur, line 46: `clang-format` what are you doing? Should I disable clang-format here or add an empty `{}` to short-circuit this weird newline thing?

AstroBarker:
Another thought @Yurlungur, this PR might be a good spot to slip in the sieMin_ etc in GetOnDevice() for stellar_collapse?

Yurlungur:
> Another thought @Yurlungur, this PR might be a good spot to slip in the sieMin_ etc in GetOnDevice() for stellar_collapse?

Done

singularity-eos/base/root-finding-1d/root_finding.hpp
```diff
@@ -228,22 +231,26 @@ template <typename T>
...
   Status status;
 
   // first check if we're at the max or min values
   const Real fmax = f(xmax);
   const Real errmax = fabs(fmax - ytarget) / (fabs(fmax) + ytol);
   if (errmax < ytol) {
     xroot = xmax;
-    counts.increment(0);
+    if (counts != nullptr) {
+      counts->increment(0);
```
jhp-lanl, line 243: Why is `counts->increment()` called with `iteration_count` in `regula_falsi()` but with the `0` literal here?

singularity-eos/eos/eos_base.hpp
```diff
@@ -34,6 +34,29 @@ static inline auto member_func_name(const char *type_name, const char *func_name
...
+  for (j = 0; source[j] != '\0'; j++) {
+    destination[i + j] = source[j];
+  }
+
+  // null terminate destination string
+  destination[i + j] = '\0';
+
+  // the destination is returned by standard `strcat()`
+  return destination;
+}
```
jhp-lanl: Thank you for doing this!

Since we're already rolling our own `StrCat()`, would it be reasonable to add explicit bound arguments for the char arrays and perform length checking or at least truncation?

test/profile_eos.cpp
jhp-lanl, line 226: C++ neophyte question: what is the advantage of emplace over assignment?

test/test_eos_unit.cpp
```diff
@@ -475,9 +471,8 @@ SCENARIO("Ideal gas entropy", "[IdealGas][Entropy]") {
     constexpr Real EntropyT0 = 100;
     constexpr Real EntropyRho0 = 1e-03;
     EOS host_eos = IdealGas(gm1, Cv, EntropyT0, EntropyRho0);
-    EOS eos = host_eos.GetOnDevice();
```
jhp-lanl, line 478: Why is there an issue with this ideal gas entropy test but not the vector ideal gas tests below? I would advocate that we fix this test on device rather than preventing it from running on device.

jhp-lanl (changes requested):
This looks good, but I'm concerned about removing device testing of the ideal gas entropy test. To me it doesn't seem necessary since we test the vector version on device anyway.

singularity-eos/eos/eos_base.hpp
jhp-lanl: I don't mind this too much since it very clearly identifies this as a `for` loop with no function other than to increment. The empty braces would also communicate this, so that's fine with me as well.

test/profile_eos.cpp
jhp-lanl: Resolving since Jonah commented on this above and I didn't notice.

Yurlungur: Assignment requires that C++ can unambiguously cast the type into the types allowed in the variant. Type inferrence was failing at that cast. `emplace<Real>` lets me explicitly tell the compiler "I want this to be stored as a `Real` in the variant."

singularity-eos/eos/eos_base.hpp
Yurlungur: It should be pretty easy to add a check to `StrCat`. I'll do so.

singularity-eos/base/root-finding-1d/root_finding.hpp
Yurlungur: That line is inside a check for "we don't need to do any iterations because the guess is already correct." That's why the `if (errmax < ytol)`. It's before any iterations have been performed.

singularity-eos/eos/eos_base.hpp
Yurlungur: Yes I think this is a good idea. I'll add it.

test/test_eos_unit.cpp
Yurlungur: There's nothing wrong with this test, but it wasn't being run on device. The call was being made on the scalar quantities on a "device" object in a host execution context. It only worked because ideal gas is the same object on host and device. But it's technically "paradigm" breaking. One should either put the call in a `par_for` or remove the `GetOnDevice()`. I chose the former since it was a smaller change and the vector calls below test the device context. We can go the other route if you'd prefer though.

singularity-eos/eos/eos_base.hpp
Yurlungur: @jhp-lanl @dholladay00 please take a look at `StrCat` now. I make the maximum string length a global constant (hidden behind an implementation namespace) and call `PORTABLE_REQUIRE` inside `StrCat`. It'll only yell at you when you compile win "debug mode" but I think that's fine.

test/profile_eos.cpp
jhp-lanl: Thank you! That makes sense

test/test_eos_unit.cpp
jhp-lanl: Ah thank you. That makes sense. I'm glad now that I included the entropy in the vector test then.

singularity-eos/eos/eos_base.hpp
```diff
@@ -34,6 +34,31 @@ static inline auto member_func_name(const char *type_name, const char *func_name
...
+  int i, j; // not in loops because they're re-used.
+
+  // specifically avoid strlen, which isn't on GPU
+  for (i = 0; destination[i] != '\0'; i++) {
+  }
+  // assumes destination has enough memory allocated
+  for (j = 0; source[j] != '\0'; j++) {
+    // MAX_NUM_CHARS-1 to leave room for null terminator
+    PORTABLE_REQUIRE((i + j) < MAX_NUM_CHARS - 1,
+                     "Concat string must be within allowed size");
```
jhp-lanl, line 51: This will work. Thanks for adding it!

singularity-eos/eos/eos_base.hpp
jhp-lanl: I approve

Yurlungur:
Sounds good. I'll wait to hear from @dholladay00

dholladay00:
@Yurlungur could you grep through the code and remove instances of root counts where they were used only b/c the API required it? I know where is one in `singularity_eos.cpp` that can be removed.

mauneyc-LANL:
(none of this implies a blocking review)
I'm a bit late to this so apologies if this has been covered. Would the better solution be to use atomic writes? This may require some "portable" atomic routines in PoC, tho.

Keeping the diagnostic would be useful even on device. Perhaps a better approach is to template on an interface

```c++
struct DiagStub
{
// has interface as RootCounts but no-op implementation, e.g. DiagStub::reset() {}
};

// solves for f(x,params) - ytarget = 0
template <typename T, typename DiagCount = DiagStub>
PORTABLE_INLINE_FUNCTION Status regula_falsi(const T &f, const Real ytarget,
                                             const Real guess, Real a, Real b,
                                             const Real xtol, const Real ytol,
                                             Real &xroot,
                                             DiagCounts &counts) {
```

For debug or evaluation builds, this can be changed at build time. You also get the benefit of reducing branch points in the function. (This is a "for instance" snippet, go easy on unintended effects).

Tangentially, there is the question of rolling a custom root-find instead of relying on a more general library, though I'm not sure that this is a "now" concern.

Yurlungur:
> @Yurlungur could you grep through the code and remove instances of root counts where they were used only b/c the API required it? I know where is one in `singularity_eos.cpp` that can be removed.

The one in `singularity_eos.cpp` was the only one I missed. I'll fix it. :)

In reply to @mauneyc-LANL : 

> Would the better solution be to use atomic writes? This may require some "portable" atomic routines in PoC, tho.

I considered atomics, but I didn't want to deal with the performance implication. Perhaps a good long-term solution is to add atomics to POC and have the atomics be behind a compile flag.

> Keeping the diagnostic would be useful even on device. Perhaps a better approach is to template on an interface

Interesting. I kind of like this idea, but I also kind of like root-finding being optional as an argument, since there were a lot of places in the code where counts were used but not examined.

How about we keep what I've done here for now but add an issue suggesting both atomics and a mixin here for the count type. Would that work?

dholladay00:
Could even do this (add default argument)
```c++
struct DiagStub
{
// has interface as RootCounts but no-op implementation, e.g. DiagStub::reset() {}
};

// solves for f(x,params) - ytarget = 0
template <typename T, typename DiagCount = DiagStub>
PORTABLE_INLINE_FUNCTION Status regula_falsi(const T &f, const Real ytarget,
                                             const Real guess, Real a, Real b,
                                             const Real xtol, const Real ytol,
                                             Real &xroot,
                                             DiagCounts &counts = DiagStub{}) {
```

jhp-lanl:
@Yurlungur @dholladay00 @mauneyc-LANL given that the CI passed on Darwin and that we seemed to resolve any open questions at the meeting today, I'm pulling the trigger and merging this.

PR #246, opened 2023-03-29
CMake updates; flush with upstream

mauneyc-LANL:
## PR Summary
Update 20230727:

CI is passing on all systems. I did need to update the CI yamls to point to `main` branches of upstream `spiner`, so there needs to be some synchronization on versions, versions listed in spackages, spackage files in the repo folder, and versions in the CI to get everything to line up.

@jonahm-LANL @jhp-lanl @dholladay00 I've re-written the building docs with some expanded information. No doubt more can be put there, but what's there should cover everything. Reused some of the old documentation, as well, though a bit of it was out-of-date. If you feel there should be some more info there, let me know.

@rbberger I made some changes in the spackage file to reflect changes made elsewhere. I also put in some `TODO` comments on a few things that arn't required, but may be useful to look over. 

Update 20230627:

- [x] Requires https://github.com/lanl/spiner/pull/71 to be accepted before merge

Update 20230623:

Several changes have been done.

1. Moved to bifurcated "submodule mode/standalone mode" for building. This should make it easier to maintain. In submodule mode, install/export steps are disabled.
2. Options compacted using `cmake_dependent_option`, which simplifies the required cmake invokes. Further, a `CMakePresets.json` file has been mocked-up to provide common configuration options when building. (see about `README.build.md` for more details)
3. Updated the `ports-of-call` and `spiner` submodules to latest versions. _These should get proper releases before merging this PR so that the submodules can be correctly pinned_. As of writting I'm trying to mangle the submodules to correctly work on the github CI, if you spot how my bumbling can be improved, let me know.
4. Some code changes necessary for `spiner`-related code due to the templating of `Spiner::DataBox`. There may be a cleaner approach to this so I'm open to suggestions; however, the change was simple (if verbose).
5. Clarifying `spiner`/`hdf5` dependency in the code defines; I've replaced the checks on `SPINER_USE_HDF` with the more descriptive `SINGULARITY_USE_SPINER_WITH_HDF5` define.
6. I'm drafting a build doc `README.build.md` with instructions for building in various contexts. Feel free to provide suggestions.
7. some `package.py` changes to reflect changes in build setup, and CI adjustments

### `spiner`, HDF5 clarity

As of writting, there is still a lot of `HDF5` code in `eos_spiner`,`eos_stellarcollapse` that is distinct (at least on glancing) from `spiner`. I haven't disentangled it enough for us to have a possible "HDF5-less" form of `spiner` in `singularity-eos`. There may be some parallel work needed in `spiner` itself, but right now I'm focused on getting everything for a stable release. Right now it's possible to tell `singularity-eos` to build with spiner but without HDF5, but the effect is just to disable `eos_spiner`, so it's not ideal.

### Status

I'm currently assessing where things stand in terms of completeness. Most of what I wanted done is complete but there are some loose ends to tie up (finishing docs, making sure spack is in-line, cleaning up some hacks). I'm able to build and test in all environments expected to be run on, but I still need to demonstrate this on the CI.

==/Update 

This is an update to streamline the CMake build and bring in changes from upstream codes.

- removed generic dependency handling into `ad-hoc` scripts, now more readable and flexible for particular packages
- better HDF5 handling
- Code cleanup; some re-ordering and formatting is done along with other changes

## PR Checklist

- [ ] Ensure minimal disruption to downstream projects
- [ ] Sort out any hdf5+spiner issues
- [ ] Document + readable diagnostics
- [ ] Prune the cmake for out-dated, redundant, or unnecessary code
- [ ] Ensure `install` stage is correct with changes 


mauneyc-LANL:
@dholladay00 @Yurlungur This is ongoing, but hopefully what remains is straightforward.

@rbberger I haven't given much thought to any spack/CI changes with these. I anticipate it to be minimal, but will ping you if any major issues arise

CMakeLists.txt
```diff
@@ -12,7 +12,7 @@
 # publicly and display publicly, and to permit others to do so.
 #------------------------------------------------------------------------------#
 
-cmake_minimum_required(VERSION 3.14)
+cmake_minimum_required(VERSION 3.20)
```
Yurlungur, line 15: Do we need to up the cmake version number? Care should go into this, as I've noticed that some LANL machines are stuck on cmake < 3.20

CMakeLists.txt
```diff
@@ -45,38 +40,43 @@ option (SINGULARITY_BUILD_STELLARCOLLAPSE2SPINER "Compile stellarcollapse2spiner
...
+#option (SINGULARITY_PATCH_MPARK_VARIANT "Apply GPU patch to mpark-variant submodule" ON)
 
 # Patches variant to be compatible with cuda
 # Assumes "patch" is present on system
-if(SINGULARITY_PATCH_MPARK_VARIANT)
-  message(STATUS "Patching mpark::variant to support GPUs")
-  execute_process(COMMAND patch -N -s -V never
-                  ${CMAKE_CURRENT_SOURCE_DIR}/utils/variant/include/mpark/variant.hpp
-                  ${CMAKE_CURRENT_SOURCE_DIR}/utils/cuda_compatibility.patch
-                  )
```
Yurlungur, line 64: is this happening elsewhere now?

CMakeLists.txt
```diff
@@ -154,28 +154,38 @@ endif()
...
 
 #------------------------------------------------------------------------------#
 # Handle dependencies
 #------------------------------------------------------------------------------#
 
+find_package(mpark_variant REQUIRED)
+target_link_libraries(singularity-eos PUBLIC mpark_variant)
+
+find_package(ports-of-call REQUIRED)
+target_link_libraries(singularity-eos PUBLIC ports-of-call::ports-of-call)
```
Yurlungur, line 183: target guards?

cmake/singularity/Eigen3.cmake
```diff
@@ -0,0 +1,5 @@
+macro(singularityeos_enable_eigen3 target)
+  find_package(Eigen3 REQUIRED)
+  target_link_libraries(${target} PUBLIC Eigen3::Eigen)
+  set(SINGULARITY_EOS_ENABLE_MPARK-VARIANT ON)
+endmacro()
```
Yurlungur, line 5: I'm fine either way, but I'm curious: why a per-file macro here rather than just doing this stuff inline? Just easier from an organization/readability perspective?

cmake/singularity/hdf5.cmake
```diff
@@ -0,0 +1,28 @@
...
+  find_package(HDF5 COMPONENTS HL REQUIRED)
+
+
+  target_include_directories(${target} SYSTEM PUBLIC ${HDF5_INCLUDE_DIRS})
+  target_link_libraries(${target} PUBLIC ${HDF5_LIBRARIES})
+
+  if(HDF5_IS_PARALLEL)
+    find_package(MPI COMPONENTS CXX REQUIRED)
+    target_link_libraries(${target} PUBLIC MPI::MPI_CXX)
+  endif()
```
Yurlungur, line 22: I can imagine this being a little finicky: If a user has both parallel and serial MPI installed and wants to specify one or the other... is that on us?

cmake/singularity/kokkos.cmake
```diff
@@ -0,0 +1,87 @@
+macro(singularity_enable_kokkos target)
+
+  if(SINGULARITY_CUSTOM_KOKKOS)
+    message(WARNING "You have selected to use a custom source of Kokkos."
+    " This is not recommended, and developer support for this feature is not secured."
+    " To relieve potential issues in the future, you should reevaluate how your build system uses TPLs.")
```
Yurlungur, line 6: I think this last line of the warning is probably not necessary

cmake/singularity/Eigen3.cmake
```diff
@@ -0,0 +1,5 @@
+macro(singularityeos_enable_eigen3 target)
+  find_package(Eigen3 REQUIRED)
```
Yurlungur, line 2: should have a target guard

cmake/singularity/eospac.cmake
```diff
@@ -0,0 +1,10 @@
+macro(singularity_enable_eospac target)
+  find_package(EOSPAC REQUIRED)
```
Yurlungur, line 2: target guard

cmake/singularity/hdf5.cmake
```diff
@@ -0,0 +1,28 @@
+macro(singularity_enable_hdf5 target)
+  # CMake's FindHDF5 package ignores any path in CMAKE_PREFIX_PATH that
```
Yurlungur, line 2: HDF5 should probably have a target guard

cmake/singularity/mpark_variant.cmake
```diff
@@ -0,0 +1,6 @@
+macro(singularity_enable_mpark_variant target)
+  
+  find_package(mpark_variant REQUIRED)
```
Yurlungur, line 3: target guard. Also what about the patch?

Yurlungur (commented):
This all seems pretty reasonable. Some nitpicks below. Mainly want to make sure everything is behind a target guard, so that everything still works for the downstream codes doing submodules. Also when you're ready I'd like to test that build path and add some docs for it.

rbberger (approved):
Overall all looks good to me. We use a similar macro scheme for FleCSI, so I see it as a bonus that this seems to be converging into the same direction. As for the CMake version, even EAP requires 3.20, so I'd support that change.

mauneyc-LANL:
@Yurlungur @dholladay00 some proposed updates:

I am working on a setup with the following constraints:

```
# There are two modes for building:
# 1.) Submodule mode
#     this mode is for projects where singularity-eos,
#     and it's dependencies, are present in the source tree. 
#     This mode is only appropriate for projects that have been
#     designed around using git submodules for dependency
#     management.
#     Submodule mode disables all export/install steps.
# 2.) Standalone mode
#     this mode will build and install singularity-eos as
#     as a complete software package for the intended platform. 
#     In standalone mode, all dependencies are expected to be 
#     found with `find_package`, and an error will be produced 
#     if the packages required for building are not located outside
#     of the source directory, except for explicit cases.
```

Long story short: Dependencies (save for explicit cases) are either *all* managed in submodules or *all* resolved with `find_package()`. This means a few changes that could be debatable:

- There is no "mix-and-match" capabilities. This is (mostly) possible with the current CMake but: a.) it's almost never used in the "mixed" case and b.) it has made maintaining everything overly-complicated.
- `kokkos-kernels` is back as a submodule. This is in-line with the "all-one-or-the-other" approach.
- The "target-guards" on `find_package()` are no longer applicable - the build with either define a target, or find it will find a target.

There are other "in-flight" updates that deserve some attention, but I wanted to start discussion on this sooner rather than later. 

Keep in mind I still consider this "WIP", so there is still plenty to be done/decided.

jhp-lanl:
>     * There is no "mix-and-match" capabilities. This is (mostly) possible with the current CMake but: a.) it's almost never used in the "mixed" case and b.) it has made maintaining everything overly-complicated.

How does this impact building with EOSPAC?

cmake/singularity/kokkos.cmake
```diff
@@ -0,0 +1,112 @@
+macro(singularity_import_kokkos)
+  
+  set(Kokkos_ENABLE_SERIAL ON CACHE BOOL "" FORCE)
+  if(SINGULARITY_USE_CUDA)
+    set(Kokkos_ENABLE_CUDA ON CACHE BOOL "" FORCE)
+    set(Kokkos_ENABLE_CUDA_LAMBDA ON CACHE BOOL "" FORCE)
+    set(Kokkos_ENABLE_CUDA_RELOCATABLE_DEVICE_CODE ON CACHE BOOL "" FORCE)
```
dholladay00, line 7: We should definitely remove this. We no longer require rdc.

mauneyc-LANL: Clarification: remove _just_ `RELOCATABLE`, or remove all but `Kokkos_ENABLE_CUDA`, or remove all these cache options entirely?

mauneyc-LANL: done

Yurlungur:
Big picture I am supportive of a simpler single switch with no mix-and-match. A few questions/comments:

> kokkos-kernels is back as a submodule. This is in-line with the "all-one-or-the-other" approach.

Do you plan to add submodules back into the project? Or would a person using submodule mode have to add them to the parent project?

> The "target-guards" on find_package() are no longer applicable - the build with either define a target, or find it will find a target.

IMO target guards should still be on the find package targets that may be needed for submodule mode, e.g., hdf5 and eospac. I don't trust the cmake find package machinery to do this correctly on its own.

mauneyc-LANL:
> @mauneyc-LANL looking at just the HDF5 change, I'm a bit confused why you aren't using the `hdf5::hdf5` and `hdf5::hdf5_hl` targets. Also, why is the error logic necessary, if you have `REQUIRED` components. Does it fall through and not error out? Who can define `HDF5_LIBRARIES` before (why isn't it a check for the targets)?
> 
> Btw, I noticed both `SPINER_USE_HDF` and `SPINER_USE_HDF5` defines. Do the both exist? (And if so, should they?)

In reverse, I'm still working on the `SPINER_USE_` changes. There is a few source changes that need to be worked on - it's on the list.

For HDF5, using the `hdf5::` targets has been tried before, but there were issues downstream. This could've been

1. the `hdf5::` targets may not be defined downstream. HDF5 handeling has been through a lot of different stages over the last few years and only somewhat recently have `FindHDF5.cmake` / `hdf5-config.cmake` defined these. When I first introduced this into the cmake, some users suddenly couldn't build. I would *like* to pin to a hdf5 version, but there are others that need to be involved in that conversation.
2. The error logic is due to debugging some CMake. Before, `HDF5_INCLUDE_DIRS` was filled in by `find_path`, and this would cause `find_package` to not bail on modifying it (either due to an explicit check, or by poping a stack of scoped vars, or something else), leading to lots of `_NOTFOUND` warnings/errors. I have left the logic in as a sanity check, but it strictly doesn't need to be there.
3. Check for targets, see (1.)

singularity-eos/eos/eos_base.hpp
```diff
@@ -21,6 +21,15 @@
 #include <ports-of-call/portability.hpp>
 #include <ports-of-call/portable_errors.hpp>
 
+#ifdef SINGULARITY_USE_SPINER_WITH_HDF5
+#include <spiner/databox.hpp>
+namespace singularity {
+
+using DataBox = Spiner::DataBox<Real, std::true_type>;
```
Yurlungur, line 28: You shouldn't use `std::true_type` here. That's a SFINAE parameter. Let it resolve to the default value.

Yurlungur: Also it's probably better to put this type alias inside each relevant EOS class, as when @dholladay00 and I are done iwth hierarchical grids, the different EOS's might have different template parameters in their databoxes.

mauneyc-LANL: I needed explicit `true_type` for some errors before you updated `spiner` with the template parameters fixes. These are now removed.

`DataBox` is now scoped to the class, save for the tests where they are declared at file-scope (in `profile_eos/eospac.cpp`).

cmake/singularity-eos/kokkos.cmake
```diff
@@ -0,0 +1,111 @@
...
+macro(singularity_find_kokkoskernels)
+  find_package(KokkosKernels REQUIRED)
+endmacro()
+
+macro(singularity_enable_kokkoskernels target)
+  target_compile_definitions(${target} PUBLIC SINGULARITY_USE_KOKKOSKERNELS)
+  target_link_libraries(${target} PUBLIC Kokkos::kokkoskernels)
+endmacro()
+
+macro(_singularity_handle_custom_kokkos_tpl)
```
Yurlungur, line 68: What does this machinery do?

cmake/singularity-eos/spiner.cmake
```diff
@@ -0,0 +1,46 @@
...
+        CACHE BOOL "" FORCE)
+  endif()
+endmacro()
+
+macro(singularity_import_spiner)
+  if(NOT TARGET spiner::spiner)
+    _singularity_set_spiner_opt(SINGULARITY_USE_SPINER_WITH_HDF5 SPINER_USE_HDF)
+    # TODO I dunno if these are used
+    _singularity_set_spiner_opt(SINGULARITY_USE_KOKKOS SPINER_USE_KOKKOS)
+    _singularity_set_spiner_opt(SINGULARITY_USE_CUDA SPINER_USE_CUDA)
```
Yurlungur, line 14: They aren't used anymore. It's all passed through ports of call.

cmake/singularity-eos/ports-of-call.cmake
```diff
@@ -0,0 +1,13 @@
+macro(singularity_import_ports_of_call)
```
Yurlungur, line 1: Portability strategy: is it set by the host code in submodule mode and at install time for ports-of-call otherwise?

doc/sphinx/src/building.rst
```diff
@@ -1,23 +1,662 @@
-.. _building:
+Overview
+========
 
-Building Singularity-EOS
-=========================
+The ``singularity-eos`` build system is designed with two goals in mind
```
Yurlungur, line 4: somewhere in here we might want to say something about portability strategy through ports of call... even if we just link to ports of call documentation.

singularity-eos/eos/eos_base.hpp
```diff
@@ -21,6 +21,16 @@
...
 
+#if 0
+//#ifdef SINGULARITY_USE_SPINER_WITH_HDF5
+#include <spiner/databox.hpp>
+namespace singularity {
+
+using DataBox = Spiner::DataBox<Real, std::true_type>;
+
+} // namespace singularity
+#endif // SINGULARITY_USE_SPINER_WITH_HDF5
```
Yurlungur, line 32:
```suggestion
#endif // 0
```

singularity-eos/eos/eos_base.hpp
Yurlungur: :+1: perfect.

spack-repo/packages/singularity-eos/package.py
```diff
@@ -60,12 +66,32 @@ class SingularityEos(CMakePackage, CudaPackage):
...
+    depends_on("ports-of-call@1.5.1:", when="@1.7.1:") #TODO make sure version is correct
+    # request the HEAD of the main branch 
+    depends_on("ports-of-call@main", when="@main")
+
+    # spiner for sp5 table reading
+    # TODO: the `+kokkos` variant of spiner is not 
+    # operative (all `kokkos` code moved to `ports-of-call`).
+    # Should make a plan to restrict this variant to 
+    # older `spiner` versions
+    # (although having it is harmless)
```
Yurlungur, line 87: :+1: good catch

Yurlungur (approved):
Thanks for this---this was a big effort and a significant rewrite of the build system. I have to admit, I like the simpler cleaner build system that results from this PR.

Some nitpicks below, but I'm willing to be pushed back on them. None of them are blocking.

Yurlungur:
@dholladay00 please take a look at this when you have the chance. I'd like your eyes on it before we merge.

CMakeLists.txt
```diff
@@ -1,187 +1,311 @@
...
+option(SINGULARITY_USE_EOSPAC "Pull in eospac" OFF)
+
+cmake_dependent_option(SINGULARITY_USE_CUDA "Use CUDA backend of Kokkos" ON
+                       "SINGULARITY_USE_KOKKOS" OFF)
+cmake_dependent_option(
+  SINGULARITY_USE_KOKKOSKERNELS "Use KokkosKernels for LA routines" ON
+  "SINGULARITY_USE_KOKKOS" OFF)
+cmake_dependent_option(
+  SINGULARITY_BUILD_CLOSURE "Mixed Cell Closure" ON
+  "SINGULARITY_USE_KOKKOSKERNELS;SINGULARITY_USE_KOKKOS" OFF)
```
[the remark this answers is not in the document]
jhp-lanl: I think for future SAP work we'll want mixed cell closures (for hybrid EOS) but definitely _without_ the Kokkos/Kokkos Kernels dependency.

The main thing is that if you have closures on, you need some sort of linear solver compatible with the chosen portability strategy; you need Kokkos Kernels for Kokkos and you need Eigen for a `None` portability strategy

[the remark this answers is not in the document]
Yurlungur: Now that I think about it---where do the kokkos and cuda flags appear anymore? Since ports-of-call handles the portability strategy, is importing kokkos kernels the only place? There might be some opportunity for simplification there.

Regarding mixed cell closures without kokkos kernels, on GPU, that is currently hard to achieve because closures require an arbitrarily sized linear solver. I was imagining that on a first pass for the hybrid EOS, we would do a closure by hand which didn't require an arbitrarily sized matrix solve?

singularity-eos/eos/eos_base.hpp
dholladay00: ?

CMakeLists.txt
```diff
@@ -1,187 +1,311 @@
...
+# version 3.19 is available on all testing platforms
+cmake_minimum_required(VERSION 3.19)
+
+# Disable "in-source" builds
+if("${CMAKE_SOURCE_DIR}" STREQUAL "${CMAKE_BINARY_DIR}")
+  message(
+    FATAL_ERROR
+      "ERROR: Will not configure \"in-source\" builds. Create a seperate directory for building"
+  )
+endif()
```
dholladay00, line 25: This is a good addition. It prevents contaminating the source tree with a bunch of files.

CMakeLists.txt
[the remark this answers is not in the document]
dholladay00: The kokkos flags live here https://github.com/lanl/singularity-eos/blob/main/cmake/submodule_configs.cmake @Yurlungur. 

Other question, does this disable the ability to use Eigen for the solve @mauneyc-LANL ?

CMakeLists.txt
```diff
@@ -1,187 +1,311 @@
...
+option(SINGULARITY_BUILD_PYTHON "Compile Python bindings" OFF)
+option(SINGULARITY_BUILD_EXAMPLES "Compile examples" OFF)
+cmake_dependent_option(
+  SINGULARITY_BUILD_SESAME2SPINER
+  "Compile sesame2spiner"
+  ON
+  "SINGULARITY_USE_SPINER;SINGULARITY_USE_SPINER_WITH_HDF5;SINGULARITY_USE_EOSPAC"
+  OFF)
+cmake_dependent_option(
+  SINGUALRITY_BUILD_STELLARCOLLAPSE2SPINER "Compiler stellarcollapse2spiner" ON
```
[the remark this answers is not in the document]
dholladay00: I think this should be independently disable-able.

CMakeLists.txt
```diff
@@ -1,187 +1,311 @@
...
+  "SINGULARITY_BUILD_TESTS;SINGULARITY_BUILD_SESAME2SPINER" OFF)
+cmake_dependent_option(
+  SINGULARITY_TEST_STELLAR_COLLAPSE "Test stellar collapse table readers" ON
+  "SINGULARITY_BUILD_TESTS;SINGULARITY_BUILD_STELLARCOLLAPSE2SPINER" OFF)
+cmake_dependent_option(SINGULARITY_TEST_PYTHON "Test the Python bindings" ON
+                       "SINGULARITY_BUILD_TESTS;SINGULARITY_BUILD_PYTHON" OFF)
+
+# modify flags options
+option(SINGULARITY_BETTER_DEBUG_FLAGS "Better debug flags for singularity" ON)
+option(SINGULARITY_HIDE_MORE_WARNINGS "hide more warnings" OFF)
```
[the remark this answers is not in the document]
dholladay00: We should remove them. Does it need to be done in this PR though @Yurlungur ?

CMakeLists.txt
```diff
@@ -1,187 +1,311 @@
...
+# out a way to do SUBMODULE_MODE (use submodules for upstream deps) but also do
+# a proper install (FORCE_ disables the install/export stages). I may come back
+# to this, but there isn't a lot of clamoring for this right now so could return
+# to it in a later PR. something like
+# cmake_dependent_option(SINGULARITY_SUBMODULE_MODE_INSTALL ...)
+
+# TODO This should depend on using offloading (`SINGULARITY_USE_CUDA` &t) Also,
+# patching in the source tree is messy, this should be done in the build dir
+option(SINGULARITY_PATCH_MPARK_VARIANT
+       "Apply GPU patch to mpark-variant submodule" ON)
```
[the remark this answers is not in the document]
dholladay00: Our patch is (and will remain) compatible with all necessary backends via preprocessor. It is able to be patched unconditionally.

CMakeLists.txt
```diff
@@ -1,187 +1,311 @@
...
   endif()
-  if(SINGULARITY_USE_EOSPAC AND SINGULARITY_BUILD_SESAME2SPINER)
-    list(APPEND _subdirs2add sesame2spiner)
+  if(SINGULARITY_USE_KOKKOS)
+    singularity_import_kokkos()
+    if(SINGULARITY_USE_KOKKOSKERNELS)
+      singularity_import_kokkoskernels()
+    endif()
+  else()
+    singularity_import_eigen()
```
dholladay00, line 262: I think the linear algebra libraries should be hidden behind a `BUILD_CLOSURE` option.

CMakeLists.txt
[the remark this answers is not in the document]
jhp-lanl: To hopefully wrap up this tangent:

> I was imagining that on a first pass for the hybrid EOS, we would do a closure by hand which didn't require an arbitrarily sized matrix solve?

Yes I agree.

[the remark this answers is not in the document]
Yurlungur: Thanks @dholladay00 yeah that's right---the flags for kokkos kernels are actually pretty extensive. 

Regarding disabling Eigen: Are there currently any use-cases where Eigen would be used? SAP? It may be that we can safely remove Eigen but if so we should be deliberate about it.

CMakeLists.txt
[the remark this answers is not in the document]
Yurlungur: I think this just changes the defaults---you can still turn it off. That said maybe the default should be off?

CMakeLists.txt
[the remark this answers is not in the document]
Yurlungur: nope I was just thinking aloud. It can be done later.

CMakeLists.txt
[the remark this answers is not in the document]
Yurlungur: Should it just always be patched then, without an option at all?

CMakeLists.txt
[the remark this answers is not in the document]
dholladay00: Eigen is used when `+closure~kokkos-kernels` in spack parlance (it may actually be `+closure~kokkos` as I'm not sure if we allow `+Eigen+kokkos`)

cmake/singularity-eos/kokkos.cmake
mauneyc-LANL: This was for a custom kokkos directory, outside of either the system or submodule. 

I'll leave it to you to decide if you want to keep it.

dholladay00: ☝️

cmake/singularity-eos/ports-of-call.cmake
mauneyc-LANL: It's set if Kokkos is enabled in either case

```cmake
if(SINGULARITY_USE_KOKKOS)
  singularity_enable_kokkos(singularity-eos)
# ...
macro(singularity_enable_kokkos target)

  target_link_libraries(${target} PUBLIC Kokkos::kokkos)

  target_compile_definitions(${target} PUBLIC PORTABILITY_STRATEGY_KOKKOS)
```
if a host code configures `singularity-eos` to use `Kokkos`, but also tries to set `PORTABILITY_STRAT_NOTKOKKOS` then there will be problems (depending on when set).

singularity-eos/eos/eos_base.hpp
mauneyc-LANL: I can do this, but if you don't want the comment then just get rid of the comment? (why `// 0`?)

cmake/singularity/hdf5.cmake
mauneyc-LANL: The `MPI` requirement is solely for `HDF5`. If the `HDF5` that is located was built with `MPI`, then we need to bring in the library, otherwise there will be build errors.

cmake/singularity/Eigen3.cmake
mauneyc-LANL: Just makes it easier to manage. Some packages require a lot of setup/configuration so it's easier to isolate all that into a file. Ones with shorter macros could be inline, but it makes it clear where the code for doing the dependency stuff is (e.g., if something fails with `find_package`, the stack will show `cmake/singularity/<package>.cmake`).

Also, in future we can put more "package-specific" stuff here, like options or maintaining version differences.

CMakeLists.txt
mauneyc-LANL: `cmake/singularity-eos/mpark_variant.cmake`

CMakeLists.txt
```diff
@@ -1,187 +1,311 @@
...
 # publicly and display publicly, and to permit others to do so.
-#------------------------------------------------------------------------------#
-
-cmake_minimum_required(VERSION 3.14)
+# ------------------------------------------------------------------------------#
+
+# We want to stay near to latest, as more compilers and third-party libraries
+# are supported. However, uptake of new CMake releases can be slow. 20230403:
+# version 3.19 is available on all testing platforms
+cmake_minimum_required(VERSION 3.19)
```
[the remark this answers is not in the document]
mauneyc-LANL: Yes

cmake/singularity-eos/spiner.cmake
mauneyc-LANL: Commented out for now

singularity-eos/eos/eos_base.hpp
jhp-lanl: Yeah I'm a bit confused as well. I appreciate comments like this at `#endif` statements

cmake/singularity-eos/kokkos.cmake
Yurlungur: can I build the code "standalone" with the submodules again? If so the custom directory outside of the system or submodule isn't needed anymore.

singularity-eos/eos/eos_base.hpp
Yurlungur: Because this if case is a comment. If you look above, the whole block of code is commented out with `#if 0`

cmake/singularity/hdf5.cmake
Yurlungur: oh yeah I had a brain fart. I meant if the user has serial vs parallel HDF5 installed and wants to specify one or the other. I guess there's cmake flags for that right? That the user can specify?

CMakeLists.txt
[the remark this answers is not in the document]
Yurlungur: Right---I know but do we ever want that?

singularity-eos/eos/eos_base.hpp
dholladay00: I think we should remove it from the PR if it's universally removed by the preprocessor.

jhp-lanl: Ah sorry for the confusion. Yes I agree then

CMakeLists.txt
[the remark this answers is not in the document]
jhp-lanl: I don't think so...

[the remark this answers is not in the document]
Yurlungur: Probably we should keep it around until we have a little bit more clarity on what the FORTRAN codes will want. If they don't want us using Kokkos maybe we should hold on to Eigen for now (i.e., let's make sure it still works in this PR). But I think we can probably remove it in the near future as we figure out the plan for closures on device in those codes.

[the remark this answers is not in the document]
dholladay00: I could foresee @Yurlungur lamenting this decision someday b/c kokkos is easy to add for quick development but kernels is not. When future @Yurlungur is angry with this decision and out of patience b/c spack and dependencies are hard and some new machine is also hard, remember this conversation and know that my ability to sympathize with the frustration will be 0. 

That being said, I'm ok either way. If we want to allow kokkos-serial+eigen or no mixing at all, either way I am ok.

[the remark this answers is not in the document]
Yurlungur: @dholladay00 what are you suggesting is the lower-pain path? Keeping Eigen? I am perfectly willing to be convinced to keep at around to avoid my own suffering

CMakeLists.txt
mauneyc-LANL: Updated; `SINGULARITY_BUILD_CLOSURE` is now a direct option; `SINGULARITY_USE_KOKKOSKERNELS` now depends on it, and some checks are done to enable a new cache flag `SINGULARITY_USE_EIGEN`.

CMakeLists.txt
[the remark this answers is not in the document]
mauneyc-LANL: Latest commit to address this

```cmake
option(SINGULARITY_BUILD_CLOSURE "Mixed Cell Closure" ON)
# ...
cmake_dependent_option(
  SINGULARITY_USE_KOKKOSKERNELS "Use KokkosKernels for LA routines" ON
  "SINGULARITY_USE_KOKKOS;SINGULARITY_BUILD_CLOSURE" OFF)
# ...
if(SINGULARITY_BUILD_CLOSURE AND NOT SINGULARITY_USE_KOKKOSKERNELS)
  set(SINGULARITY_USE_EIGEN
      ON
      CACHE BOOL "" FORCE)
else()
  set(SINGULARITY_USE_EIGEN
      OFF
      CACHE BOOL "" FORCE)
endif()
```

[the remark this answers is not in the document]
jhp-lanl: > Latest commit to address this
> 
> ```cmake
> option(SINGULARITY_BUILD_CLOSURE "Mixed Cell Closure" ON)
> # ...
> cmake_dependent_option(
>   SINGULARITY_USE_KOKKOSKERNELS "Use KokkosKernels for LA routines" ON
>   "SINGULARITY_USE_KOKKOS;SINGULARITY_BUILD_CLOSURE" OFF)
> # ...
> if(SINGULARITY_BUILD_CLOSURE AND NOT SINGULARITY_USE_KOKKOSKERNELS)
>   set(SINGULARITY_USE_EIGEN
>       ON
>       CACHE BOOL "" FORCE)
> else()
>   set(SINGULARITY_USE_EIGEN
>       OFF
>       CACHE BOOL "" FORCE)
> endif()
> ```

That seems to make sense to me!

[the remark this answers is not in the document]
dholladay00: Unfortunately we need to add in a guard against mixing this with cuda: if use cuda and build closure and not use kokkos kernels, then error.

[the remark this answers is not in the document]
Yurlungur: probably fine to just add a line in the first if clause to error out if cuda is enabled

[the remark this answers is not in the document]
mauneyc-LANL: I've added a check for cuda/closure/kk

```cmake
# Use to determine if Eigen is used or not 
set(SINGULARITY_USE_EIGEN
      OFF
      CACHE BOOL "" FORCE)
if(SINGULARITY_BUILD_CLOSURE AND NOT SINGULARITY_USE_KOKKOSKERNELS)
  if(SINGULARITY_USE_CUDA)
    message(FATAL_ERROR "\"SINGULARITY_BUILD_CLOSURE=ON\" and \"SINGULARITY_USE_CUDA=ON\" requires \"SINGULARITY_USE_KOKKOSKERNELS=ON\"")
  set(SINGULARITY_USE_EIGEN
      ON
      CACHE BOOL "" FORCE)
endif()
```

doc/sphinx/src/building.rst
mauneyc-LANL: Does the build system need to be able interpret `PORTABILITY_STRATEGY_KOKKOS` as implying `SINGULARITY_USE_KOKKOS`? (right now it's the reverse).

singularity-eos/eos/eos_base.hpp
mauneyc-LANL: Removed the dead code

CMakeLists.txt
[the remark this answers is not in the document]
dholladay00: We should create an issue before resolving this convo.

CMakeLists.txt
```diff
@@ -1,187 +1,311 @@
...
+
+# modify flags options
+option(SINGULARITY_BETTER_DEBUG_FLAGS "Better debug flags for singularity" ON)
+option(SINGULARITY_HIDE_MORE_WARNINGS "hide more warnings" OFF)
+
+# toggle code options
+option(SINGULARITY_USE_SINGLE_LOGS
+       "Use single precision logs. Can harm accuracy." OFF)
+option(SINGULARITY_USE_TRUE_LOG_GRIDDING
+       "Use grids that conform to log spacing." OFF)
```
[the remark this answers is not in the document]
dholladay00: Should we create an issue to track @Yurlungur ?

CMakeLists.txt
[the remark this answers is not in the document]
dholladay00: I think it would be good to do so such that we always maintain compatibility, but we can save this for later if needed (if we wait for later, we should make an issue).

dholladay00 (approved):

dholladay00:
I have kicked off most recent branch on gitlab CI. Pending results, I think this is ready.

Yurlungur:
I'm happy to have this merged in when tests pass. Thanks for all the effort @mauneyc-LANL !

spack-repo/packages/singularity-eos/package.py
```diff
@@ -77,13 +96,18 @@ class SingularityEos(CMakePackage, CudaPackage):
...
     for _flag in ("~cuda", "+cuda", "~openmp", "+openmp"):
-        depends_on("kokkos@3.2: ~shared" + _flag, when="+kokkos" + _flag)
-        depends_on("kokkos-kernels@3.2:" + _flag, when="+kokkos-kernels" + _flag)
+        depends_on("kokkos ~shared" + _flag, when="+kokkos" + _flag)
+        depends_on("kokkos-kernels" + _flag, when="+kokkos-kernels" + _flag)
         depends_on("spiner" + _flag, when="+kokkos" + _flag)
 
     # specfic specs when using GPU/cuda offloading
+    # TODO Do we need `+aggressive_vectorization`, `+cuda_constexpr`, `~compiler_warnings` ?
     depends_on("kokkos +wrapper+cuda_lambda+cuda_relocatable_device_code", when="+cuda+kokkos")
```
dholladay00, line 111: how did I let this slip by. I thought we remove rdc from all of our spackages @mauneyc-LANL.

mauneyc-LANL: I don't recall doing anything with this; it's been in there for some time 

https://github.com/lanl/singularity-eos/blob/b5d7d8cd5c8525cc9d51a71102a645b9c1df6d6e/spack-repo/packages/singularity-eos/package.py#L80C9-L80C9

doc/sphinx/src/building.rst
```diff
@@ -1,23 +1,662 @@
...
+
+Dependencies
+------------
+
+``singularity-eos`` has a number of required and optional depdencies. 
+
+====================================== =============================== ===========================================
+  Package Name                          Distribution                    Comment 
+====================================== =============================== ===========================================
+ `spiner`_                              submodule [*]_ / external [*]_   Required
```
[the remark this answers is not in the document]
mauneyc-LANL: That's true, @mauneyc-LANL mark for update

doc/sphinx/src/building.rst
```diff
@@ -1,23 +1,662 @@
...
+ ``SINGULARITY_USE_CUDA``                       ``SINGULARITY_USE_KOKKOS=ON``                                                     Target nvidia GPUs for ``Kokkos`` offloading.
+ ``SINGULARITY_USE_KOKKOSKERNELS``              ``SINGULARITY_USE_KOKKOS=ON``                                                     Use Kokkos Kernels for linear algebra. Needed for mixed cell closure models on GPU.
+ ``SINGULARITY_BUILD_CLOSURE``                  ``SINGULARITY_USE_KOKKOS=ON`` ``SINGULARITY_USE_KOKKOSKERNELS=ON``                Mixed cell closure.
+ ``SINGULARITY_BUILD_SESAME2SPINER``            ``SINGULARITY_USE_SPINER=ON`` ``SINGULARITY_USE_SPINER_WITH_HDF5=ON``             Builds the conversion tool sesame2spiner which makes files readable by SpinerEOS.
+ ``SINGULARITY_BUILD_STELLARCOLLAPSE2SPINER``   ``SINGULARITY_USE_SPINER=ON`` ``SINGULARITY_USE_SPINER_WITH_HDF5=ON``             Builds the conversion tool stellarcollapse2spiner which optionally makes stellar collapse files faster to read.
+ ``SINGULARITY_TEST_SESAME``                    ``SINGULARITY_BUILD_TESTS=ON`` ``SINGULARITY_BUILD_SESAME2SPINER=ON``             Test the Sesame table readers.
```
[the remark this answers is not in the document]
mauneyc-LANL: This led me the realize an error in the testing with the update

PR #248, opened 2023-04-04
WIP: Use object inheritance rather than 'using' statements

jhp-lanl:
## PR Summary

As discussed with @Yurlungur and @dholladay00 , we use a lot of `using` statements in the code where I think standard object inheritance works instead. Wherever possible, this MR gets rid of `using` statements where they are not needed, or where the `this->` pointer can provide access to protected members variables in the base class. Of course, to do this I needed to make the derived classes `friend` of the base class to access these protected variables.

Note that members in the derived class will "shadow" or "hide" members of the same in the base class. We need the `using` statements to properly overload functions in the derived class with those in the base class, e.g. the vector overloads.

I would advocate for this sort of programming since the scope of the variable is very obvious where the variable is being modified or read. The `using` statements, especially in the PTE solver, do shorten the amount of code we need to write, but at the cost of increased ambiguity about where a variable is defined.

I've tested this on the x86-volta nodes of Darwin and it seems to work.

Tagging @jdolence as well since he wrote the PTE code where this changes a lot.

EDIT - I should add that this is a proposal rather than something I feel is a hard requirement.

## PR Checklist

### DIFF
--- a/singularity-eos/closure/mixed_cell_models.hpp
+++ b/singularity-eos/closure/mixed_cell_models.hpp
@@ -155,11 +155,6 @@ class CacheAccessor {
 
 template <typename EOSIndexer, typename RealIndexer>
 class PTESolverBase {
- friend class PTESolverRhoT;
- friend class PTESolverRhoU;
- friend class PTESolverRhoU;
- friend class PTESolverFixedP;
- friend class PTESolverFixedT;
  public:
   PTESolverBase() = delete;
   PORTABLE_INLINE_FUNCTION int Nmat() const { return nmat; }
--- a/singularity-eos/closure/mixed_cell_models.hpp
+++ b/singularity-eos/closure/mixed_cell_models.hpp
@@ -155,6 +155,11 @@ class CacheAccessor {
 
 template <typename EOSIndexer, typename RealIndexer>
 class PTESolverBase {
+ friend class PTESolverRhoT;
+ friend class PTESolverRhoU;
+ friend class PTESolverRhoU;
+ friend class PTESolverFixedP;
+ friend class PTESolverFixedT;
  public:
   PTESolverBase() = delete;
   PORTABLE_INLINE_FUNCTION int Nmat() const { return nmat; }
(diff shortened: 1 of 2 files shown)


singularity-eos/closure/mixed_cell_models.hpp
```diff
@@ -155,6 +155,12 @@ class CacheAccessor {
 
 template <typename EOSIndexer, typename RealIndexer>
 class PTESolverBase {
+  friend class PTESolverRhoT;
+  friend class PTESolverRhoU;
+  friend class PTESolverRhoU;
+  friend class PTESolverFixedP;
+  friend class PTESolverFixedT;
```
Yurlungur, line 162: I really dislike using `friend class` here as it mixes inheritance access in a way that I worry is hard to disentangle. For example, having the base class know about its children with `friend` means that a downstream code can't build on our solver infrastructure to add their own solver with our base functionality. They have to make a whole new class, since they can't add "friendship" downstream. It's also just one more place things have to be modified when a new class is added.

(Big picture, I think `friend` is a language feature that should be used very sparingly, as it breaks encapsulation. It essentially means all the classes/functions that are `friends` are part of a single monolithic implementation. Sometimes this is the right thing to do. But when faced with that design choice, I think it's worth thinking carefully about alternatives.)

That said, it's actually not clear to me: why is this necessary here? Maybe I'm missing something about the inheritance structure.

singularity-eos/eos/eos_base.hpp
```diff
@@ -79,8 +79,7 @@ char *StrCat(char *destination, const char *source) {
   using EosBase<EOSDERIVED>::PTofRE;                                                     \
   using EosBase<EOSDERIVED>::FillEos;                                                    \
   using EosBase<EOSDERIVED>::EntropyFromDensityTemperature;                              \
-  using EosBase<EOSDERIVED>::EntropyFromDensityInternalEnergy;                           \
-  using EosBase<EOSDERIVED>::EntropyIsNotEnabled;
+  using EosBase<EOSDERIVED>::EntropyFromDensityInternalEnergy;
```
Yurlungur, line 82: :+1:

Yurlungur (commented):
I like the cleanup of `eos_base`. I'm not sure how I feel about the changes in `mixed_cell_models.hpp`.

Personally I find the added `this->` calls hinder readability, but I don't feel so strongly that I would veto this change if I'm outvoted.

That said I do worry about using the `this` pointer on device. My intuition says this might be problematic, though I can't point to a particular reason at the moment. Some additional comments below.

singularity-eos/closure/mixed_cell_models.hpp
jhp-lanl: I want to emphasize that these changes were made to duplicate the functionality that already exists through the `using` statements we already had.

> I really dislike using friend class here as it mixes inheritance access in a way that I worry is hard to disentangle.

Is this worse than mixing data access through `using` statements?

> ...having the base class know about its children with friend means that a downstream code can't build on our solver infrastructure to add their own solver with our base functionality

They can still adopt the `using` pattern that we currently have and accomplish the same thing. This change only _enables_ data access through inheritance.

> Big picture, I think friend is a language feature that should be used very sparingly, as it breaks encapsulation. It essentially means all the classes/functions that are friends are part of a single monolithic implementation.

Maybe I'm failing to grasp some of the nuances here, but I'd argue that our current implementation already does this.


I'm certainly open to alternatives, but to me (and again, I'm new to C++ so I may be completely wrong) the `using` statements feel like a hack to expose variables to a different scope.

dholladay00: I tend to agree with @Yurlungur. While neither solution is optimal, I'd prefer the `using` statements over `this->`. As my simple mind (and procedural programming tendencies) sees things, the usage of a base class is to reduce code duplication. Both of these implementations accomplish this task, but I find the `using` route to be more readable and the absence of `friend` classes an additional benefit as it communicates to downstream projects that one must register their implementations with singularity-eos in order to "correctly" inherit our base class's functionality, which is not something we want to communicate IMO.

Yurlungur: Discussing with @jhp-lanl offline---I'm actually confused why the `friend` stuff is needed. I can't reproduce in some simple examples.

We also discovered that the `this` pointer is only required because of the templates (which is my memory of the language). This has to do with implementation details of how unqualified name lookup interacts with templates in the language. From the C++11 standard:

> In the definition of a class or class template, if a base class depends on a template-parameter, the base class scope is not examined during unqualified name lookup either at the point of definition of the class template or member or during an instantiation of the class template or member. 

To my mind, this makes both `using` and `this` essentially equivalent "hacks" around the failure of the language to properly do unqualified name lookup. So I think it's just a question of which we as a team prefer stylistically. (I feel pretty strongly I don't want `friend` classes here... but maybe we can work around that if we decide to go with the `this` pointer.)

jhp-lanl: I removed the `friend` stuff because it isn't needed. I _thought_ it was needed at some point, but that was my misunderstanding.

I agree on the `using` vs `this->`. Given that we have two votes against, I'm inclined to scrap it.

jhp-lanl:
I think this is far outdated and not needed anymore

PR #230, opened 2023-02-08
Draft: propogate EOSPAC to wrapper correctly on install

mauneyc-LANL:
## PR Summary

`eospac` was handled a little differently as a dependency for reasons I can't recall, but this different handling got things mixed up with `eospac-wrapper` down the line.

This is hacked together to get it working, but I will return to a more streamlined solution soon

## PR Checklist

- [ ] Adds a test for any bugs fixed. Adds tests for new features.
- [ ] Format your changes by using the `make format` command after configuring with `cmake`.
- [ ] Document any new features, update documentation for changes made.
- [ ] Make sure the copyright notice on any files you modified is up to date.
- [ ] After creating a pull request, note it in the CHANGELOG.md file
- [ ] If preparing for a new release, update the version in cmake.

PR #253, opened 2023-04-11
Fix for EOSPAC export/install scripts

mauneyc-LANL:
## PR Summary

This is a replacement for #230 .

The export/install stage will generate the file `singularity-eosConfig.cmake`, which is used by the `find_package()` command downstream. This file checks for the required dependencies, defines targets and some variables.

```cmake
# example of prior singularity-eosConfig.cmake, notice that EOSPAC is missing from the pattern. This is a bug.
if(NOT spiner_FOUND OR NOT TARGET spiner::spiner)
        find_package(spiner  REQUIRED)
endif()
if(NOT HDF5_FOUND OR NOT TARGET hdf5::hdf5 OR NOT TARGET hdf5::hdf5_hl)
        find_package(HDF5 COMPONENTS C;HL REQUIRED)
endif()
if(NOT MPI_FOUND OR NOT TARGET MPI::MPI_CXX)
        find_package(MPI COMPONENTS CXX REQUIRED)
endif()
```

There was a typo in the variable used to add `EOSPAC` to the set of required packages, and a variable for checking on the correct target guard. With this fix, we have the correct propagation:

```cmake
# example of new singularity-eosConfig.cmake
if(NOT spiner_FOUND OR NOT TARGET spiner::spiner)
        find_package(spiner  REQUIRED)
endif()
if(NOT HDF5_FOUND OR NOT TARGET hdf5::hdf5 OR NOT TARGET hdf5::hdf5_hl)
        find_package(HDF5 COMPONENTS C;HL REQUIRED)
endif()
if(NOT MPI_FOUND OR NOT TARGET MPI::MPI_CXX)
        find_package(MPI COMPONENTS CXX REQUIRED)
endif()
if(NOT EOSPAC_FOUND OR NOT TARGET EOSPAC::eospac)
        find_package(EOSPAC  REQUIRED)
endif()
```

This only effects downstream codes that rely on `find_package()` for resolving `singularity-eos`; in-tree builds, or otherwise "install-less" use-cases, would be unaffected.

## PR Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -318,7 +318,9 @@ if (SINGULARITY_USE_EOSPAC)
   list(APPEND SINGULARITY_PUBLIC_DEFINES SINGULARITY_USE_EOSPAC)
   list(APPEND SINGULARITY_PUBLIC_LIBS EOSPAC::eospac)
   list(APPEND SINGULARITY_PUBLIC_LIBS eospac-wrapper)
-  list(APPEND SINGULARITY_DEPS_PKGS EOSPAC)
+  list(APPEND SINGULARITY_DEP_PKGS EOSPAC)
+  # tell the singularity-eosConfig.cmake to get this target
+  set(SINGULARITY_DEP_TARGETS_EOSPAC "EOSPAC::eospac")
 
   list(APPEND _subdirs2add eospac-wrapper)
 endif ()


dholladay00:
@mauneyc-LANL when are the circumstances in which eospac would be a target like that already? 

Would this need to change again with your other larger cmake changes? Is this a change that could be part of that PR or is it better to separate? If not, we could make the target branch your cmake changes branch.

mauneyc-LANL:
> @mauneyc-LANL when are the circumstances in which eospac would be a target like that already?
> 
> Would this need to change again with your other larger cmake changes? Is this a change that could be part of that PR or is it better to separate? If not, we could make the target branch your cmake changes branch.

If it was defined beforehand. `find_package(EOSPAC)`, `find_package(singularity-eos)`.

This is simple change, merging it now is easiest path.

mauneyc-LANL:
> This is just a 1-line change? Assuming this resolves build issues, LGTM.
> 
> Longer-term it seems like our "install" tests aren't catching this stuff. Is there another CI test we could add that catches the use-case we're hitting?

Yes; we could update the examples to be "isolated" from the build and do a `find_package` (this is how I'm testing locally)

mauneyc-LANL:
> This is just a 1-line change? Assuming this resolves build issues, LGTM.
> 

This resolves a hypothetical build issue, e.g. someone does `find_package(singularity-eos)` but hasn't defined the `EOSPAC` target. In all extant use-cases, this isn't strictly necessary as this is already present in the build. But this is to cover any case where it would possibly not be (that is, the installed `singularity-eos` was built with `EOSPAC`, but for whatever reason some host code doesn't do `find_package(EOSPAC)` or similar before doing `find_package(singularity-eos)`).

Yurlungur:
> > This is just a 1-line change? Assuming this resolves build issues, LGTM.
> > Longer-term it seems like our "install" tests aren't catching this stuff. Is there another CI test we could add that catches the use-case we're hitting?
> 
> Yes; we could update the examples to be "isolated" from the build and do a `find_package` (this is how I'm testing locally)

This is probably a good idea. So long as we can also let users build examples easily still. For another day though.

PR #254, opened 2023-04-11
Add the ability to pull an unmodified EOS out of a modified one

Yurlungur:
## PR Summary

Sometimes you may wish to have access to the unmodified model... for example, because you wish to access model-specific features (i.e., the mass fractions in the stellar collapse tables), because you want to drop out of code units and use cgs, or because you wish to swap one modifier for another at runtime.

This PR lets you do those things. I provide three functions, shared across all models, that let you do this:
1. `bool Model::IsModified() const` returns true if the EOS has a modifier applied to it and false otherwise.
2. `auto UnmodifyOnce()` if the EOS is modified, pull off the most recent modifier. Returns either the variant or an appropriate type.
3. `auto GetUnmodifiedObject()` returns the underlying EOS with all modifiers removed. Type again is either the variant or the appropriate type.

This feature is needed by the NGP codes but I believe it will be useful broadly for any downstream code that uses C++. (Obviously it's of no help to the fortran codes.)

## PR Checklist

### DIFF
--- a/singularity-eos/eos/eos_base.hpp
+++ b/singularity-eos/eos/eos_base.hpp
@@ -424,6 +424,14 @@ class EosBase {
     impl::StrCat(msg, "' EOS");
     PORTABLE_ALWAYS_THROW_OR_ABORT(msg);
   }
+
+  // Tooling for modifiers
+  PORTABLE_FORCEINLINE_FUNCTION
+  bool IsModified() const { return false; }
+  PORTABLE_FORCEINLINE_FUNCTION
+  auto UnmodifyOnce() { return *static_cast<CRTP *>(this); }
+  PORTABLE_FORCEINLINE_FUNCTION
+  auto GetUnmodifiedObject() { return *static_cast<CRTP *>(this); }
 };
 } // namespace eos_base
 } // namespace singularity
--- a/singularity-eos/eos/eos_variant.hpp
+++ b/singularity-eos/eos/eos_variant.hpp
@@ -840,6 +840,22 @@ class Variant {
         eos_);
   }
 
+  // Tooling for modifiers
+  PORTABLE_FORCEINLINE_FUNCTION
+  bool IsModified() const {
+    return mpark::visit([](const auto &eos) { return eos.IsModified(); }, eos_);
+  }
+  PORTABLE_FORCEINLINE_FUNCTION
+  Variant UnmodifyOnce() {
+    return mpark::visit(
+        [](auto &eos) { return eos_variant<EOSs...>(eos.UnmodifyOnce()); }, eos_);
+  }
+  PORTABLE_FORCEINLINE_FUNCTION
+  Variant GetUnmodifiedObject() {
+    return mpark::visit(
+        [](auto &eos) { return eos_variant<EOSs...>(eos.GetUnmodifiedObject()); }, eos_);
+  }
+
   PORTABLE_INLINE_FUNCTION
   unsigned long PreferredInput() const noexcept {
     return mpark::visit([](const auto &eos) { return eos.PreferredInput(); }, eos_);
--- a/singularity-eos/eos/modifiers/eos_unitsystem.hpp
+++ b/singularity-eos/eos/modifiers/eos_unitsystem.hpp
@@ -269,6 +269,13 @@ class UnitSystem : public EosBase<UnitSystem<T>> {
     printf("Units = %e %e %e %e\n", rho_unit_, sie_unit_, temp_unit_, press_unit_);
   }
 
+  PORTABLE_FORCEINLINE_FUNCTION
+  bool IsModified() const { return true; }
+  PORTABLE_FORCEINLINE_FUNCTION
+  T UnmodifyOnce() { return t_; }
+  PORTABLE_FORCEINLINE_FUNCTION
+  auto GetUnmodifiedObject() { return t_.GetUnmodifiedObject(); }
+
  private:
   T t_;
 
--- a/test/test_eos_unit.cpp
+++ b/test/test_eos_unit.cpp
@@ -232,7 +232,7 @@ SCENARIO("EOS Variant Type", "[Variant][EOS]") {
   std::cout << demangle(typeid(EOS).name()) << std::endl;
 }
 
-SCENARIO("EOS Builder and Modifiers", "[EOSBuilder],[Modifiers][IdealGas]") {
+SCENARIO("EOS Builder and Modifiers", "[EOSBuilder][Modifiers][IdealGas]") {
 
   GIVEN("Parameters for a shifted and scaled ideal gas") {
     constexpr Real Cv = 2.0;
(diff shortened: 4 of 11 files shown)


singularity-eos/eos/eos_base.hpp
```diff
@@ -424,6 +424,14 @@ class EosBase {
...
     PORTABLE_ALWAYS_THROW_OR_ABORT(msg);
   }
+
+  // Tooling for modifiers
+  PORTABLE_FORCEINLINE_FUNCTION
+  bool IsModified() const { return false; }
+  PORTABLE_FORCEINLINE_FUNCTION
+  auto UnmodifyOnce() { return *static_cast<CRTP *>(this); }
+  PORTABLE_FORCEINLINE_FUNCTION
+  auto GetUnmodifiedObject() { return *static_cast<CRTP *>(this); }
```
Yurlungur, line 434: These provide default behavior for the "unmodified" EOS models. The modifiers must implement these differently and shadow them, (although there's an obvious template).

singularity-eos/eos/modifiers/eos_unitsystem.hpp
```diff
@@ -269,6 +269,13 @@ class UnitSystem : public EosBase<UnitSystem<T>> {
     printf("Units = %e %e %e %e\n", rho_unit_, sie_unit_, temp_unit_, press_unit_);
   }
 
+  PORTABLE_FORCEINLINE_FUNCTION
+  bool IsModified() const { return true; }
+  PORTABLE_FORCEINLINE_FUNCTION
+  T UnmodifyOnce() { return t_; }
+  PORTABLE_FORCEINLINE_FUNCTION
+  auto GetUnmodifiedObject() { return t_.GetUnmodifiedObject(); }
```
Yurlungur, line 277: Unfortunately, these are copy-pasted into each modifier definition. It's maybe possible to add them all with a macro or some other trick, although I don't know how worthwhile that is.

test/test_eos_unit.cpp
```diff
@@ -232,7 +232,7 @@ SCENARIO("EOS Variant Type", "[Variant][EOS]") {
   std::cout << demangle(typeid(EOS).name()) << std::endl;
 }
 
-SCENARIO("EOS Builder and Modifiers", "[EOSBuilder],[Modifiers][IdealGas]") {
+SCENARIO("EOS Builder and Modifiers", "[EOSBuilder][Modifiers][IdealGas]") {
```
Yurlungur, line 235: This was a harmless bug. The comma never should have been there. We need it not there, in fact, to correctly parse this string for categories with catch.

singularity-eos/eos/modifiers/eos_unitsystem.hpp
jhp-lanl: Two tier CRTP? There would be the standard base class and then a modifier base class that inherits from the standard base class and provides these functions. I played around with it a bit a while back and it is doable.
https://stackoverflow.com/a/18174442/3491638

jhp-lanl:
> @jhp-lanl I can't remember exactly, but I want to make sure that this won't break any plans for vector interface modifiers. There is something about multiple parallel_fors but I am forgetting why.

I think the underlying issue was with modifiers being applied to EOSPAC. Currently the modifiers inherit from the base class, their default vector behavior is to call their own scalar functions which modify scalar EOS calls. In the EOSPAC case, we need the vector modifier to call the vector EOS call with the appropriate modifiers applied to the inputs/outputs.

I remember we had discussed whether this required another set of parallel for loops or whether there was a better way that maybe could use serial for loops to apply the modifiers and then use the parallel for to apply the modifier. This might be less performant for EOS other than EOSPAC though, so we would want to do some profiling.

> I don't think we modify any underlying data such that an unmodify would require (un)modifying internal EOS state (like tables), but I'd like you to think that through and double check me on that.

If we did apply the modifiers to the data rather than the inputs/outputs, you'd be correct that this might be a non-trivial operation, but right now we don't do that. I'd advocate for trying out a simple for loop to modify inputs/outputs and assess the performance impact from such an approach before we go down the road of modifying internal class states.

jhp-lanl (approved):
I don't like the copy/paste but it's not enough for me not to approve this.

singularity-eos/eos/modifiers/eos_unitsystem.hpp
Yurlungur: Hmm... that could work. If we add more modifier-specific stuff, let's consider it.

Yurlungur:
I'm actually not sure what's going on with the docs. The docs built, but the deploy stage failed because the hash is wrong. Very strange.

Yurlungur:
Must have been a momentary thing. Docs build now.

Yurlungur:
Gotten approval from @jhp-lanl  and(verbally) from @dholladay00 so I'm merging it. Please feel free to add feedback after the fact though. We're not set in stone here.

singularity-eos/eos/modifiers/eos_unitsystem.hpp
mauneyc-LANL: For consideration: mixins fit this pattern well, tho (from experience) I can say using mixins and crtp can be give tongue-twisty code. https://www.fluentcpp.com/2017/12/12/mixin-classes-yang-crtp/

dholladay00: If @jhp-lanl wants to make an PR, I'd be interested to see how tongue-twisty it is. May be worth it, though the current duplication overhead is small and not likely to be modified often.

jhp-lanl: @mauneyc-LANL I was looking at mixins as well a little bit ago but I thought that they weren't as elegant with our `Variant` class. We define a closed set of functions that are available for all types, whereas I feel a mixin is about expanding the functionality. If we were using class inheritance for our type erasure, maybe a mixin would be the preferable approach, but with the Variant, I lean towards CRTP since the end user doesn't have to change anything.

However, I wonder if a mixin could enable functionality like this?
```c++
EOS my_modified_eos =  Shifted<Scaled<IdealGas(...)>>>;
EOS once_unmodified = UnmodifyOnce(my_modified_eos);
EOS unmodified_eos = UnmodifyAll(my_modified_eos);
```

singularity-eos/eos/eos_variant.hpp
```diff
@@ -840,6 +840,22 @@ class Variant {
...
+  PORTABLE_FORCEINLINE_FUNCTION
+  Variant UnmodifyOnce() {
+    return mpark::visit(
+        [](auto &eos) { return eos_variant<EOSs...>(eos.UnmodifyOnce()); }, eos_);
+  }
+  PORTABLE_FORCEINLINE_FUNCTION
+  Variant GetUnmodifiedObject() {
+    return mpark::visit(
+        [](auto &eos) { return eos_variant<EOSs...>(eos.GetUnmodifiedObject()); }, eos_);
+  }
```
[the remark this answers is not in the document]
Yurlungur: All EOS's are assumed to have reference semantics, including databoxes, so copying will not cause correctness problems, though it may cause performance bottlenecks.

I was not aware that one could do a type-matching visit. That's cool. There may be other uses for that.

PR #264, opened 2023-06-20
Helmholtz EOS

Yurlungur:
## PR Summary

Preliminary work on implementing a Helmholtz EOS in singularity-eos

## PR Checklist


jhp-lanl:
I'm sure with a paper and some time I could decipher much of this, but as it stands I'm a bit concerned by all the short variable names flying around without context. Is there any way the variable names can be more descriptive?

Yurlungur:
> I'm sure with a paper and some time I could decipher much of this, but as it stands I'm a bit concerned by all the short variable names flying around without context. Is there any way the variable names can be more descriptive?

:+1: I blindly copied this out of the reference implementation---I will add some documentation in sphinx as well as some comments pointing to the reference implementation an the paper. Would that be sufficient to clarify?

jhp-lanl:
> > I'm sure with a paper and some time I could decipher much of this, but as it stands I'm a bit concerned by all the short variable names flying around without context. Is there any way the variable names can be more descriptive?
> 
> 👍 I blindly copied this out of the reference implementation---I will add some documentation in sphinx as well as some comments pointing to the reference implementation an the paper. Would that be sufficient to clarify?

I think that will be helpful, but I just don't feel that reading the paper should be needed for the code review. Some basic comments to explain the general structure of the code and explanations for what blocks of variables are doing (or more descriptive variable names) will probably aid the review process.

Yurlungur:
> I think that will be helpful, but I just don't feel that reading the paper should be needed for the code review. Some basic comments to explain the general structure of the code and explanations for what blocks of variables are doing (or more descriptive variable names) will probably aid the review process.

I can do that. The variable names are, for the most part, meaningfully self consistent. I think some comments explaining the convention will clarify.

singularity-eos/base/math_utils.hpp
```diff
@@ -22,32 +22,40 @@ namespace math_utils {
...
-  return x * pow<P - 1>(x);
+template <int P>
+PORTABLE_FORCEINLINE_FUNCTION constexpr Real pow(const Real x) {
+  // defining Ppos is important. Otherwise compiler is not smart
+  // enough to terminate recursion.
+  constexpr int Ppos = (P < 0) ? -P : P;
+  return (P >= 0)
+             ? ((Ppos % 2) ? x * pow<Ppos - 1>(x) : pow<Ppos / 2>(x) * pow<Ppos / 2>(x))
+             : 1 / pow<Ppos>(x);
 }
```
Yurlungur, line 39: Very pleased with this new version of our integer power function. It's generically recursive, with the even/odd power performance trick, so we don't need the strategic template specializations anymore. And it handles negative integer powers.

singularity-eos/eos/eos.hpp
```diff
@@ -63,7 +64,7 @@ using singularity::detail::transform_variadic_list;
 
 // all eos's
 static constexpr const auto full_eos_list =
-    tl<IdealGas, Gruneisen, Vinet, JWL, DavisReactants, DavisProducts
+    tl<IdealGas, Gruneisen, Helmholtz, Vinet, JWL, DavisReactants, DavisProducts
```
dholladay00, line 67: Should this be behind a preprocessor macro?

dholladay00: Given the applicability is primarily targeted at astrophysical regimes, I think maybe this EOS should be opt in, meaning that all headers and implementation related files and tests should not be added to cmake headers, etc., or the type included in the variant, unless turned on. @Yurlungur thoughts?

Yurlungur: I think that makes sense @dholladay00 . I think my preference would be that the header file is always included, but tests are opt in and it's simply not included in the variant. Instead, downstream codes that expect to use it must add it to their own local variant.

Perhaps we should do the same for the stellar collapse tables.

dholladay00: If this solution is not overly onerous for downstream codes that rely on these EOSs that is a reasonable solution, another good solution is something like what is done for eospac's inclusion in the variant.

Yurlungur: I think it's not too bad for them. Also given that astro codes are unlikely to want the terrestrial EOS'S and modifiers, it's maybe a good solution

test/test_eos_helmholtz.cpp
```diff
@@ -0,0 +1,170 @@
...
+#include <ports-of-call/portable_arrays.hpp>
+#include <ports-of-call/portable_errors.hpp>
+#include <singularity-eos/base/constants.hpp>
+#include <singularity-eos/eos/eos.hpp>
+#include <test/eos_unit_test_helpers.hpp>
+
+using Catch::Matchers::WithinRel;
+using singularity::Helmholtz;
+const std::string filename = "../test/helmholtz/helm_table.dat";
+SCENARIO("Helmholtz equation of state - Table interpolation (tgiven)", "[HelmholtzEOS]") {
```
Yurlungur, line 39: Looks like the tests are not GPU. We should test on device before merge. It's an easy change, and I can add it.

Yurlungur (commented):
This looks pretty good! Thanks for all the hard work and bug fixing @AlexHls !

.github/workflows/tests.yml
```diff
@@ -35,6 +35,8 @@ jobs:
                   -DSINGULARITY_BUILD_EXAMPLES=ON \
                   -DSINGULARITY_BUILD_PYTHON=ON \
                   -DSINGULARITY_TEST_SESAME=OFF \
+                  -DSINGULARITY_USE_HELMHOLTZ=ON \
```
mauneyc-LANL, line 38: @Yurlungur Do we want to have an option for this? It doesn't change anything in the configuration (e.g. require some other dependency or disable some other feature). If it's just a new EoS in a header file, treat it like all the others. The test can be put into the standard test targets.

I suppose that it could be the a case that since the data directory likely needs to be part of the install, but that would probably be standard anyway. Your call.

Yurlungur: I think the main benefit is it prevents it from being in the variant by default. That said, we could just not include it in the variant, and remove the build option.

CMakeLists.txt
```diff
@@ -233,6 +239,9 @@ endif()
 if(SINGULARITY_BUILD_CLOSURE)
   target_compile_definitions(singularity-eos PRIVATE SINGULARITY_BUILD_CLOSURE)
 endif()
+if(SINGULARITY_USE_HELMHOLTZ)
```
mauneyc-LANL, line 242: you should add something like

```cmake
add_custom_command(
  TARGET singularity-eos
  PRE_BUILD
  COMMAND ${CMAKE_COMMAND} -E copy_directory ${CMAKE_SOURCE_DIR}/data
          $<TARGET_FILE_DIR:singularity-eos>/data)
```
which will copy the data directory to the build directory.

Also will need to add similar in `cmake/install.cmake` for installing

.github/workflows/tests.yml
brryan: In the limit of infinite EOSes in `singularity-eos`, is the solution to infinite compile times due to variant to give each EOS, not just Helmholtz, a similar enable/disable flag?

Yurlungur: I think the long term plan is not to provide a single variant, but instead have a few predefined variants for different use-cases. And if those don't meet needs, then the downstream code builds their own. TO enable this, I need to refactor `EOSBuilder`, which is on my TODO list. This enable/disable flag is a short-term hack that gets us short compile times until I can get that cleaned up.

Yurlungur:
I think that's a good point @mauneyc-LANL

CMakeLists.txt
mauneyc-LANL: Ah, I see this was done in the test directory. Nevermind.

.github/workflows/tests.yml
brryan: Naively I would think it would be easier to let the downstream code request the EOSes they want through compile time flags like this rather than build their own variant which can be time consuming to work out. Is it difficult for `singularity-eos` to provide a single mega-variant that is pruned down at compile time through options like this? Such a mega-variant could presumably also be pre-pruned down into the predefined variants you have in mind,

Yurlungur: Yeah I'm open to suggestions here, including the mega variant with pruning options.

My main concern about that is that the combinatoric list of EOS's and variants might be quite large and complex. Letting the downstream C++ codes build their own variants moves some of that burden away from singularity-eos. 

That said, a few predefined variants probably makes sense for common use cases... e.g., I want all the astro-relevant EOS's or all the terrestrial-relevant EOS's. That kind of thing.

CMakeLists.txt
mauneyc-LANL: Still need to add a step to install the data. Something like

```cmake
# make sure `include(GNUInstallDirs)` was done beforehand - should be already
install(DIRECTORY ${PROJECT_SOURCE_DIR}/data DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/singularity-eos)
#NB this usually means `<install-prefix>/share`
```

.github/workflows/tests.yml
jhp-lanl: If the issue is compile times, I'd also like to know exactly where the compilation is getting stuck. @dholladay00 today just told me that `PTofRE()` might be a prime target for example.

dholladay00: @jhp-lanl `PTofRE` is on my list to remove entirely. The improved compile times should now be available in the `main` branch presently.

dholladay00: I am open to both encouraging downstream codes to use the variadic utils to create their own variants as well as enabling the main variant to have configuration-time toggle-able options. I'd love to hear of folks' thoughts on how to best manage that. Is there a way to get cmake to do some of that code-gen magic for us @mauneyc-LANL ? Then there's the question of if we expose this at the cmake level do we also expose it at the spack level?

Long term, I think the best solution would be to JIT the variant based on requested models in a simulation, but that is a distant dream at this point. 

This is probably best suited for a broader discussion and not in this PR.

singularity-eos/eos/eos_helmholtz.hpp
```diff
@@ -0,0 +1,1439 @@
...
+  Real T;
+
+  if (options_.ENABLE_RAD || options_.GAS_DEGENERATE ||
+      options_.ENABLE_COULOMB_CORRECTIONS) {
+    Real lTguess = lambda[Lambda::lT];
+    if (!((electrons_.lTMin() <= lTguess) && (lTguess <= electrons_.lTMax()))) {
+      lTguess = lTAnalytic_(rho, e, ni, ne);
+      if (!((electrons_.lTMin() <= lTguess) && (lTguess <= electrons_.lTMax()))) {
+        // Fallback initial guess
+        lTguess = 7;
```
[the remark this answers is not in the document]
AlexHls: This was meant to be a fallback initial guess in the middle of the table range (in case the analytic guess fails/ is outside the table range for some reason), I just failed to do a simple calculation. Fixed now to 8 (table range is from 3 to 13). It should in any case not matter that much since it is an arbitrary guess anyway.

test/test_eos_helmholtz.cpp
AlexHls: I included a test on device, not sure if that is what you meant. In the second test I also tired define the input values as ``Kokkos::View`` in the same way as in the first test, but for some reason the test then fails. (I made sure to use the host values for the comparison, but the ``temp_new`` value reported by the ``REQUIRE_THAT`` call differs from the value to match. However, if I print the ``temp_new`` value before that, it matches the compared value perfectly. I presume it is a trivial thing I just missed, but I wasn't able to find the reason for that so far.

In any case, the current tests should test on device. Tested locally on a RTX 4060 with CUDA 12.2.91. + gcc 12.3.0

AlexHls:
I added a Fortran interface, but I could not test it. Compiling (more precisely linking) my test program lead to quite a few error messages. I'm not sure if is because I forgot something in the interface or due to the issue described in #284 (given that the error message mentions several EOS is suspect the latter). I would need someone else to have a look at the interface to be sure.

AlexHls:
_(This part is for archival purposes, i.e. that there is a reference somewhere that we thoroughly tested the implementation against a reference implementation)_

# Testing

## Table interpolation

-  Tested a 100x100 grid of (rho, T) values covering the table range, avoiding the edges.
-  Compared all relevant quantities that singularity-eos computes against a reference implementation (Fig 1).
-  The tests here do not consider Coulomb corrections as the cutoff is handled differently in the reference implementation (i.e. it just cuts the corrections off at a certain point, whereas we implement a butterworth filter to achieve a smooth transition)
-  A Figure is included to illustrate the effects of Coulomb corrections (Fig. 2). One can see that once the parameter space approaches the cut-off region, the difference between the implementations becomes substantial.

TableInterpolation_NoCoulombCorr
Figure 1
TableInterpolation_WithCoulombCorr
Figure 2

## Root finding

The root finding is checked rather through a self-consistency check than a comparison against a reference implementation. We did compare it against the reference implementation, however, due to the different root finding algorithms and slightly different bounding mechanisms this comparison is rather difficult to interpret.

Here we test self-consistency by going 'full circle', i.e. we calculate the internal energy through the table interpolation 
$$e_{int} = eos(\rho, T)$$ which is used as input for the root find: $$T_{new} = eos(\rho, e_{int})$$.
This new temperature should be identical with the input temperature; all other quantities will be computed from this quantity. As long as $T_{new}$ is correct the other quantities will be computed correctly as long as the table interpolation itself works correctly.

Fig 3. shows the difference between the original value and the value obtained from the root find without an initial guess for the temperature. It can be seen that the root find struggles in the low temperature - low density region. But once a sensible initial guess is provided (which it realistically will be in a production scenario) the root find works much better (see Fig 4).

(Both Figures also show $e_{int,new} = eos(\rho, T_{new})$ vs $e_{int}$ to illustrate that once the correct temperature is obtained, other quantities will be computed correctly.)

RootFinding_NoInitialGuess
Figure 3.

RootFinding_WithInitialGuess
Figure 4.

CMakeLists.txt
Yurlungur: Can't hurt to document the location of the file.

singularity-eos/eos/eos_helmholtz.hpp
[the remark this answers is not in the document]
Yurlungur: Tmin and tmax are defined as constants right? We could slightly future proof this by setting ltguess to explicitly be the average of ltmin and ltmax.

test/test_eos_helmholtz.cpp
Yurlungur: I'm not sure why either test would work---you weren't testing on device, because you had normal for loops for the test code, not `portableFor` loops. I've fixed it, and tests now run on device.

Yurlungur:
From my perspective, this is ready to go. @mauneyc-LANL are you happy with the current cmake?

test/test_eos_helmholtz.cpp
AlexHls: Thanks for fixing that! I've included a fix to make the tests pass when not using Kokkos. If you could double check that, I'm not sure if I understood your intention for ``nwrong = 1`` correct. You set it to one such that in case the reduction fails the test fails, right? I'm not sure how the ``portableReduce`` behaves without Kokkos, but I assume it does not do anything in that case, i.e. there is no danger of setting ``nwrong = 0`` initially?

Yurlungur: No there's no danger---Kokkos sets it to zero before the reduction but I'm not sure that the cpu version does. Good fix setting it to zero initially

jhp-lanl (approved):

doc/sphinx/src/models.rst
```diff
@@ -1288,6 +1288,101 @@ indicates log base 10.
...
+ - the first adiabatic exponent (c&g 9.97)  
+ - the second adiabatic exponent (c&g 9.105)  
+ - the specific heat at constant pressure (c&g 9.98)  
+ - and relativistic formula for the sound speed (c&g 14.29)  
+
+
+.. _Timmes and Swesty: https://doi.org/10.1086/313304
+
+.. _Frank Timmes's website: https://cococubed.com/code_pages/eos.shtml
+
```
jhp-lanl, line 1385: Could you include the signature for the constructor like the other EOS?

jhp-lanl (changes requested):

jhp-lanl:
I requested changes but I'll just approve for now since I don't want to hold anything up if I'm not around

jhp-lanl (approved):

doc/sphinx/src/models.rst
AlexHls: Added a basic signature for the constructor.

CMakeLists.txt
```diff
@@ -69,6 +69,9 @@ cmake_dependent_option(
 cmake_dependent_option(
   SINGUALRITY_BUILD_STELLARCOLLAPSE2SPINER "Compile stellarcollapse2spiner" ON
   "SINGULARITY_USE_SPINER;SINGULARITY_USE_SPINER_WITH_HDF5" OFF)
+cmake_dependent_option(
+  SINGULARITY_USE_HELMHOLTZ "Include Helmholtz equation of state" OFF
+  "SINGULARITY_USE_SPINER;SINGULARITY_USE_SPINER_WITH_HDF5" OFF)
```
mauneyc-LANL, line 74: Just to be clear, you want this `OFF` by default _even if_ `spiner`+`hdf5` is enabled?

cmake/install.cmake
```diff
@@ -91,3 +91,14 @@ export(
...
+# ----------------------------------------------------------------------------#
+# Data files
+# ----------------------------------------------------------------------------#
+
+# install data files needed for various eos models
+if(SINGULARITY_USE_HELMHOLTZ)
+  # install data files
+  install(DIRECTORY ${PROJECT_SOURCE_DIR}/data/helmholtz
+          DESTINATION ${CMAKE_INSTALL_DATADIR}/singularity-eos)
+endif()
```
mauneyc-LANL, line 104: Raising a point here to check that this is correct - I need to look more closely but I'm worried that the install path will be `<install-prefix>/share/singularity-eos/helmholtz`, when the source expects `<install-prefix>/share/singularity-eos/data/helmholtz`

singularity-eos/CMakeLists.txt
```diff
@@ -32,6 +32,7 @@ set(EOS_HEADERS
     eos/eos_vinet.hpp
     eos/eos_builder.hpp
     eos/eos_jwl.hpp
+    eos/eos_helmholtz.hpp
```
mauneyc-LANL, line 35: Not an issue to resolve here, but I'm a little worried that we include headers that are not used with specific build parameters (e.g. `_USE_HELMHOLTZ` or `_USE_EOSPAC`). We should go through these in the future and guard their inclusion.

test/CMakeLists.txt
```diff
@@ -21,7 +21,13 @@ add_executable(
...
   test_eos_noble_abel.cpp
   test_eos_stiff.cpp
-)
+  test_eos_helmholtz.cpp
+  )
+
+if(SINGULARITY_TEST_HELMHOLTZ)
+    configure_file(${PROJECT_SOURCE_DIR}/data/helmholtz/helm_table.dat ${CMAKE_BINARY_DIR}/data/helmholtz/helm_table.dat COPYONLY)
+  target_compile_definitions(eos_unit_tests PRIVATE SINGULARITY_TEST_HELMHOLTZ SINGULARITY_USE_HELMHOLTZ)
+endif()
```
mauneyc-LANL, line 30: I see what's happening here, and don't object to holding up a merge, but this should be returned to.

`test_eos_helmholtz.cpp` should only be included in the `eos_unit_tests` source list if `_USE_HELMOLTZ` is on. The compile def here is unnecessary (aside from the conditional compilation of `test_eos_helmholtz.cpp`, which previous would make redundant). 

`_USE_HELMOLTZ` should be included in the target interface of `singularity-eos`, e.g.

```cmake
# defines
if(SINGULARITY_USE_SINGLE_LOGS)
  target_compile_definitions(singularity-eos PUBLIC SINGULARITY_USE_SINGLE_LOGS)
endif()
``` 

Relatedly, the tests in `test_eos_unit.cpp` related to `_TEST_SESAME` and `_TEST_EOSPAC` should be moved into separate source files.

mauneyc-LANL (changes requested):
Only one issue to address, everything else looks ok.

test/CMakeLists.txt
mauneyc-LANL: To clarify, the `_USE_HELMHOLTZ` compile option _should_ be changed (downstream codes will need to know that `_USE_HELMHOLTZ` is needed). Other issues can wait.

mauneyc-LANL: Ah wait, I got mixed up. Sorry - it is present as it should be. The second inclusion in the tests is what threw me.

CMakeLists.txt
Yurlungur: Yes that's the desired behaviour.

cmake/install.cmake
AlexHls: Your're right, that was an oversight by me. Fixed.

Yurlungur:
Triggered tests on re-git. Merging after they pass.

PR #266, opened 2023-07-06
Remove 'make install' to avoid confusion in build instructions

jhp-lanl:
## PR Summary

This is a very simple update to the documentation to remove `make install` since it seems to cause confusion.

I've left it in the python build instructions because I assume it's needed there to install the appropriate python bindings.

## PR Checklist

- N/A Format your changes by using the `make format` command after configuring with `cmake`.
- N/A After creating a pull request, note it in the CHANGELOG.md file
- N/A If preparing for a new release, update the version in cmake.

### DIFF
--- a/doc/sphinx/src/building.rst
+++ b/doc/sphinx/src/building.rst
@@ -95,7 +95,7 @@ At it's simplest, the cmake build process looks like this:
    mkdir bin
    cd bin
    cmake ..
-   make install
+   make
 
 You can set options on the cmake line via, e.g.,
 
@@ -192,7 +192,7 @@ Building ``singularity-eos`` with Python wrappers in a virtual environment:
    cd build
    cmake -DSINGULARITY_USE_HDF55=ON -DSINGULARITY_BUILD_PYTHON=ON -DCMAKE_INSTALL_PREFIX=$VIRTUAL_ENV ../..
    make -j
-   make install
+   make install  # Only install if you have the correct permissions
 
 
 Dependencies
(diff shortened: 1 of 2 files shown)


Yurlungur:
I'm confused. Why does `make install` not work? It's fine to remove it... but I think typing `make install` does work if you have the right permissions, or if you set 
```
-DCMAKE_INSTALL_PREFIX:PATH=/path/to/target/directory
```
so that you install to some user directory rather than globally.

rbberger:
How about adding it like this?
```bash
make
make install    # install into directory defined via CMAKE_INSTALL_PREFIX
```

Yurlungur:
:+1: for @rbberger 's suggestion

jhp-lanl:
> I'm confused. Why does `make install` not work? It's fine to remove it... but I think typing `make install` does work if you have the right permissions, or if you set
> 
> ```
> -DCMAKE_INSTALL_PREFIX:PATH=/path/to/target/directory
> ```
> 
> so that you install to some user directory rather than globally.

See, _that's_ not documented clearly and many people _don't_ have the right permissions which is why it's confusing. I'll follow Richard's suggestion

jhp-lanl:
@rbberger and @Yurlungur I added back in the `make install` command with the appropriate CMake variable in the comment. I also added that it's optional for the basic build procedure since it definitely is optional in most cases.

jhp-lanl:
FYI @gopsub although this is merged, feel free to comment on this if there is anything else confusing or unclear about `make install`, and we can add more changes in a future PR.

If there are other confusing aspects of the build process you would like us to document better, feel free to create an issue.

PR #274, opened 2023-07-12
Bjm/stiffgas

c0sm0-kramer:
## PR Summary

Added Stiffened Gas EoS according to reference Élaboration des lois d'état d'un liquide et de sa vapeur pour les modèles d'écoulements diphasiques, International Journal of Thermal Sciences, O Le Métayer and J Massoni and R Saurel, https://doi.org/10.1016/j.ijthermalsci.2003.09.002. Built with cmake 3.26.4 and make 3.82, compiled and tested with gcc 10.3.0.

This allows for the use of a stiffened gas EoS.

## PR Checklist

### DIFF
--- a/doc/sphinx/src/models.rst
+++ b/doc/sphinx/src/models.rst
@@ -448,8 +448,7 @@ The entropy for a stiffened gas can be expressed as
     S = C_V \ln\left(\frac{T}{T_0}\right) + (\gamma-1) C_V \ln\left(\frac{\rho_0}
      {\rho}\right) + q',
 
-we have assumed that the entropy and offset :math:`q'` are zero at the reference state given
-by :math:`T_0` and :math:`\rho_0`. By default, :math:`T_0 = 298` K and the
+where :math:`S(\rho_0,T_0)=q'`. By default, :math:`T_0 = 298` K and the
 reference density is given by
 
 .. math::
--- a/doc/sphinx/src/models.rst
+++ b/doc/sphinx/src/models.rst
@@ -425,6 +425,64 @@ these values are not set, they will be the same as those returned by the
 conditions are given, the return values of the :code:`ValuesAtReferenceState()`
 function will not be the same.
 
+Stiffened Gas
+`````````````
+
+The following implementation was guided by the reference `StiffGas`_. The stiffened gas model in ``singularity-eos`` takes
+the form
+
+.. math::
+
+    P = \Gamma \rho (e-q) - \gamma P_{\infty}
+
+.. math::
+
+    e = \frac{ P + \gamma P_{\infty} }{ P + P_{\infty} } C_V T + q,
+
+where quantities are largely the same as ideal gas with the exception of :math:`P_{\infty}` and :math:`q`, which can be thought of as an offset pressure and internal energy, respectively.
+
+The entropy for a stiffened gas can be expressed as
+
+.. math::
+
+    S = C_V \ln\left(\frac{T}{T_0}\right) + (\gamma-1) C_V \ln\left(\frac{\rho_0}
+     {\rho}\right) + q',
+
+we have assumed that the entropy and offset :math:`q'` are zero at the reference state given
+by :math:`T_0` and :math:`\rho_0`. By default, :math:`T_0 = 298` K and the
+reference density is given by
+
+.. math::
+
+    \rho_0 = \frac{P_0 + P_{\infty}}{(\gamma-1) C_V T_0},
+
+where the reference pressure, :math:`P_0`, is 1 bar by default.
+
+The settable parameters for this EOS are the Gruneisen parameter, :math:`\gamma-1`, specific
+heat capacity, :math:`C_V`, offset pressure, :math:`P_{\infty}`, and the offset internal energy, :math:`q`. Optionally, the reference state for the entropy calculation can
+be provided by setting values for the reference temperature, pressure, and offset entropy (:math:`q'`).
+
+The ``StiffGas`` EOS constructor has four arguments, ``gm1``, which is
+the Gruneisen parameter :math:`\Gamma`, ``Cv``, which is the
+specific heat :math:`C_V`, ``P_{\infty}`` which is the offset pressure, and ``q`` which is the offset internal energy:
+
+.. code-block:: cpp
+
+    StiffGas(Real gm1, Real Cv, Real Pinf, Real q)
+
+Optionally, the reference temperature and density for the entropy calculation,
+can be provided in the constructor via ``qp``, ``T0``, and ``P0``:
+
+.. code-block:: cpp
+
+    StiffGas(Real gm1, Real Cv, Real Pinf, Real q, Real qp, Real T0, Real P0)
+
+Note that these parameters are provided solely for the entropy calculation. When
+these values are not set, they will be the same as those returned by the
+:code:`ValuesAtReferenceState()` function. However, if the entropy reference
+conditions are given, the return values of the :code:`ValuesAtReferenceState()`
+function will not be the same.
+
 Gruneisen EOS
 `````````````
 
@@ -1164,60 +1222,4 @@ See :ref:`EOSPAC Vector Functions <eospac_vector>` for more details.
 
 .. _EOSPAC: https://laws.lanl.gov/projects/data/eos/eospacReleases.php
 
-Stiffened Gas
-`````````````
 
-The following implementation was guided by the reference `StiffGas`_. The stiffened gas model in ``singularity-eos`` takes
-the form
-
-.. math::
-
-    P = \Gamma \rho (e-q) - \gamma P_{\infty}
-
-.. math::
-
-    e = \frac{ P + \gamma P_{\infty} }{ P + P_{\infty} } C_V T + q,
-
-where quantities are largely the same as ideal gas with the exception of :math:`P_{\infty}` and :math:`q`, which can be thought of as an offset pressure and internal energy, respectively.
-
-The entropy for a stiffened gas can be expressed as
-
-.. math::
-
-    S = C_V \ln\left(\frac{T}{T_0}\right) + (\gamma-1) C_V \ln\left(\frac{\rho_0}
-     {\rho}\right) + q',
-
-we have assumed that the entropy and offset :math:`q'` are zero at the reference state given
-by :math:`T_0` and :math:`\rho_0`. By default, :math:`T_0 = 298` K and the
-reference density is given by
-
-.. math::
-
-    \rho_0 = \frac{P_0 + P_{\infty}}{(\gamma-1) C_V T_0},
-
-where the reference pressure, :math:`P_0`, is 1 bar by default.
-
-The settable parameters for this EOS are the Gruneisen parameter, :math:`\gamma-1`, specific
-heat capacity, :math:`C_V`, offset pressure, :math:`P_{\infty}`, and the offset internal energy, :math:`q`. Optionally, the reference state for the entropy calculation can
-be provided by setting values for the reference temperature, pressure, and offset entropy (:math:`q'`).
-
-The ``StiffGas`` EOS constructor has four arguments, ``gm1``, which is
-the Gruneisen parameter :math:`\Gamma`, ``Cv``, which is the
-specific heat :math:`C_V`, ``P_{\infty}`` which is the offset pressure, and ``q`` which is the offset internal energy:
-
-.. code-block:: cpp
-
-    StiffGas(Real gm1, Real Cv, Real Pinf, Real q)
-
-Optionally, the reference temperature and density for the entropy calculation,
-can be provided in the constructor via ``qp``, ``T0``, and ``P0``:
-
-.. code-block:: cpp
-
-    StiffGas(Real gm1, Real Cv, Real Pinf, Real q, Real qp, Real T0, Real P0)
-
-Note that these parameters are provided solely for the entropy calculation. When
-these values are not set, they will be the same as those returned by the
-:code:`ValuesAtReferenceState()` function. However, if the entropy reference
-conditions are given, the return values of the :code:`ValuesAtReferenceState()`
-function will not be the same.
--- a/doc/sphinx/src/models.rst
+++ b/doc/sphinx/src/models.rst
@@ -1178,7 +1178,7 @@ the form
 
     e = \frac{ P + \gamma P_{\infty} }{ P + P_{\infty} } C_V T + q,
 
-where quantities are largely the same as ideal gas with the exception of :math:`P_{\infty}` and :math:`q`, which can be thought of as an offset pressure and energy, respectively.
+where quantities are largely the same as ideal gas with the exception of :math:`P_{\infty}` and :math:`q`, which can be thought of as an offset pressure and internal energy, respectively.
 
 The entropy for a stiffened gas can be expressed as
 
@@ -1187,7 +1187,7 @@ The entropy for a stiffened gas can be expressed as
     S = C_V \ln\left(\frac{T}{T_0}\right) + (\gamma-1) C_V \ln\left(\frac{\rho_0}
      {\rho}\right) + q',
 
-we have assumed that the entropy is zero at the reference state given
+we have assumed that the entropy and offset :math:`q'` are zero at the reference state given
 by :math:`T_0` and :math:`\rho_0`. By default, :math:`T_0 = 298` K and the
 reference density is given by
 
@@ -1214,7 +1214,7 @@ can be provided in the constructor via ``qp``, ``T0``, and ``P0``:
 
 .. code-block:: cpp
 
-    IdealGas(Real gm1, Real Cv, Real Pinf, Real q, Real qp, Real T0, Real P0)
+    StiffGas(Real gm1, Real Cv, Real Pinf, Real q, Real qp, Real T0, Real P0)
 
 Note that these parameters are provided solely for the entropy calculation. When
 these values are not set, they will be the same as those returned by the
--- a/doc/sphinx/src/models.rst
+++ b/doc/sphinx/src/models.rst
@@ -15,6 +15,8 @@
 
 .. _WillsThermo: https://www.osti.gov/biblio/1561015
 
+.. _StiffGas: https://doi.org/10.1016/j.ijthermalsci.2003.09.002
+
 
 EOS Models
 ===========
@@ -1161,3 +1163,61 @@ See :ref:`EOSPAC Vector Functions <eospac_vector>` for more details.
 .. _Sesame: https://www.lanl.gov/org/ddste/aldsc/theoretical/physics-chemistry-materials/sesame-database.php
 
 .. _EOSPAC: https://laws.lanl.gov/projects/data/eos/eospacReleases.php
+
+Stiffened Gas
+`````````````
+
+The following implementation was guided by the reference `StiffGas`_. The stiffened gas model in ``singularity-eos`` takes
+the form
+
+.. math::
+
+    P = \Gamma \rho (e-q) - \gamma P_{\infty}
+
+.. math::
+
+    e = \frac{ P + \gamma P_{\infty} }{ P + P_{\infty} } C_V T + q,
+
+where quantities are largely the same as ideal gas with the exception of :math:`P_{\infty}` and :math:`q`, which can be thought of as an offset pressure and energy, respectively.
+
+The entropy for a stiffened gas can be expressed as
+
+.. math::
+
+    S = C_V \ln\left(\frac{T}{T_0}\right) + (\gamma-1) C_V \ln\left(\frac{\rho_0}
+     {\rho}\right) + q',
+
+we have assumed that the entropy is zero at the reference state given
+by :math:`T_0` and :math:`\rho_0`. By default, :math:`T_0 = 298` K and the
+reference density is given by
+
+.. math::
+
+    \rho_0 = \frac{P_0 + P_{\infty}}{(\gamma-1) C_V T_0},
+
+where the reference pressure, :math:`P_0`, is 1 bar by default.
+
+The settable parameters for this EOS are the Gruneisen parameter, :math:`\gamma-1`, specific
+heat capacity, :math:`C_V`, offset pressure, :math:`P_{\infty}`, and the offset internal energy, :math:`q`. Optionally, the reference state for the entropy calculation can
+be provided by setting values for the reference temperature, pressure, and offset entropy (:math:`q'`).
+
+The ``StiffGas`` EOS constructor has four arguments, ``gm1``, which is
+the Gruneisen parameter :math:`\Gamma`, ``Cv``, which is the
+specific heat :math:`C_V`, ``P_{\infty}`` which is the offset pressure, and ``q`` which is the offset internal energy:
+
+.. code-block:: cpp
+
+    StiffGas(Real gm1, Real Cv, Real Pinf, Real q)
+
+Optionally, the reference temperature and density for the entropy calculation,
+can be provided in the constructor via ``qp``, ``T0``, and ``P0``:
+
+.. code-block:: cpp
+
+    IdealGas(Real gm1, Real Cv, Real Pinf, Real q, Real qp, Real T0, Real P0)
+
+Note that these parameters are provided solely for the entropy calculation. When
+these values are not set, they will be the same as those returned by the
+:code:`ValuesAtReferenceState()` function. However, if the entropy reference
+conditions are given, the return values of the :code:`ValuesAtReferenceState()`
+function will not be the same.
--- a/singularity-eos/eos/eos.hpp
+++ b/singularity-eos/eos/eos.hpp
@@ -40,8 +40,8 @@
 #include <singularity-eos/eos/eos_jwl.hpp>
 #include <singularity-eos/eos/eos_spiner.hpp>
 #include <singularity-eos/eos/eos_stellar_collapse.hpp>
-#include <singularity-eos/eos/eos_vinet.hpp>
 #include <singularity-eos/eos/eos_stiff.hpp>
+#include <singularity-eos/eos/eos_vinet.hpp>
 
 // Modifiers
 #include <singularity-eos/eos/modifiers/eos_unitsystem.hpp>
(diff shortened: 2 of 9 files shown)


singularity-eos/eos/eos.hpp
```diff
@@ -63,7 +64,7 @@ using singularity::detail::transform_variadic_list;
 
 // all eos's
 static constexpr const auto full_eos_list =
-    tl<IdealGas, Gruneisen, Vinet, JWL, DavisReactants, DavisProducts
+    tl<IdealGas, Gruneisen, Vinet, JWL, DavisReactants, DavisProducts, StiffGas
```
Yurlungur, line 67: @jhp-lanl you mentioned this afternoon that we might want to move this EOS to be in the variant only optionally?

singularity-eos/eos/eos_stiff.hpp
```diff
@@ -0,0 +1,200 @@
...
+// Ports-of-call
+#include <ports-of-call/portability.hpp>
+
+// Base stuff
+#include <singularity-eos/base/constants.hpp>
+#include <singularity-eos/base/eos_error.hpp>
+#include <singularity-eos/base/robust_utils.hpp>
+#include <singularity-eos/eos/eos_base.hpp>
+
+#define MYMAX(a, b) a > b ? a : b
```
Yurlungur, line 32: fine to use this. It's also safe to use `std::max` now.

singularity-eos/eos/eos_stiff.hpp
```diff
@@ -0,0 +1,200 @@
...
+  }
+  PORTABLE_INLINE_FUNCTION void checkParams() const {
+    // Portable_require seems to do the opposite of what it should. Conditions
+    // reflect this and the code should be changed when ports-of-call changes
+    PORTABLE_ALWAYS_REQUIRE(_Cv >= 0, "Heat capacity must be positive");
+    PORTABLE_ALWAYS_REQUIRE(_gm1 >= 0, "Gruneisen parameter must be positive");
+  }
+  PORTABLE_INLINE_FUNCTION Real InternalEnergyFromDensityTemperature(
+      const Real rho, const Real temperature, Real *lambda = nullptr) const {
+    return MYMAX( 0.0, (rho*_Cv*temperature+_Pinf)/rho + _qq );
```
Yurlungur, line 67: you may want to replace `0` with `robust::small()` or something similar

singularity-eos/eos/eos_stiff.hpp
```diff
@@ -0,0 +1,200 @@
...
+      const Real rho, const Real temperature, Real *lambda = nullptr) const {
+    return MYMAX(0.0, _gm1 * rho * _Cv * temperature - _Pinf );
+  }
+  PORTABLE_INLINE_FUNCTION Real PressureFromDensityInternalEnergy(
+      const Real rho, const Real sie, Real *lambda = nullptr) const {
+    return MYMAX(0.0, _gm1 * rho * (sie-_qq) - (_gm1+1.0)*_Pinf );
+  }
+  PORTABLE_INLINE_FUNCTION Real EntropyFromDensityTemperature(
+      const Real rho, const Real temperature, Real *lambda = nullptr) const {
+    return _Cv * log( temperature / (pow(rho*_gm1*_Cv,_gm1)) ) + _qp;
```
Yurlungur, line 79: If there's any chance the denominator might go to zero even numerically, you should use `robust::ratio` instead of division.

Similarly, you may also wish to guard against `log(0)`. 

Also in C++ it's good practice to write `std::log` instead of `log`, though it's not a big deal either way.

Yurlungur (changes requested):
Overall looks pretty good! Thanks for adding this. A few minor nitpicks below. In addition, some bigger picture things:

1. We should test this on GPUs. the Darwin CI pipeline will do that for us, but if you have access to a GPU machine it's worth trying to run the tests there yourself too.
2. We try to add the PR to the `CHANGELOG.md` file
3. It would be good to add documentation about this EOS to the `models.rst` file in `doc/sphinx/src`. You can see how the others there are done. It doesn't have to be very extensive.
4. Looks like your code isn't formatted. We use `clang-format` to automatically format our code. If you have `clang-format` installed, you can run it on `singularity-eos` automatically in one of two ways: first, you can type `make format_singularity-eos` in your build directory. Alternatively, if that isn't working for you, there's a script in `utils/scripts` called `format.sh`. You can run it to format your code. I recommend running it as:
```bash
FMT=path/to/clang-format VERBOSE=1 utils/scripts/format.sh
```
where `FMT` points to the path of your `clang-format` executable.

singularity-eos/eos/eos_stiff.hpp
c0sm0-kramer: I see. Should I also use `std::pow` ? I'm not sure what the convention normally is in C++ on a lot of things. How would you recommend guarding against `log(0)`

singularity-eos/eos/eos_stiff.hpp
c0sm0-kramer: OK

singularity-eos/eos/eos_stiff.hpp
c0sm0-kramer: OK sounds good

singularity-eos/eos/eos_stiff.hpp
dholladay00: Recommend using the robust ratio function to accomplish the division.

singularity-eos/eos/eos_stiff.hpp
Yurlungur: Yeah `std::pow` over `pow` is good too. For guarding against `log(0)` it's enough to add a small number to the value you're taking the log of so it's never exactly zero.

dholladay00: One thing you'll learn about me is that I am not a fan of pow. I'd actually recommend trying replacing `log( temperature / (pow(rho*_gm1*_Cv,_gm1)) )` with `log( temperature) - _gm1* log(rho*_gm1*_Cv)`, 2 logs is likely much cheaper than 1 log and 1 pow. 

I'll be going through these expressions and making many similar suggestions. Please don't take it as criticism @c0sm0-kramer, I kinda like doing it and it ensures better performance.

singularity-eos/eos/eos_stiff.hpp
```diff
@@ -0,0 +1,200 @@
...
+  }
+  PORTABLE_INLINE_FUNCTION Real EntropyFromDensityInternalEnergy(
+      const Real rho, const Real sie, Real *lambda = nullptr) const {
+    //const Real temp = TemperatureFromDensityInternalEnergy(rho, sie, lambda);
+    //return EntropyFromDensityTemperature(rho, temp, lambda);
+    return _Cv * log( (rho*(sie-_qq)-_Pinf) /
+           (pow(_Cv*rho,_gm1+1.0)*pow(_gm1,_gm1)) ) +
+           _qp;
```
dholladay00, line 87: I think the most efficient expression would be: `_Cv * ( log( (rho*(sie-_qq)-_Pinf) / (_Cv*rho) /*use robust ratio I'm just lazy*/ ) - _gm1*log(_Cv*_gm1*rho) ) + _qp;`

singularity-eos/eos/eos_stiff.hpp
```diff
@@ -0,0 +1,200 @@
...
+  PORTABLE_INLINE_FUNCTION Real SpecificHeatFromDensityInternalEnergy(
+      const Real rho, const Real sie, Real *lambda = nullptr) const {
+    return _Cv;
+  }
+  PORTABLE_INLINE_FUNCTION Real BulkModulusFromDensityTemperature(
+      const Real rho, const Real temperature, Real *lambda = nullptr) const {
+    const Real entr = EntropyFromDensityTemperature(rho, temperature, lambda);
+    return MYMAX(0.0, pow(rho ,_gm1    ) * 
+                      pow(_Cv ,_gm1+1.0) * 
+                      pow(_gm1,_gm1+2.0) * exp((entr-_qp)/_Cv) );
```
dholladay00, line 102: I think an equivalent expression is `_gm1*_gm1*_Cv*exp(_gm1*log(rho*_Cv*_gm1) + (entr - _qp) / _Cv /*use robust ratio*/);`

singularity-eos/eos/eos_stiff.hpp
```diff
@@ -0,0 +1,200 @@
...
+    return 0;
+  }
+  static inline unsigned long max_scratch_size(unsigned int nelements) { return 0; }
+  PORTABLE_INLINE_FUNCTION void PrintParams() const {
+    printf("Stiff Gas Parameters:\nGamma = %g\nCv    = %g\nPinf  = %g\nq     = %g\nqprime= %g\n", _gm1 + 1.0, _Cv,_Pinf,_qq,_qp);
+  }
+  PORTABLE_INLINE_FUNCTION void
+  DensityEnergyFromPressureTemperature(const Real press, const Real temp, Real *lambda,
+                                       Real &rho, Real &sie) const {
+    sie = MYMAX(0.0, (press + (_gm1+1.0)*_Pinf)/(press + _Pinf)*_Cv*temp+_qq );
```
dholladay00, line 151: robust ratios for divide.

singularity-eos/eos/eos_stiff.hpp
jhp-lanl: Before we move too much further on this, we should ask _how_ we want to calculate entropy. If you use the integrals [here](https://lanl.github.io/singularity-eos/main/src/models.html#entropy-availability), you need to specify a reference density and temperature at which the entropy is defined to be zero.

This is how we have done it for the [ideal gas EOS](https://lanl.github.io/singularity-eos/main/src/models.html#ideal-gas), and actually the stiffened gas has the exact same entropy formula. My preference is towards defining a reference density and temperature rather than a reference entropy myself since the former are more meaningful.

I'll also add that you don't need the `pow` function if you use the formulation I linked above. Also you will need to specify a reference temperature and density in the input, but you won't have to specify `qp`.

singularity-eos/eos/eos_stiff.hpp
```diff
@@ -0,0 +1,200 @@
...
+    return _Cv * log( temperature / (pow(rho*_gm1*_Cv,_gm1)) ) + _qp;
+  }
+  PORTABLE_INLINE_FUNCTION Real EntropyFromDensityInternalEnergy(
+      const Real rho, const Real sie, Real *lambda = nullptr) const {
+    //const Real temp = TemperatureFromDensityInternalEnergy(rho, sie, lambda);
+    //return EntropyFromDensityTemperature(rho, temp, lambda);
+    return _Cv * log( (rho*(sie-_qq)-_Pinf) /
+           (pow(_Cv*rho,_gm1+1.0)*pow(_gm1,_gm1)) ) +
+           _qp;
+  }
```
jhp-lanl, line 88: It may be easier to just call the `TemperatureFromDensityInternalEnergy()` function and then call the `EntropyFromDensityTemperature()` function you defined above.

singularity-eos/eos/eos_stiff.hpp
```diff
@@ -0,0 +1,200 @@
...
+    return MYMAX(0.0, pow(rho ,_gm1    ) * 
+                      pow(_Cv ,_gm1+1.0) * 
+                      pow(_gm1,_gm1+2.0) * exp((entr-_qp)/_Cv) );
+  }
+  PORTABLE_INLINE_FUNCTION Real BulkModulusFromDensityInternalEnergy(
+      const Real rho, const Real sie, Real *lambda = nullptr) const {
+    const Real entr = EntropyFromDensityInternalEnergy(rho, sie, lambda);
+    return MYMAX(0.0, pow(rho ,_gm1    ) * 
+                      pow(_Cv ,_gm1+1.0) * 
+                      pow(_gm1,_gm1+2.0) * exp((entr-_qp)/_Cv) );
```
jhp-lanl, line 109: I'm worried about the numerics of the exponent and possible overflow/underflow. Can you derive a formula for the bulk modulus from the more general derivatives like we talked about? I.e.
```math
B_S = \rho \left(\frac{\partial P}{\partial \rho}\right)_e + \frac{P}{\rho} \left(\frac{\partial P}{\partial e}\right)_\rho 
```

singularity-eos/eos/eos_stiff.hpp
```diff
@@ -0,0 +1,200 @@
...
+                              Real &bmod, Real &dpde, Real &dvdt,
+                              Real *lambda = nullptr) const {
+    // use STP: 1 atmosphere, room temperature
+    rho = _rho0;
+    temp = _T0;
+    sie = _sie0;
+    press = _P0;
+    cv = _Cv;
+    bmod = _bmod0;
+    dpde = _dpde0;
```
jhp-lanl, line 133: I was skimming a bit, but are all these defined?

singularity-eos/eos/eos_stiff.hpp
jhp-lanl, line 151: We'll probably want to just force `press` to be larger than `_Pinf` here. Again, we don't want to floor the energy at zero either.

singularity-eos/eos/eos_stiff.hpp
```diff
@@ -0,0 +1,200 @@
...
+  }
+  static inline unsigned long max_scratch_size(unsigned int nelements) { return 0; }
+  PORTABLE_INLINE_FUNCTION void PrintParams() const {
+    printf("Stiff Gas Parameters:\nGamma = %g\nCv    = %g\nPinf  = %g\nq     = %g\nqprime= %g\n", _gm1 + 1.0, _Cv,_Pinf,_qq,_qp);
+  }
+  PORTABLE_INLINE_FUNCTION void
+  DensityEnergyFromPressureTemperature(const Real press, const Real temp, Real *lambda,
+                                       Real &rho, Real &sie) const {
+    sie = MYMAX(0.0, (press + (_gm1+1.0)*_Pinf)/(press + _Pinf)*_Cv*temp+_qq );
+    rho = MYMAX(0.0, (press + _Pinf) / (_gm1 * _Cv * temp) );
```
jhp-lanl, line 152: We'll want a robust ratio on this

singularity-eos/eos/eos_stiff.hpp
```diff
@@ -0,0 +1,200 @@
...
+    sie = (press+(_gm1+1.0)*_Pinf) / (_gm1 * rho) + _qq;
+  }
+  if (output & thermalqs::pressure) press = PressureFromDensityInternalEnergy(rho, sie);
+  if (output & thermalqs::temperature)
+    temp = TemperatureFromDensityInternalEnergy(rho, sie);
+  if (output & thermalqs::bulk_modulus)
+    bmod = BulkModulusFromDensityInternalEnergy(rho, sie);
+  if (output & thermalqs::specific_heat)
+    cv = SpecificHeatFromDensityInternalEnergy(rho, sie);
+}
```
jhp-lanl, line 195: We haven't been the best at ensuring that `FillEos` logic is sound throughout the code. Can you make sure that there aren't any edge cases where something weird could happen here?

jhp-lanl (changes requested):
Good start Ben!

We should decide how to do entropy. I'd prefer this to match the ideal gas case, but I'm open to arguments as to why this is better. Numerically, I think the ideal gas case is simpler though.

I'd like to see a few sanity tests like:

- Can you create an ideal gas EOS and a stiffened gas EOS with appropriate parameters and have them return the same result?
- Are the asymptotic limits correct? You'll need to address some of the numerical issues above (divide by zero, etc.), but for example $P(\rho, 0)$ or $P(0, T)$ should return $-P_\infty$.
- Check the bulk modulus against a finite difference of the pressure (see the Gruneisen EOS for an example)
- Use a finite difference to approximate $T\left(\frac{\partial S}{\partial T}\right_V$ and see if that's equal to the constant heat capacity.

singularity-eos/eos/eos_stiff.hpp
jhp-lanl: Just want to document here from other conversations with @c0sm0-kramer that a reference entropy might still be useful as well. I think it's just more useful to pair that with a specified reference state than to have a value that isn't as clearly tied to a reference state. We're still iterating on what this should look like

singularity-eos/eos/eos.hpp
c0sm0-kramer: 1. Testing still to be done on Darwin
2. Added PR to changelog
3. Documentation is still lacking, but a brief report is already done in LaTeX format - to be passed on or put in sphinx eventually.
4. Code formatted

singularity-eos/eos/eos_stiff.hpp
c0sm0-kramer: MYMAX replaced with std::max

singularity-eos/eos/eos_stiff.hpp
c0sm0-kramer: Corrected

singularity-eos/eos/eos_stiff.hpp
c0sm0-kramer: Entropy converted to a more general format like in the ideal gas case

c0sm0-kramer: Additional input is left to reference pressure/temperature and an entropy offset.

singularity-eos/eos/eos_stiff.hpp
c0sm0-kramer: Changed equation

singularity-eos/eos/eos_stiff.hpp
```diff
@@ -0,0 +1,200 @@
...
+        _sie0( (_P0 + (_gm1+1.0)*_Pinf)/(_P0 + _Pinf) * _Cv*_T0 + _qq ),
+        _entr0( _Cv * log(_T0/( pow(_rho0*_gm1*_Cv, _gm1) ) + _qp )),
+        _bmod0( pow(_rho0,_gm1)*pow(_Cv,_gm1+1.0)*pow(_gm1,_gm1+2.0)*exp((_entr0-_qp)/_Cv) ),
+        _dpde0( _rho0*_gm1 ) {
+    checkParams();
+  }
+  StiffGas GetOnDevice() { return *this; }
+  PORTABLE_INLINE_FUNCTION Real TemperatureFromDensityInternalEnergy(
+      const Real rho, const Real sie, Real *lambda = nullptr) const {
+    return MYMAX( 0.0, (rho*(sie-_qq)-_Pinf) / (rho*_Cv) );
```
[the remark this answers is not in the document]
c0sm0-kramer: Corrected

singularity-eos/eos/eos_stiff.hpp
c0sm0-kramer: Changed equation

singularity-eos/eos/eos_stiff.hpp
```diff
@@ -0,0 +1,200 @@
...
+      const Real rho, const Real temperature, Real *lambda = nullptr) const {
+    const Real entr = EntropyFromDensityTemperature(rho, temperature, lambda);
+    return MYMAX(0.0, pow(rho ,_gm1    ) * 
+                      pow(_Cv ,_gm1+1.0) * 
+                      pow(_gm1,_gm1+2.0) * exp((entr-_qp)/_Cv) );
+  }
+  PORTABLE_INLINE_FUNCTION Real BulkModulusFromDensityInternalEnergy(
+      const Real rho, const Real sie, Real *lambda = nullptr) const {
+    const Real entr = EntropyFromDensityInternalEnergy(rho, sie, lambda);
+    return MYMAX(0.0, pow(rho ,_gm1    ) *
```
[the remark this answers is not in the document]
c0sm0-kramer: Equation changed

singularity-eos/eos/eos_stiff.hpp
c0sm0-kramer: Corrected

singularity-eos/eos/eos_stiff.hpp
c0sm0-kramer: Equation changed

singularity-eos/eos/eos_stiff.hpp
c0sm0-kramer: Bulk modulus equation changed to the suggested

singularity-eos/eos/eos_stiff.hpp
c0sm0-kramer: These are all defined I believe

singularity-eos/eos/eos_stiff.hpp
c0sm0-kramer: Lower bound adjusted

singularity-eos/eos/eos_stiff.hpp
c0sm0-kramer: Corrected

singularity-eos/eos/eos_stiff.hpp
c0sm0-kramer: I think this fills the EoS properly

singularity-eos/eos/eos.hpp
jhp-lanl: I would propose we keep this in the typelist for the Variant since it's going to be used in xRAGE.

jhp-lanl:
@c0sm0-kramer after you add documentation I think this is good to go

test/test_eos_stiff.cpp
```diff
@@ -0,0 +1,303 @@
...
+        Kokkos::deep_copy(h_gruneisen, v_gruneisen);
+#endif // PORTABILITY_STRATEGY_KOKKOS
+        THEN("The returned Gamma(rho, e) should be equal to the true value") {
+          array_compare(num, density, energy, h_gruneisen, gruneisen_true, "Density",
+                        "Energy");
+        }
+      }
+    }
+  }
+}
```
jhp-lanl, line 303: @c0sm0-kramer can you add the following tests:

- [ ] Test against a finite difference approximation of the bulk modulus (i.e. the same formula but use pressure lookups to calculate the finite differences).
- [ ] Make sure that $S(\rho, e) == S(\rho, T(\rho, e)$, i.e. for a given energy and density test that `TemperatureFromDensityInternalEnergy()` followed by `EntropyFromDensityTemperature()` gives the same result as `EntropyFromDensityInternalEnergy()`.
- [ ] Test that you can set the `qq`, `Pinf`, and `qp` parameters equal to zero and recover the ideal gas behavior for at least some of the functions (as many as you're willing).

jhp-lanl (changes requested):

test/test_eos_stiff.cpp
c0sm0-kramer: @jhp-lanl tests now added

jhp-lanl (changes requested):
Looks like it's just about there. A few more changes and it should be good to go.

Ping me on mattermost when you make the changes and I'll run the CI on gitlab and get this merged when that passes

test/test_eos_stiff.cpp
```diff
@@ -0,0 +1,574 @@
...
+        Kokkos::deep_copy(h_entropy, v_entropy);
+#endif // PORTABILITY_STRATEGY_KOKKOS
+        THEN("The returned S(rho, e) should be equal to the true value") {
+          array_compare(num, density, energy, h_entropy, entropy_true, "Density",
+                        "Energy");
+        }
+      }
+      WHEN("A S(rho, T(rho,e)) lookup is performed") {
+        eos.TemperatureFromDensityInternalEnergy(v_density, v_energy, v_local_temp, num);
+        // printf("\nT_look = %g\n",v_local_temp[0]); // T_look returns the right value
```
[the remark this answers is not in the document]
c0sm0-kramer: Removed

test/test_eos_stiff.cpp
```diff
@@ -0,0 +1,574 @@
...
+      auto energy = Kokkos::create_mirror_view(v_energy);
+      Kokkos::deep_copy(density, v_density);
+      Kokkos::deep_copy(energy, v_energy);
+#endif // PORTABILITY_STRATEGY_KOKKOS
+
+      // Gold standard values for a subset of lookups
+      constexpr std::array<Real, num> pressure_true{2.0265000000000000e+06};
+      constexpr std::array<Real, num> bulkmodulus_true{2.6952450000000000e+06};
+      constexpr std::array<Real, num> temperature_true{4.0000000000000000e+02};
+      constexpr std::array<Real, num> gruneisen_true{0.33};
```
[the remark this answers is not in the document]
c0sm0-kramer: Changes made

jhp-lanl (approved):
Looks good to me! I'll run the CI on gitlab and then if that passes we just need @Yurlungur to approve and this is good to go. Thanks a lot @c0sm0-kramer !

jhp-lanl:
CI failed on GPUs because of missing `i` declaration. I've moved the declaration of `i` into the loops in the test so that it's more limited in scope.

@c0sm0-kramer all we need is documentation and this is good to go. Please add the EOS to the `models` section of the documentation.

doc/sphinx/src/models.rst
```diff
@@ -1161,3 +1163,61 @@ See :ref:`EOSPAC Vector Functions <eospac_vector>` for more details.
...
+where quantities are largely the same as ideal gas with the exception of :math:`P_{\infty}` and :math:`q`, which can be thought of as an offset pressure and internal energy, respectively.
+
+The entropy for a stiffened gas can be expressed as
+
+.. math::
+
+    S = C_V \ln\left(\frac{T}{T_0}\right) + (\gamma-1) C_V \ln\left(\frac{\rho_0}
+     {\rho}\right) + q',
+
+we have assumed that the entropy and offset :math:`q'` are zero at the reference state given
```
jhp-lanl, line 1190: I think you mean
```
where :math:`S(\rho_0, T_0) = q'`.
```
This is also less verbose and I think gets the point across.

jhp-lanl (approved):
Small nitpick: can you move the stiff gas EOS model up just after the ideal gas model in the documentation rather than at the end? Otherwise this looks good!

doc/sphinx/src/models.rst
c0sm0-kramer: Corrected

Yurlungur (approved):
This looks very good! Thanks for all your effort in pulling this through.

PR #288, opened 2023-08-04
Disable tests that use databox when spiner is off

mauneyc-LANL:
## PR Summary

@jhp-lanl hit an issue where `SINGULARITY_USE_SPINER=OFF` would fail tests that used `spiner`. In a way, the tests worked!

This disables building those tests when `SINGULARITY_USE_SPINER=OFF`

## PR Checklist

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -40,7 +40,7 @@ if(SINGULARITY_USE_EOSPAC AND SINGULARITY_TEST_SESAME)
                         PRIVATE Catch2::Catch2 singularity-eos::singularity-eos)
 endif()
 
-if(SINGULARITY_BUILD_CLOSURE)
+if(SINGULARITY_BUILD_CLOSURE AND SINGULARITY_USE_SPINER)
   add_executable(test_pte test_pte.cpp)
   target_link_libraries(test_pte PRIVATE Catch2::Catch2
                                          singularity-eos::singularity-eos)
@@ -85,8 +85,10 @@ if(SINGULARITY_BUILD_PYTHON AND SINGULARITY_TEST_PYTHON)
                                                  "${PYTHON_TEST_ENVIRONMENT}")
 endif()
 
-add_executable(profile_eos profile_eos.cpp)
-target_link_libraries(profile_eos singularity-eos::singularity-eos)
+if(SINGULARITY_USE_SPINER)
+  add_executable(profile_eos profile_eos.cpp)
+  target_link_libraries(profile_eos singularity-eos::singularity-eos)
+endif()
 
 if(SINGULARITY_USE_EOSPAC
    AND SINGULARITY_TEST_SESAME
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 - [[PR269]](https://github.com/lanl/singularity-eos/pull/269) Add SAP Polynomial EoS
 
 ### Fixed (Repair bugs, etc)
+- [[PR288]](https://github.com/lanl/singularity-eos/pull/288) Don't build tests that depend on spiner when spiner is disabled
 - [[PR287]](https://github.com/lanl/singularity-eos/pull/287) Fix testing logic with new HDF5 options
 - [[PR282]](https://github.com/lanl/singularity-eos/pull/282) Fix missing deep copy in sap polynomial tests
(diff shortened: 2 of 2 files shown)


jhp-lanl (approved):
I think it will take significant effort to make these tests run without databoxes. This seems like an appropriate near-term solution to me

Yurlungur (approved):

test/CMakeLists.txt
```diff
@@ -40,7 +40,7 @@ if(SINGULARITY_USE_EOSPAC AND SINGULARITY_TEST_SESAME)
                         PRIVATE Catch2::Catch2 singularity-eos::singularity-eos)
 endif()
 
-if(SINGULARITY_BUILD_CLOSURE)
+if(SINGULARITY_BUILD_CLOSURE AND SINGULARITY_USE_SPINER)
```
dholladay00, line 43: I think we should add fortran to the if list.

dholladay00:
> I think it will take significant effort to make these tests run without databoxes. This seems like an appropriate near-term solution to me

I'm not sure that I agree. `portablemdarray` could be dropped in w/o much additional effort I think.

test/CMakeLists.txt
mauneyc-LANL: The closure code is built independent of Fortran, `singularity-eos/CMakeLists.txt`

```cmake
if (SINGULARITY_BUILD_CLOSURE)
  list(APPEND EOS_SRCS eos/singularity_eos.cpp)
  list(APPEND EOS_HEADERS eos/singularity_eos.hpp closure/mixed_cell_models.hpp)
endif()

if (SINGULARITY_USE_FORTRAN)
  list(APPEND EOS_SRCS eos/singularity_eos.f90)
  # would rather handle this more robustly, being sloppy for now
  list(APPEND EOS_MODS singularity_eos.mod singularity_eos_types.mod)
endif()
```

jhp-lanl:
> > I think it will take significant effort to make these tests run without databoxes. This seems like an appropriate near-term solution to me
> 
> I'm not sure that I agree. `portablemdarray` could be dropped in w/o much additional effort I think.

That's good to hear then! I took a brief look at it and didn't know the code enough to make that assessment. I would propose making an issue to do that and move forward with this since it correctly aligns the code with its dependencies.

test/CMakeLists.txt
dholladay00: Yes, but you'd be saving me some time when I have to remember to do this in my PR b/c soon it'll only be built when also fortran.

Yurlungur:
> That's good to hear then! I took a brief look at it and didn't know the code enough to make that assessment. I would propose making an issue to do that and move forward with this since it correctly aligns the code with its dependencies.

Issue #289 . As I mention there and in mattermost, we may want to investigate moving some of the memory management niceties out of spiner and into ports of call. That might be good for the project as a whole anyway.

jhp-lanl:
> > That's good to hear then! I took a brief look at it and didn't know the code enough to make that assessment. I would propose making an issue to do that and move forward with this since it correctly aligns the code with its dependencies.
> 
> Issue #289 . As I mention there and in mattermost, we may want to investigate moving some of the memory management niceties out of spiner and into ports of call. That might be good for the project as a whole anyway.

I agree. I think that will be useful for other projects that depend on `ports_of_call`

test/CMakeLists.txt
dholladay00: After looking at it closer, you're correct, this test can remain but other things will need to be ifdef'd out.

dholladay00: Requested change rescinded.

dholladay00:
@Yurlungur, @jhp-lanl while all of that may be true. I am sensing scope creep here. I think the current tests could be implemented with portablemdarray as-is and that should be the focus for now.

jhp-lanl:
> @Yurlungur, @jhp-lanl while all of that may be true. I am sensing scope creep here. I think the current tests could be implemented with portablemdarray as-is and that should be the focus for now.

I'm not familiar enough with the MDArrays to be honest. But I still think this MR should go through in the meantime

Yurlungur:
> while all of that may be true. I am sensing scope creep here. I think the current tests could be implemented with portablemdarray as-is and that should be the focus for now.

I think we should just move forward with disabling these tests when spiner is disabled for now. And we can rethink/refactor later.

PR #290, opened 2023-08-07
Fix configuration export

mauneyc-LANL:
## PR Summary

@jhp-lanl found an issue where the exported `singularity-eosConfig.cmake` didn't have guards to check if a target has already been created for dependencies. This would cause an issue with builds where there are common dependencies downstream.

This update includes those guards, and also simplifies the logic.

## PR Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 - [[PR269]](https://github.com/lanl/singularity-eos/pull/269) Add SAP Polynomial EoS
 
 ### Fixed (Repair bugs, etc)
+- [[PR290]](https://github.com/lanl/singularity-eos/pull/290) Added target guards on export config
 - [[PR288]](https://github.com/lanl/singularity-eos/pull/288) Don't build tests that depend on spiner when spiner is disabled
 - [[PR287]](https://github.com/lanl/singularity-eos/pull/287) Fix testing logic with new HDF5 options
 - [[PR282]](https://github.com/lanl/singularity-eos/pull/282) Fix missing deep copy in sap polynomial tests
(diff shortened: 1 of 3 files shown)


Yurlungur (approved):

jhp-lanl:
It blows my mind that with how much boilerplate this has that there isn't already CMake functionality for this. And I don't want to block anything, but would it make sense for this to be a macro wrapping `find_package()`? I'm also fine with these changes as-is

jhp-lanl (approved):

mauneyc-LANL:
> It blows my mind that with how much boilerplate this has that there isn't already CMake functionality for this. And I don't want to block anything, but would it make sense for this to be a macro wrapping `find_package()`? I'm also fine with these changes as-is

Probably [this](https://cmake.org/cmake/help/latest/module/CMakeFindDependencyMacro.html) should be used. I can do this for this PR or later

mauneyc-LANL:
> @mauneyc-LANL do we need to run the CI on gitlab as well?

There's no test for this on the CI; there's been some discussion on doing this and maybe should plan for it. Locally I just copy the examples folder and have a CMake with `find_package(singularity-eos)`. That would be easy to get onto the CI i think

mauneyc-LANL:
@jhp-lanl I've updated to use [`find_dependency`](https://cmake.org/cmake/help/latest/module/CMakeFindDependencyMacro.html) (from comment above) that should avoid `find_package` collisions without doing a manual check (though if it still has issues then we can do manual checks).

jhp-lanl:
Okay the CMake configure stage seems to work now (which is where I was seeing the failure before). Merging

PR #291, opened 2023-08-09
Updates to package.py

mauneyc-LANL:
## PR Summary

- Add variants for `spiner`
- Refine some depends_on to not grab unnecessary packages
- Make sure CMake args are correct

@dholladay00 @jhp-lanl @jonahm-LANL @rbberger 
Comments/merges/changes. We can discuss more during scheduled meetings.

## PR Checklist

### DIFF
--- a/spack-repo/packages/singularity-eos/package.py
+++ b/spack-repo/packages/singularity-eos/package.py
@@ -58,10 +58,12 @@ class SingularityEos(CMakePackage, CudaPackage):
     # configuration 
     variant("hdf5", default=False, description="Use hdf5")
 
-    variant("spiner", default=False, description="Use Spiner")
+    variant("spiner", default=True, description="Use Spiner")
+
+    variant("closure", default=True, description="Build closure module")
 
     # building/testing/docs
-    depends_on("cmake@3.14:")
+    depends_on("cmake@3.19:")
     depends_on("catch2@2.13.7", when="+tests")
     depends_on("python@3:", when="+python")
     depends_on("py-numpy", when="+python+tests")
@@ -76,7 +78,7 @@ class SingularityEos(CMakePackage, CudaPackage):
     depends_on("eospac", when="+eospac")
 
     depends_on("ports-of-call@1.4.2:", when="@:1.7.0")
-    depends_on("ports-of-call@1.5.1:", when="@1.7.1:")
+    depends_on("ports-of-call@1.5.2:", when="@1.7.1:")
     # request HEAD of main branch
     depends_on("ports-of-call@main", when="@main")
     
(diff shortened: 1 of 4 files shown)


Yurlungur:
This all seems reasonable to me. Thanks for attacking this so quickly

mauneyc-LANL:
> This all seems reasonable to me. Thanks for attacking this so quickly

@Yurlungur I'd like to have a discussion/input on this before moving forward. Since we have a meet scheduled for day I think that'd be a good place to hammer this into shape.

Yurlungur:
Yep sounds good

jhp-lanl:
> This all seems reasonable to me. Thanks for attacking this so quickly

To echo Chris, yes please don't merge this yet. I plan to provide some input too

jhp-lanl:
Another issue I just found when trying to link with EAP: Cuda gets turned on by default in the CMake when Kokkos is on. I _think_ there isn't logic to allow building with Kokkos but without Cuda in the package.py right now. Please correct me if I'm wrong though

Yurlungur:
You can do `-DSINGULARITY_ENABLE_KOKKOS=ON -DSINGULARITY_ENABLE_CUDA=OFF` to override the cmake dependent option default settings. But I think the logic here probably should be reversed.

jhp-lanl:
> You can do `-DSINGULARITY_ENABLE_KOKKOS=ON -DSINGULARITY_ENABLE_CUDA=OFF` to override the cmake dependent option default settings. But I think the logic here probably should be reversed.

I'm fine with either to be honest as long as the hooks are there to ensure we can build without Cuda both through CMake (exists) and through spack. I think that logic doesn't exist in the `package.py` yet.

spack-repo/packages/singularity-eos/package.py
```diff
@@ -74,17 +78,17 @@ class SingularityEos(CMakePackage, CudaPackage):
     depends_on("eospac", when="+eospac")
 
     depends_on("ports-of-call@1.4.2:", when="@:1.7.0")
```
jhp-lanl, line 80: Do we want to specifically avoid 1.5.1 here since it has the inline issue? Something like
```suggestion
    depends_on("ports-of-call@1.4.2,1.5.2:", when="@:1.7.0")
```

jhp-lanl (approved):
One last request:

Can you update the build documentation for a few things:

1. Change "Overview" to something about building. "Overview" becomes the page title and it's a bit too broad.
2. Change spiner to an optional dependency and add something like "Optional; used for EOS that rely on databox or spiner"
3. Move `SINGULARITY_BUILD_CLOSURE` to the standard options rather than dependent options
```
``SINGULARITY_BUILD_CLOSURE``           ON       Build the mixed cell closure models
```
4. Reflect the fact that `SINGULARITY_USE_KOKKOSKERNELS` depends on `SINGULARITY_BUILD_CLOSURE` and not the other way around

Approving now so I don't forget to approve later

Yurlungur (approved):
Approved pending @jhp-lanl 's additional requests

mauneyc-LANL:
@jhp-lanl @jonahm-LANL Docs updated

jhp-lanl:
@dholladay00 or @rbberger if you have any comments can you make them by the end of today? Otherwise if not I think we can merge this.

rbberger (approved):
LGTM

jhp-lanl:
I'll go ahead and merge this in. If there are any changes you want @dholladay00 we can open up some issues

PR #293, opened 2023-08-17
Remove PTofRE implementations from source. Tests pass on my p9 darwin…

dholladay00:
… node.

## PR Summary
As discussed with @Yurlungur, this function was the sole reason for very excessive GPU compile times. The singular place it was used was replaced with an equivalent functionality. It does not appear to be used anywhere else. 

This PR removes `PTofRE` function. I'd like to hear from folks downstream, @jdolence @jhp-lanl do you foresee issues with the removal of this function to downstream use cases and/or future use cases (PT table equivalent for example @jhp-lanl?)

## PR Checklist

### DIFF
--- a/doc/sphinx/src/using-eos.rst
+++ b/doc/sphinx/src/using-eos.rst
@@ -549,10 +549,3 @@ returns the density at which pressure is minimized for a given
 temperature. This is again useful for root finds.
 
 Finally the method
-
-.. cpp:function:: void PTofRE(Real &rho, Real &sie, Real *lambda, Real &press, Real &temp, Real &dpdr, Real &dpde, Real &dtdr, Real &dtde) const;
-
-returns pressure and temperature, as well as the thermodynamic
-derivatives of pressure and temperature with respect to density and
-specific internal energy, as a function of density and specific
-internal energy.
--- a/doc/sphinx/src/using-eos.rst
+++ b/doc/sphinx/src/using-eos.rst
@@ -547,5 +547,3 @@ root finder to meaningful bound the root search. Similarly,
 
 returns the density at which pressure is minimized for a given
 temperature. This is again useful for root finds.
-
-Finally the method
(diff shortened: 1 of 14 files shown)


Yurlungur:
To add some context, this function was originally written as a convenience for the flag-type PTE solver. But the current solvers don't use this function, and the same functionality is available as a collection of calls, or through FillEos. 

While I believe we may need functions that compute quantities given Pressure as an input, this function doesn't provide it. So I see this as just removing code that (a) we don't need and shouldn't continue to maintain and (b) apparently slows compile times way down for some reason.

Yurlungur (approved):
The list of files changed doesn't seem like the entire list of Eoses and modifiers. Let's just make sure we got them all.

dholladay00:
> The list of files changed doesn't seem like the entire list of Eoses and modifiers. Let's just make sure we got them all.

I found things to remove via grep, so I think things that didn't include probably relied on the base implementation.

jhp-lanl:
Obviously not a blocker for this MR: I'd like to understand _why_ this was so bad at some point though so that we know what we can't feasibly do in the CRTP base class. Like is it a problem to call multiple class member functions there? Or is it a combination of that with the parallel for in the vector version? I suppose we'll find out whenever we try something that drastically increases compilation times 🙃

doc/sphinx/src/using-eos.rst
```diff
@@ -549,10 +549,3 @@ returns the density at which pressure is minimized for a given
 temperature. This is again useful for root finds.
 
 Finally the method
```
jhp-lanl, line 551:
```suggestion

```

jhp-lanl (approved):
I think you left a dangling remnant of the documentation. Otherwise looks good

Yurlungur:
@dholladay00 is this ready to go? I re-triggered the re-git pipelline

Yurlungur:
Tests passing. Pulling the trigger.

PR #295, opened 2023-08-19
Fast logs in stellar collapse

Yurlungur:
## PR Summary

For a while, I've been meaning to pull the fast logs code through the stellar collapse EOS model, as well as the singularity-opac models. This PR implements the first part of that. I thread fast logs through the stellar collapse eos. A few notes:

- `StellarCollapse` supports reading either `sp5` files or `h5` files in original stellar collapse format. I maintain that capability, and add machinery to re-interpolate the the stellar collapse files to the new fast log grid. That said, this means loading from h5 files will be even slower, and loading from sp5 files for production is highly recommended.
- I cleaned up the docs a little bit.
- The README was getting pretty out of sync with the build docs. I just removed all non-essential information.

## PR Checklist


Yurlungur:
FYI @brryan @AstroBarker

singularity-eos/base/fast-math/logs.hpp
```diff
@@ -67,11 +67,53 @@
...
+double fastpow2(const double x) {
+  // Magic numbers constexpr because C++ doesn't constexpr reinterpret casts
+  // these are floating point numbers as reinterpreted as integers.
+  // as_int(1.0)
+  constexpr long long int one_as_int = 4607182418800017408;
+  // as_int(2.0) - as_int(1.0)
+  constexpr double scale_up = 4503599627370496;
+  return as_double(static_cast<long long int>(x*scale_up) + one_as_int);
+}
+*/
```
brryan, line 115: Will these functions be uncommented later?

singularity-eos/eos/eos_stellar_collapse.hpp
```diff
@@ -158,6 +159,8 @@ class StellarCollapse : public EosBase<StellarCollapse> {
     return 0;
   }
   static inline unsigned long max_scratch_size(unsigned int nelements) { return 0; }
+  // NOTE: The logs here are table logs. SO when fast logs are
+  // enabled, these are fast logs.
```
brryan, line 163: Should not-really-log functions be private methods for internal lookups only? Pretty nasty bug if someone accidentally uses these elsewhere in their own code thinking they are logs.

doc/sphinx/src/models.rst
```diff
@@ -1277,10 +1277,13 @@ average atomic mass ``Abar`` and atomic number ``Zbar`` for heavy
...
-``TMin()``, ``TMax()``, ``YeMin()``, ``YeMax()``, ``sieMin()``, and
-``sieMax()``, which all return a ``Real`` number. The ``l`` prefix
-indicates log base 10.
+functions ``rhoMin()``, ``rhoMax()``, ``TMin()``, ``TMax()``,
+``YeMin()``, ``YeMax()``, ``sieMin()``, and ``sieMax()``, which all
+return a ``Real`` number.
+
+.. warning::
+    As with the SpinerEOS models, the stellar collapse models use fast
+    logs.
```
brryan, line 1286: Is fast log a configurable compile time option? If so should that option be documented here?

singularity-eos/eos/eos_stellar_collapse.hpp
```diff
@@ -819,23 +879,61 @@ inline void StellarCollapse::readBounds_(const hid_t &file_id, const std::string
...
+      }
+    }
+  }
+  // loop through a second time and copy
+  for (int i = 0; i < numYe_ * numT_ * numRho_; ++i) {
+    tab(i) = scratch(i);
+  }
+  tab.setRange(2, YeGrid_);
+  tab.setRange(1, lTGrid_);
+  tab.setRange(0, lRGrid_);
```
brryan, line 936: Does this function deserve a unit test? I'm assuming you already checked it for accuracy elsewhere

brryan (approved):
Some questions but I'll approve now.

AstroBarker (approved):
Thanks @Yurlungur ! LGTM

README.md
```diff
@@ -9,135 +9,6 @@ A collection of closure models and tools useful for multiphysics codes.
...
-| Quantity                  | Alternate names      | Units         |
-| ------------------------- | -------------------- | ------------- |
-| Mass density              | rho                  | g cm^-3       |
-| Temperature               | temperature          | K             |
-| Specific Internal Energy  | sie, Internal Energy | erg g^-1      |
-| Pressure                  |                      | erg cm^-3     |
-| Specific Heat             |                      | erg g^-1 K^-1 |
-| Bulk Modulus              |                      | erg cm^-3     |
-| Gruneisen Parameter       |                      |               |
-
```
dholladay00, line 140: Were these all out of date @Yurlungur ? In general it seems like a good idea to direct people to the docs rather than have to keep a readme and docs up to date.

doc/sphinx/src/getting-started.rst
```diff
@@ -11,12 +11,13 @@ At it's most basic, you can download and compile ``singularity-eos`` with:
   cd singularity-eos
   mkdir bin
   cd bin
-  cmake ..
+  cmake -DSINGULARITY_FORCE_SUBMODULEMODE=ON -DSINGULARITY_USE_FORTRAN=OFF ..
```
dholladay00, line 14: I think we should also direct to the location where more detailed configure/build information is supplied.

doc/sphinx/src/getting-started.rst
```diff
@@ -11,12 +11,13 @@ At it's most basic, you can download and compile ``singularity-eos`` with:
...
   make -j 
   make install # optional: install into directory defined via CMAKE_INSTALL_PREFIX
 
 This will downloady singularity-eos with no optional dependencies and
-compile the capabilities available in that form.
+compile the capabilities available in that form. For more details, see
+our build page.
```
dholladay00, line 20: Oh I see it now, could we provide a link to that page?

singularity-eos/base/fast-math/logs.hpp
dholladay00: I second this question. It is best practice not to include commented code unless there is a good reason to include it.

dholladay00:
How does this perform both speed/accuracy vs the previous implementation @Yurlungur ?

singularity-eos/base/fast-math/logs.hpp
Yurlungur: The integer aliasing code is faster---but I didn't feel confident uncommenting it now, until I have a better understanding of implications on all systems we care about. I'd also need to add single-precision overloads if needed, as the magic numbers for single-precision are different.

I can remove it again if y'all prefer.

singularity-eos/eos/eos_stellar_collapse.hpp
Yurlungur: I left these in by accident---meant to remove them. They're already gone from the docs.

doc/sphinx/src/models.rst
Yurlungur: It is. I'll add a comment to that effect.

singularity-eos/eos/eos_stellar_collapse.hpp
Yurlungur: You mean `setRange`? That's comprehensively tested in `spiner`.

README.md
Yurlungur: Some of them were not out of date, but I just figured keeping everything in one place is better.

doc/sphinx/src/getting-started.rst
Yurlungur: I'll add one

singularity-eos/base/fast-math/logs.hpp
brryan: I see. Interesting that it's faster -- what about uncommenting these functions but requiring that they be explicitly enabled by a compile time flag? `HIGH_RISK_MATH=ON` or something. That would also make it easier to test on platforms of interest. Could also for now add a check that forces failure if running in single precision mode.

mauneyc-LANL: why `KOKKOS_` and not `PORTABILITY_` ?

Yurlungur: > I see. Interesting that it's faster -- what about uncommenting these functions but requiring that they be explicitly enabled by a compile time flag? `HIGH_RISK_MATH=ON` or something. That would also make it easier to test on platforms of interest. Could also for now add a check that forces failure if running in single precision mode.

I can do that.

> why KOKKOS_ and not PORTABILITY_ ?

Because I screwed up when copy-pasting. I'll fix it.

singularity-eos/eos/eos_stellar_collapse.hpp
Yurlungur: Oh---no I haven't carefully tested it. It probably does deserve its own unit test, though.

singularity-eos/base/fast-math/logs.hpp
```diff
@@ -67,11 +67,53 @@
...
+}
+
+KOKKOS_FORCEINLINE_FUNCTION
+auto as_int(double f) {
+  return *reinterpret_cast<long long int*>(&f);
+}
+KOKKOS_FORCEINLINE_FUNCTION
+auto as_double(long long int i) {
+  return *reinterpret_cast<double*>(&i);
+}
```
mauneyc-LANL, line 92: (this isn't very significant) I was curious about this, took a look: https://godbolt.org/z/GWGdGGvKj

If there's something I'm missing I would be interested.

singularity-eos/base/fast-math/logs.hpp
Yurlungur: Adding `USE_HIGH_RISK_MATH` option and changing from `KOKKOS_` to `PORTABLE_`.

singularity-eos/eos/eos_stellar_collapse.hpp
Yurlungur: Added a test.

Yurlungur:
@dholladay00 @jhp-lanl @brryan I believe all comments have been addressed.

> How does this perform both speed/accuracy vs the previous implementation @Yurlungur ?

In principle, there's no error due to the fast log gridding. However, reinterpolation introduces an error proportional to 1/(ngrid points0^2. Performance improvement of P(rho, T) is a factor of a few, similar to what we saw with the SpinerEOS for sesame tables. T(rho, e) shows a 20% or so performance improvement.

dholladay00:
@Yurlungur it looks like the python tests are failing. Could be some of the stellarcollapse changes weren't propagated.

Yurlungur:
> @Yurlungur it looks like the python tests are failing. Could be some of the stellarcollapse changes weren't propagated.

Yep I see. That's exactly what it is. I'll take a look.

Yurlungur:
Thanks for the fix @dholladay00 ! I'll go ahead and merge this then.

PR #298, opened 2023-08-25
Robustify JWL EOS, using robust divide function for divisions. Added …

dholladay00:
…bounded argument exponential function to robust utils.

## PR Summary
Improve the robustness of JWL eos. Better checking to prevent FPE's. Added bounded argument exponentials to prevent under/over flow. 

## PR Checklist


dholladay00:
I've added asserts in the ctor, @jhp-lanl @Yurlungur could you let me know what the physical bounds on those parameters so that appropriately bounding checks can be added to the ctor.

singularity-eos/eos/eos_jwl.hpp
```diff
@@ -40,7 +41,16 @@ class JWL : public EosBase<JWL> {
   JWL() = default;
   PORTABLE_INLINE_FUNCTION JWL(const Real A, const Real B, const Real R1, const Real R2,
                                const Real w, const Real rho0, const Real Cv)
-      : _A(A), _B(B), _R1(R1), _R2(R2), _w(w), _rho0(rho0), _Cv(Cv) {}
+      : _A(A), _B(B), _R1(R1), _R2(R2), _w(w), _rho0(rho0), _Cv(Cv),
+        _c1(robust::ratio(A, rho0 * R1)), _c2(robust::ratio(B, rho0 * R2)) {
+    assert(A != 0.0);
+    assert(B != 0.0);
```
jhp-lanl, line 47: We don't need this restriction I don't think. Plus, these parameters can be negative so I think it's fine to leave them unconstrained.

singularity-eos/eos/eos_jwl.hpp
```diff
@@ -40,7 +41,16 @@ class JWL : public EosBase<JWL> {
...
+    assert(R1 != 0.0);
+    assert(R2 != 0.0);
+    assert(w != 0.0);
```
jhp-lanl, line 50: We should require `w > 0`

singularity-eos/eos/eos_jwl.hpp
```diff
@@ -40,7 +41,16 @@ class JWL : public EosBase<JWL> {
   JWL() = default;
   PORTABLE_INLINE_FUNCTION JWL(const Real A, const Real B, const Real R1, const Real R2,
                                const Real w, const Real rho0, const Real Cv)
-      : _A(A), _B(B), _R1(R1), _R2(R2), _w(w), _rho0(rho0), _Cv(Cv) {}
+      : _A(A), _B(B), _R1(R1), _R2(R2), _w(w), _rho0(rho0), _Cv(Cv),
+        _c1(robust::ratio(A, rho0 * R1)), _c2(robust::ratio(B, rho0 * R2)) {
+    assert(A != 0.0);
+    assert(B != 0.0);
+    assert(R1 != 0.0);
+    assert(R2 != 0.0);
```
jhp-lanl, line 49: I'm almost positive that these must be positive.

singularity-eos/eos/eos_jwl.hpp
```diff
@@ -96,22 +106,21 @@ class JWL : public EosBase<JWL> {
...
+  const Real x = robust::ratio(_rho0, rho);
+  return _A * robust::safe_arg_exp(-_R1 * x) + _B * robust::safe_arg_exp(-_R2 * x);
 }
-PORTABLE_INLINE_FUNCTION Real JWL::ReferenceEnergy(const Real rho) const {
-  const Real x = _rho0 / rho;
-  return _A / (_rho0 * _R1) * std::exp(-_R1 * x) +
-         _B / (_rho0 * _R2) * std::exp(-_R2 * x);
+PORTABLE_FORCEINLINE_FUNCTION Real JWL::ReferenceEnergy(const Real rho) const {
+  const Real x = robust::ratio(_rho0, rho);
+  return _c1 * robust::safe_arg_exp(-_R1 * x) + _c2 * robust::safe_arg_exp(-_R2 * x);
```
jhp-lanl, line 123: This equation (and the one for pressure) could still overflow if `robust::safe_arg_exp` returns `std::numeric_limits<T>::max()`. I see three ways to solve this:
1) Split the result and max at the numeric limit at each stage
2) Have the safe exponent return infinity and then add a ternary expression to return the max if infinity is reached
3) Test the arguments before putting them into the exponent and just return the maximum value if they're going to overflow.

jhp-lanl (changes requested):

singularity-eos/eos/eos_jwl.hpp
dholladay00: oof this opens up such a can of worms for every mathematical expression in the code. In a simple example, making use of infinity could be a way around FPE's. And all of this to *not* error out on a code that will pass garbage data with FPE trapping enabled. I think this is targeting a discussion much larger than the intended scope of this PR.

Yurlungur: why not just have the exp return `std::numeric_limits<T>::max() / 10` as the maximum allowed output? Doesn't solve the problem necessarily, but less likely to overflow/underflow at least.

dholladay00: I think returning infinity fixes it as well.

Yurlungur: Won't infinity trigger a floating point exception? If it doesn't, I agree---that's the better solution. It's what infinity is for.

singularity-eos/eos/eos_jwl.hpp
dholladay00: fixed

singularity-eos/eos/eos_jwl.hpp
jhp-lanl: My experience is more with the numpy implementation, but since the c++ standard and the numpy documentation both cite IEEE 754, I _think_ it shouldn't trigger FPEs. I'm not 100% sure about this though.

On this point, I remember trying to get compiler explore to let me trigger FPEs and I remember it being hard. Does anybody have a working example to do this?

jhp-lanl: > I think that is asking too much @jhp-lanl

I agree 😄 . Which is why I _think_ using infinity is the wise choice here. I _think_ infinity fails the IEEE normal check, but doesn't trigger FPEs, which I think is desirable here.

jhp-lanl (approved):

Yurlungur:
@dholladay00 where is this MR at? Should we merge it?

dholladay00:
We could, I think I did this to reduce FPE's in SAP. Perhaps it's good to do in general. I'm not opposed to merging it.

Yurlungur:
It's a pretty small changeset. I think we might as well.

dholladay00:
we should probably create an issue to robusify all analytic EOS's in a similar fashion.

Yurlungur:
> we should probably create an issue to robusify all analytic EOS's in a similar fashion.

#327

PR #304, opened 2023-09-12
Newton-Raphson root find for Helmholtz EOS

AlexHls:
## PR Summary

Adds a Newton-Raphson root find, specifically for use with the Helmholtz EOS. Although the current ``regula falsi`` produces on average more precise results, it is significantly more expensive. Moreover it is the method used in the originally published paper by Timmes.

<details>

<summary> Plot comparing both root finds </summary>

Overview

This benchmark was generated evaluating the EOS at random input values. It is probably not super scientific/ accurate, but should suffice the advantage of the Newton-Raphson root find.

</details>

A few points may (or may not) need to be addressed still:

- The current implementation expects a different function to be passed compared to the other root finds, i.e. it expects a tuple containing the derivative. This works straightforwards with the Helmholtz EOS, but might not work in general. However since this is only an internal function call this should be fine.
- I choose to make the root find method a runtime option, but I'm not sure if this is the way to go. So far I think it is better to have this as a runtime option so one can easily switch between the two methods, but I might be mistaken here.
- The default is now the Newton-Raphson root find, purely based on the better performance.

## PR Checklist

### DIFF
--- a/singularity-eos/base/root-finding-1d/root_finding.hpp
+++ b/singularity-eos/base/root-finding-1d/root_finding.hpp
@@ -246,7 +246,8 @@ PORTABLE_INLINE_FUNCTION Status newton_raphson(const T &f, const Real ytarget,
                                                const Real guess, const Real a,
                                                const Real b, const Real ytol, Real &xroot,
                                                const RootCounts *counts = nullptr,
-                                               const bool &verbose = false) {
+                                               const bool &verbose = false,
+                                               const bool &fail_on_bound_root = true) {
 
   constexpr int max_iter = NEWTON_RAPHSON_NITER_MAX;
   Real _x = guess;
@@ -273,13 +274,15 @@ PORTABLE_INLINE_FUNCTION Status newton_raphson(const T &f, const Real ytarget,
     // because one might want to handle this on a case by case basis
     // (e.g. if the boundary is a physical boundary, then one might want to
     // set the root to the boundary value).
-    // For this reason, we also do not return a status of FAIL in this case.
-    // Instead, we return a status of SUCCESS and let the caller decide what
-    // to do.
+    // Per default, we fail if the root is out of bounds controlled by
+    // fail_on_bound_root.
     if ((_x <= a && _xold <= a) || (_x >= b && _xold >= b)) {
       if (verbose) {
         printf("newton_raphson out of bounds! %.14e %.14e %.14e %.14e\n", ytarget, guess,
                a, b);
+      }
+      if (fail_on_bound_root) {
+        status = Status::FAIL;
       }
       break;
     }
--- a/singularity-eos/base/root-finding-1d/root_finding.hpp
+++ b/singularity-eos/base/root-finding-1d/root_finding.hpp
@@ -238,6 +238,9 @@ PORTABLE_INLINE_FUNCTION Status regula_falsi(const T &f, const Real ytarget,
 }
 
 // solves for f(x,params) - ytarget = 0
+// WARNING: this root finding expects a different callable f than the other
+// root finding methods. f should return a tuple of (f(x), f'(x)) where f'(x)
+// is the derivative of f with respect to x.
 template <typename T>
 PORTABLE_INLINE_FUNCTION Status newton_raphson(const T &f, const Real ytarget,
                                                const Real guess, const Real a,
--- a/singularity-eos/base/root-finding-1d/root_finding.hpp
+++ b/singularity-eos/base/root-finding-1d/root_finding.hpp
@@ -28,6 +28,7 @@
 #include <math.h>
 #include <ports-of-call/portability.hpp>
 #include <stdio.h>
+#include <tuple>
 
 #define SINGULARITY_ROOT_DEBUG (0)
 #define SINGULARITY_ROOT_VERBOSE (0)
@@ -38,6 +39,7 @@ namespace RootFinding1D {
 constexpr const int SECANT_NITER_MAX{1000};
 constexpr const int BISECT_NITER_MAX{1000};
 constexpr const int BISECT_REG_MAX{1000};
+constexpr const int NEWTON_RAPHSON_NITER_MAX{100};
 enum class Status { SUCCESS = 0, FAIL = 1 };
 
 /*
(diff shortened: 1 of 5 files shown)


jhp-lanl:
Is there a possibility that the Newton method won't converge? If so, it might make sense to structure the root find option as use RF is NR fails or if the user specifically requests _not_ to use NR.

AlexHls:
> Is there a possibility that the Newton method won't converge? If so, it might make sense to structure the root find option as use RF is NR fails or if the user specifically requests _not_ to use NR.

Thanks for the suggestion! I've added RF as fallback to the NR in cases the NR root find fails.

As far as convergence is concerned, this a bit more difficult to answer. In my naive test program, RF was actually more likely to not converge, but I know that in astrophysical simulations the NR also has trouble converging in certain parameter spaces. The question is if the RF would actually converge better in those cases (e.g. there is nothing that can be done once the root find runs into the edges of the underlying table) and, more importantly, if the non-convergence actually matters in these cases (vs the increased computational cost). To my knowledge, most astrophysical simulations have 'ignored' this and just bound the result of the root find back to the tabulated values. I might be mistaken though, but I guess this is something that one would first need to investigate extensively.

For now I have implemented RF as a fallback if NR does not converge, in my opinion this is a sensible default.

jhp-lanl:
> As far as convergence is concerned, this a bit more difficult to answer. In my naive test program, RF was actually more likely to not converge, but I know that in astrophysical simulations the NR also has trouble converging in certain parameter spaces. The question is if the RF would actually converge better in those cases (e.g. there is nothing that can be done once the root find runs into the edges of the underlying table) and, more importantly, if the non-convergence actually matters in these cases (vs the increased computational cost). To my knowledge, most astrophysical simulations have 'ignored' this and just bound the result of the root find back to the tabulated values. I might be mistaken though, but I guess this is something that one would first need to investigate extensively.

Interesting! I suppose I'm used to using the RF method on problems that have an easy way to bracket the solution a priori via some known bounds. In those cases, the RF method is guaranteed to converge albeit slowly sometimes. I assume that when you say that RF didn't converge, it means that you were unable to bracket the solution? Or was it simply converging so slowly that it ran out of iterations?

Another thing we could consider in the future is to improve the RF method to use either the Pegasus or Illinois algorithms (http://paulklein.se/newsite/teaching/rootfinding.pdf) to get superlinear convergence.

singularity-eos/base/root-finding-1d/root_finding.hpp
```diff
@@ -235,6 +237,71 @@ PORTABLE_INLINE_FUNCTION Status regula_falsi(const T &f, const Real ytarget,
   return status;
 }
 
+// solves for f(x,params) - ytarget = 0
+template <typename T>
+PORTABLE_INLINE_FUNCTION Status newton_raphson(const T &f, const Real ytarget,
+                                               const Real guess, const Real a,
+                                               const Real b, const Real ytol, Real &xroot,
+                                               const RootCounts *counts = nullptr,
+                                               const bool &verbose = false) {
```
jhp-lanl, line 246: I think it's okay to use the same function signature as the regula falsi method, but I think you should at least add a comment about the behavior of the callable `f` being different (i.e. needing to return both function and derivative). Anybody using it now will quickly run into a compile time error if they try to just return the function, but why not save them the work of compiling to figure that out?

In the future I could see possibly wanting to develop an overload for this signature that uses the same behavior for `f` as the regula falsi and then wraps that callable in a function to perform a finite difference of some sort when an analytic derivative isn't available. Regardless of whether that happens though, comments will always help in understanding the subtle differences in template behavior between the root finding methods.

singularity-eos/base/root-finding-1d/root_finding.hpp
```diff
@@ -235,6 +237,71 @@ PORTABLE_INLINE_FUNCTION Status regula_falsi(const T &f, const Real ytarget,
...
+      }
+      break;
+    }
+    if (_x >= b && _xold >= b) {
+      if (verbose) {
+        printf("newton_raphson out of bounds! %.14e %.14e %.14e %.14e\n", ytarget, guess,
+               a, b);
+      }
+      break;
+    }
```
jhp-lanl, line 282: Maybe I'm missing the difference, but it looks like these conditions could be combined to make this code more DRY

singularity-eos/base/root-finding-1d/root_finding.hpp
```diff
@@ -235,6 +237,71 @@ PORTABLE_INLINE_FUNCTION Status regula_falsi(const T &f, const Real ytarget,
...
+    _xold = _x;
+    _x = _x - (yg - ytarget) / dfunc;
+
+    // check if we are out of bounds
+    if (_x <= a && _xold <= a) {
+      if (verbose) {
+        printf("newton_raphson out of bounds! %.14e %.14e %.14e %.14e\n", ytarget, guess,
+               a, b);
+      }
+      break;
```
jhp-lanl, line 274: Should being out of bounds return a `SUCCESS` status?

singularity-eos/base/root-finding-1d/root_finding.hpp
Yurlungur: I agree with @jhp-lanl here

singularity-eos/base/root-finding-1d/root_finding.hpp
```diff
@@ -235,6 +237,71 @@ PORTABLE_INLINE_FUNCTION Status regula_falsi(const T &f, const Real ytarget,
...
+    // not converged; compute the next step
+    _xold = _x;
+    _x = _x - (yg - ytarget) / dfunc;
+
+    // check if we are out of bounds
+    if (_x <= a && _xold <= a) {
+      if (verbose) {
+        printf("newton_raphson out of bounds! %.14e %.14e %.14e %.14e\n", ytarget, guess,
+               a, b);
+      }
```
Yurlungur, line 273: This is fine for now, but I strongly suspect this can be made much more robust with, e.g., a line search, or even a simple bounds enforcement. NR is prone to overshooting, and simply forcing it back in the domain is often enough to get it to converge.

Yurlungur (commented):
I approve. A few minor nitpicks/comments below that match @jhp-lanl 's thoughts.

singularity-eos/base/root-finding-1d/root_finding.hpp
AlexHls: Thanks for the catch! This is a bit of a tricky one. In the original reference implementation, in each condition the root value was set to the appropriate boundary value, hence the two conditions. This has been moved to the Helmholtz call itself (since it is also necessary for the RF roots). I'm not sure if this is the smart way to do, i.e. if this bounding should be handled by the root finding methods or by the function calling the root find. For now I have decided that this should be handled by the function calling the root find (since one might not always want to bound the value to the boundary values) and added a comment in the appropriate location.

This is also the reason that this out-of-bounds case does not return a FAILURE status. One could argue that this is not really a successful root, but at least in the context of the Helmholtz EOS a root bound to the edges of the table is (at least in the reference implementation) taken as a success. Most likely this is a peculiarity for the Helmholtz EOS due to the fine tuning and won't translate to other application. I am however hesitant to implement a FAILURE return here since this would trigger an additional root find using RF. The bound result should be good enough, at least good enough as to not trigger a new attempt.

singularity-eos/base/root-finding-1d/root_finding.hpp
AlexHls: I agree. There is most likely several ways to improve the convergence properties. For now the point is more to implement the NR as given by the reference implementation. I'm planning to conduct some more extensive benchmarking in the future to establish the performance vs. convergence in this context (since I suspect that this implementation is tuned for performance with just "good enough" convergence properties), but there is no timeline yet.

singularity-eos/base/root-finding-1d/root_finding.hpp
AlexHls: Fair point. Thanks for the suggestion! I've added the relevant option to the root find.

jhp-lanl (approved):
I approve, but @Yurlungur should take a final look before this gets merged.

Yurlungur (approved):

PR #306, opened 2023-10-09
Add a generic evaluator function

jdolence:
## PR Summary
As discussed with @Yurlungur...

This is a tiny PR that adds a fair bit more flexibility to how downstream codes interact with singularity-eos.  The new functionality is a new `Evaluate` template function for both `EosBase` and `Variant` that can take in a downstream-defined functor and evaluate it with the underlying eos object.  This allows one to, for example, express a parallel loop that calls several EOS functions at once in the downstream code, and then have singularity execute that parallel loop with one of the basic EOSs (as opposed to having to visit the variant in every call).

The current vector interface is quite constraining.  For example, if you're running on GPU, it assumes you're calling from the host and that you want to launch a new kernel to make the eos call.  In our downstream code, we use hierarchical parallelism where what we really want is to have an inner loop inside our kernel that calls the EOSs.  This new feature allows us to write our inner loop as we do all our other inner loops, pass it in to singularity, and have it execute as we expressed but using the CRTP stuff to avoid many calls to `mpark::visit`.

I think the python interface requires something like the existing vector interface, otherwise I would argue that this is much more flexible and may be preferred in general for interfacing with downstream codes.  The only downside is that it requires the downstream code to write a functor (or some number of functors to capture all the EOS calling patterns).

## PR Checklist

### DIFF
--- a/doc/sphinx/src/using-eos.rst
+++ b/doc/sphinx/src/using-eos.rst
@@ -248,6 +248,13 @@ desired use-case.
   support templates, and the ``operator()`` call must be
   templated.
 
+.. warning::
+
+  It can be dangerous to use functors with side-effects. Especially
+  with GPUs it can produce very unintuitive behaviour. We recommend
+  you only make the ``operator()`` non-const if you really know what
+  you're doing.
+
 To see the utlity of the ``Evaluate`` function, it's probably just
 easiest to provide an example. The following code evaluates the EOS on
 device and compares against a tabulated pressure. The total difference
--- a/doc/sphinx/src/using-eos.rst
+++ b/doc/sphinx/src/using-eos.rst
@@ -35,6 +35,10 @@ conditions. The type of parallelism used depends on how
 ``singularity-eos`` is compiled. If the ``Kokkos`` backend is used,
 any parallel dispatch supported by ``Kokkos`` is supported.
 
+A more generic version of the vector calls exists in the ``Evaluate``
+method, which allows the user to specify arbitrary parallel dispatch
+models by writing their own loops. See the relevant section below.
+
 .. _variant section:
 
 Variants
--- a/doc/sphinx/src/using-eos.rst
+++ b/doc/sphinx/src/using-eos.rst
@@ -216,6 +216,92 @@ available member function.
    eos.TemperatureFromDensityInternalEnergy(density.data(), energy.data(), temperature.data(),
                                             scratch.data(), density.size());
 
+The Evaluate Method
+~~~~~~~~~~~~~~~~~~~
+
+A special call related to the vector calls is the ``Evaluate``
+method. The ``Evaluate`` method requests the EOS object to evaluate
+almost arbitrary code, but in a way where the type of the underlying
+EOS object is resolved *before* this arbitrary code is evaluated. This
+means the code required to resolve the type of the variant is only
+executed *once* per ``Evaluate`` call. This can enable composite EOS
+calls, non-standard vector calls, and vector calls with non-standard
+loop structure.
+
+The ``Evaluate`` call has the signature
+
+.. code-block:: cpp
+
+  template<typename Functor_t>
+  PORTABLE_INLINE_FUNCTION
+  void Evaluate(Functor_t f);
+
+where a ``Functor_t`` is a class that *must* provide a ``void
+operator() const`` method templated on EOS type. ``Evaluate`` is decorated
+so that it may be evaluated on either host or device, depending on
+desired use-case.
+
+.. note::
+
+  Note that for C++ versions less than C++20, the functor must be
+  defined as a class. Anonymous functions won't do, as they do not
+  support templates, and the ``operator()`` call must be
+  templated.
+
+To see the utlity of the ``Evaluate`` function, it's probably just
+easiest to provide an example. The following code evaluates the EOS on
+device and compares against a tabulated pressure. The total difference
+is summed using the ``Kokkos::parallel_reduce`` functionality in the
+``Kokkos`` performance portability library.
+
+.. code-block:: cpp
+
+  // The functor we use is defined here.
+  // This class definition needs to be of appropriately global scope.
+  class CheckPofRE {
+   public:
+    CheckPofRE(Real *P, Real *rho, Real *sie, int N) : P_(P), rho_(rho), sie_(sie), N_(N) {}
+    template <typename T>
+    // this is a host-only call, but if you wanted to write
+    // a function that you wanted to evaluate on device
+    // you could add the
+    // PORTABLE_INLINE_FUNCTION
+    // decorator here.
+    void operator()(const T &eos) const {
+      Real *P = P_; // gets around warnings regarding capture of "this" pointer
+      Real *rho = rho_;
+      Real *sie = sie_;
+      Kokkos::parallel_reduce(
+          "MyCheckPofRE", N_,
+          KOKKOS_LAMBDA(const int i, Real &diff) {
+            diff += std::abs(P[i] - eos.PressureFromDensityInternalEnergy(rho[i], sie[i]));
+          },
+          tot_diff);
+      std::cout << "Total difference = " << tot_diff << std::endl;
+    }
+  
+   private:
+    int N_;
+    Real *P_;
+    Real *rho_;
+    Real *sie_;
+  };
+
+  // Here we construct our functor
+  // it is assumed the pointers to device memory P, rho, sie, are defined elsewhere.
+  CheckPofRE my_op(P, rho, sie, N);
+
+  // Here we call the evaluate function
+  eos.Evaluate(my_op);
+
+  // The above two lines could have been called "in-one" with:
+  // eos.Evaluate(CheckPofRE(P, rho, sie, N));
+
+This is not functionality that would be available with the standard
+vector calls provided by ``singularity-eos``, at least not without
+chaining multiple parallel dispatch calls. Here we can do it in a
+single call.
+
 Lambdas and Optional Parameters
 --------------------------------
 
--- a/doc/sphinx/src/using-eos.rst
+++ b/doc/sphinx/src/using-eos.rst
@@ -244,7 +244,8 @@ function with an `auto` argument as the input, e.g.,
 
 .. code-block::
 
-   eos.Evaluate([=](auto eos) { /* my code snippet */ });
+   // equivalent to [=], but with device markings
+   eos.Evaluate(PORTABLE_LAMBDA(auto eos) { /* my code snippet */ });
 
 .. warning::
 
--- a/doc/sphinx/src/using-eos.rst
+++ b/doc/sphinx/src/using-eos.rst
@@ -278,6 +278,9 @@ is summed using the ``Kokkos::parallel_reduce`` functionality in the
       Real *P = P_; // gets around warnings regarding capture of "this" pointer
       Real *rho = rho_;
       Real *sie = sie_;
+      // reduction target
+      Real tot_diff;
+      // reduction op
       Kokkos::parallel_reduce(
           "MyCheckPofRE", N_,
           KOKKOS_LAMBDA(const int i, Real &diff) {
--- a/doc/sphinx/src/using-eos.rst
+++ b/doc/sphinx/src/using-eos.rst
@@ -275,16 +275,13 @@ is summed using the ``Kokkos::parallel_reduce`` functionality in the
     // PORTABLE_INLINE_FUNCTION
     // decorator here.
     void operator()(const T &eos) const {
-      Real *P = P_; // gets around warnings regarding capture of "this" pointer
-      Real *rho = rho_;
-      Real *sie = sie_;
       // reduction target
       Real tot_diff;
       // reduction op
       Kokkos::parallel_reduce(
           "MyCheckPofRE", N_,
           KOKKOS_LAMBDA(const int i, Real &diff) {
-            diff += std::abs(P[i] - eos.PressureFromDensityInternalEnergy(rho[i], sie[i]));
+            diff += std::abs(P_[i] - eos.PressureFromDensityInternalEnergy(rho_[i], sie_[i]));
           },
           tot_diff);
       std::cout << "Total difference = " << tot_diff << std::endl;
--- a/doc/sphinx/src/using-eos.rst
+++ b/doc/sphinx/src/using-eos.rst
@@ -237,23 +237,22 @@ The ``Evaluate`` call has the signature
   void Evaluate(Functor_t f);
 
 where a ``Functor_t`` is a class that *must* provide a ``void
-operator() const`` method templated on EOS type. ``Evaluate`` is decorated
-so that it may be evaluated on either host or device, depending on
-desired use-case.
+operator() const`` method templated on EOS type. ``Evaluate`` is
+decorated so that it may be evaluated on either host or device,
+depending on desired use-case. Alternatively, you may use an anonymous
+function with an `auto` argument as the input, e.g.,
 
-.. note::
+.. code-block::
 
-  Note that for C++ versions less than C++20, the functor must be
-  defined as a class. Anonymous functions won't do, as they do not
-  support templates, and the ``operator()`` call must be
-  templated.
+   eos.Evaluate([=](auto eos) { /* my code snippet */ });
 
 .. warning::
 
   It can be dangerous to use functors with side-effects. Especially
   with GPUs it can produce very unintuitive behaviour. We recommend
   you only make the ``operator()`` non-const if you really know what
-  you're doing.
+  you're doing. And in the anonymous function case, we recommend you
+  capture by value, not reference.
 
 To see the utlity of the ``Evaluate`` function, it's probably just
 easiest to provide an example. The following code evaluates the EOS on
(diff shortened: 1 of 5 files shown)


jhp-lanl (approved):
I approve, but I have some questions and a comment:

If I'm understanding correctly, the main advantage here is that you can string together an arbitrary number of EOS lookup functions under a single `mpark::visit()`, right? This begs the question, is the `mpark::visit()` function that impactful for performance? It doesn't get optimized away? Have you profiled this change?

As a tangent in defense of the current vector interface, it was never intended to be called on device. The predominant wisdom at the time it was written was that the only real way that the EOS would be called from device would be via the scalar interface and that the host could chain together whatever calls it needed. This duplicates the `visit()` calls, yes, but otherwise allows the flexible parallelism you mention in your PR description.

EDIT - I may have come off too critical. I think this is a very useful addition and could be powerful in a number of different scenarios.

jhp-lanl:
As an aside, does this functionality eliminate the need for `FillEOS()` since you can basically inject it now?

singularity-eos/eos/eos_base.hpp
```diff
@@ -128,6 +128,13 @@ class EosBase {
   struct is_raw_pointer
       : std::is_same<std::remove_reference_t<std::remove_cv_t<T>>, R *> {};
 
+  // Generic evaluator
```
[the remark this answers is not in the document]
dholladay00: I was curious about this one as well.

jhp-lanl:
> I think the python interface requires something like the existing vector interface, otherwise I would argue that this is much more flexible and may be preferred in general for interfacing with downstream codes. The only downside is that it requires the downstream code to write a functor (or some number of functors to capture all the EOS calling patterns).

I know this isn't what you were intending here, but if you extend the existing vector interface to have a vector version of `Evaluate()`, you could also think about using the python interface to inject arbitrary functions as well.

The other downside of this approach is that a fortran-based host code would have to provide its own C++ interface, which isn't something we're currently doing.

dholladay00:
> I approve, but I have some questions and a comment:
> 
> If I'm understanding correctly, the main advantage here is that you can string together an arbitrary number of EOS lookup functions under a single `mpark::visit()`, right? This begs the question, is the `mpark::visit()` function that impactful for performance? It doesn't get optimized away? Have you profiled this change?
> 
> As a tangent in defense of the current vector interface, it was never intended to be called on device. The predominant wisdom at the time it was written was that the only real way that the EOS would be called from device would be via the scalar interface and that the host could chain together whatever calls it needed. This duplicates the `visit()` calls, yes, but otherwise allows the flexible parallelism you mention in your PR description.

My understanding is that the visit removes the possibility of vectorization. This lifts the visit higher up and thus in a lower level loop vectorization is possible. This basically allows downstream codes to write functors that can get at the underlying eos type and use that directly similar to how the vector interface accomplishes this. @jdolence has found this to have noticeable speedup. I am not clear on the main contributor to the speedup, that would be helpful to know though. 

As for the vector interface, it is still necessary and helpful as it provides concrete functions that can be called from fortran.

singularity-eos/eos/eos_base.hpp
```diff
@@ -128,6 +128,13 @@ class EosBase {
...
+  template <typename Functor_t>
+  PORTABLE_INLINE_FUNCTION void Evaluate(const Functor_t &f) const {
+    CRTP copy = *(static_cast<CRTP const *>(this));
+    f(copy);
```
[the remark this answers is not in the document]
Yurlungur: I'll just answer for Josh, since he and I discussed this design before he implemented. Yes--the functor handles everything. I'm imagining a call pattern that looks something like this:

```C++
struct MyPofRhoT {
  MyPofRhoT(Real *rho_, Real *T_, Real *P_, int N_)
    : rho(rho_), T(T_), P(P_), N(N_) {}
  
  template <typename CRTP>
  void operator(CRTP crtp) {
    Kokkos::parallel_for(N, KOKKOS_LAMBDA(int i) {
      P[i] = crtp.PressureFromDensityTemperature(rho[i], T[i], lambda[i]);
    });
  }

  int N;
  Real *rho;
  Real *T;
  Real *P
  LambdaIndexer lambda = NullIndexr; // or something
};
```

And then you could call it like
```C++
eos.Evaluate(MyPofRhoT(rho, T, N));
```

Of course the real benefit is that parallel for can be something else, for example it could be a Kokkos team policy, so that `Evaluate` can be called on device. Or you can do more complex things in the loop body.

singularity-eos/eos/eos_base.hpp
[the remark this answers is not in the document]
jdolence: Not sure.  I was wondering this too, but it doesn't seem to be required, at least for the compilers I've tried.  Works as is downstream on both CPU and GPU.

singularity-eos/eos/eos_base.hpp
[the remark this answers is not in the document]
jdolence: I think that lives in the lambda, yes.

Yurlungur:
I'm not going to review the code, since @jdolence and I came up with it over a whiteboard conversation. I might take a stab at adding a unit test, though, which I want before this gets merged. A few comments regarding the above discussion:

> If I'm understanding correctly, the main advantage here is that you can string together an arbitrary number of EOS lookup functions under a single mpark::visit(), right? This begs the question, is the mpark::visit() function that impactful for performance? It doesn't get optimized away? Have you profiled this change?

There are, I think, a few benefits:

1. As @dholladay00 mentioned---the visit call precludes vectorization, because there's a branch internal to the visit (unavoidably). This lifts the branch above the innermost loop. 
2. This allows the vector calls to be made on device. I agree with you @jhp-lanl that was never the intent of the vector calls. That said, it seems like a nice win. 
3. More generally, it allows a downstream C++ code to completely generalize what the vector call looks like and completely control both the loop body and the loop pattern. This allows one to fuse multiple EOS calls into a single loop body (which was the original purpose of `FillEos`), and it allows the downstream code to pick whatever loop they want. They're not bound to what's provided in `ports-of-call`, which I think is a big win because we don't want to be in the business of providing loop patterns through `ports-of-call`. In fact, I recall @clellsolomon requesting a feature like this when we discussed the vectorized code some time ago. So maybe it's of some use to EAP. (And CJ if you happen to see this, I'm curious what you think.)

> This basically allows downstream codes to write functors that can get at the underlying eos type and use that directly similar to how the vector interface accomplishes this. @jdolence has found this to have noticeable speedup. I am not clear on the main contributor to the speedup, that would be helpful to know though.

It is worth noting that @jdolence only tested for ideal gas. It's possible our other EOS models are too complex to vectorize anyway. That said, we use ideal gas sometimes and being able to vectorize for it is probably a win.

> As for the vector interface, it is still necessary and helpful as it provides concrete functions that can be called from fortran.

I don't think we can eliminate the current vector interface at this time. I thought about rolling it into this API, i.e., redefining the vector valls to just call `Evaluate` with different functors, but I think that actually ends up being *more* code than what we have. And I'm not sure how it works with the modifiers.

> As an aside, does this functionality eliminate the need for FillEOS() since you can basically inject it now?

No---because `FillEos` is intended to *know* the proper order to fill variables, i.e., it should get you what you request in the order that has the fewest internal EOS calls possible. For thermodynamically complete EOS's, like Helmholtz or Vinet, this is probably a significant savings, because the underlying potential need only be computed once.

> EDIT - Or does it make sense to provide pre-made injectables rather than defining these within the EOS or base class? For example, we could use this to eliminate DensityEnergyFromPressureTemperature() and instead define functors that could take different approaches to finding the density and energy given that there may not be a unique solution. This would also allow the downstream code to make their own decisions if they prefer.

This I am a fan of! Would love to have an injectable instead of using `PTofRE`.

> I know this isn't what you were intending here, but if you extend the existing vector interface to have a vector version of Evaluate(), you could also think about using the python interface to inject arbitrary functions as well.

This would be cool! I don't know if this kind of thing is possible with pybind. Maybe @rbberger knows?

singularity-eos/eos/eos_base.hpp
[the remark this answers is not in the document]
jhp-lanl: Ah right, that makes sense. Since we're not overloading it in the EOS, it just comes along via inheritance.

singularity-eos/eos/eos_base.hpp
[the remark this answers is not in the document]
jhp-lanl: That makes sense.

My line of thought at the time was that if you're chaining together calls you might want to pass a lambda between them.

jdolence:
> I approve, but I have some questions and a comment:
> 
> If I'm understanding correctly, the main advantage here is that you can string together an arbitrary number of EOS lookup functions under a single `mpark::visit()`, right? This begs the question, is the `mpark::visit()` function that impactful for performance? It doesn't get optimized away? Have you profiled this change?
> 
> As a tangent in defense of the current vector interface, it was never intended to be called on device. The predominant wisdom at the time it was written was that the only real way that the EOS would be called from device would be via the scalar interface and that the host could chain together whatever calls it needed. This duplicates the `visit()` calls, yes, but otherwise allows the flexible parallelism you mention in your PR description.
> 
> EDIT - I may have come off too critical. I think this is a very useful addition and could be powerful in a number of different scenarios.

Sorry -- reread my PR description and I think it may have come off more critical than I intended of the current vector interface.  There's nothing wrong with that interface, it just does a very specific thing that happens not to align with what I'm trying to do.

As @dholladay00 points out, I'm not sure if it's the cost of the `visit`, per se, or instead the fact that loops can vectorize that couldn't before.  Whether or not a loop actually vectorizes ultimately depends on the implementation of each EOS, and I'm not sure which do and which don't.  `IdealGas` clearly should, at least, and it's possible others do or could be transformed in a way that allows it.  In my tests downstream, our whole code sped up by ~10% (maybe a bit more) when I just explicitly pulled out an `IdealGas` before looping over cells and making EOS calls, for example.

Another thing this PR enables are other kinds of operations that may or may not have anything to do with EOS in the same loop.  For example, say you wanted to actually compute sound speed.  You could write a functor that calls the EOS for bulk modulus and uses that result to compute the adiabatic sound speed.  Really you can write whatever you want to execute in the functor just as you would anything else in your downstream code.

And another thing this enables is for downstream codes to use their native data structures, if they desire.  There may still be advantages to copying state out into a contiguous vector and then copying back results, but it's probably not always advantageous to do so.

It could be that many of the EOSs won't vectorize anyway, so the only effective difference from just calling the EOS functions on the variant is that you call `visit` less.  It's not clear if this matters performance-wise (and I suppose it could be different on CPU and GPU).  A key selling point of this PR is that it only requires 9 lines of new code, and (at least some) downstream codes would write things almost identically to how they would otherwise have been written, so there's not much downside.

Yurlungur (changes requested):
Needs tests. But otherwise LGTM. Pleased how this came out. :D

jhp-lanl:
Thanks for answering my questions @dholladay00 @Yurlungur and @jdolence ! I think I understand this better now. I think this is ready to be merged.

I'd like to also decrease the number of overloads we have on the vector interface in the future and this might be a tool to do so. I'll have to give it more thought though.

dholladay00:
I'd be interested to see if this improves performance in places where consecutive eos calls are present such as in get_sg_eos and in the PTE solvers.

jhp-lanl:
> Needs tests. But otherwise LGTM. Pleased how this came out. :D

Thanks for the reminder!. A really simple test should cover it and then it can be merged

Yurlungur:
> I'd be interested to see if this improves performance in places where consecutive eos calls are present such as in get_sg_eos and in the PTE solvers.

Can't be used in the PTE solvers yet. We'd need a vector solver.

jhp-lanl:
> > I'd be interested to see if this improves performance in places where consecutive eos calls are present such as in get_sg_eos and in the PTE solvers.
> 
> Can't be used in the PTE solvers yet. We'd need a vector solver.

But it can be used where we request multiple things from an EOS in quick succession (like within a PTE solver)

dholladay00:
> > I'd be interested to see if this improves performance in places where consecutive eos calls are present such as in get_sg_eos and in the PTE solvers.
> 
> Can't be used in the PTE solvers yet. We'd need a vector solver.

I am thinking specifically in material loops there are several places where multiple eos calls are made per loop iteration, the evaluate could be used in such a way as to only require 1 visit per loop iteration. This would only be helpful if visits themselves are expensive. It's possible there are situations where they may be if, for example, every visit consumes a lot of registers.

Yurlungur:
Ah I see. Yes that's true. That might buy us something.

singularity-eos/eos/eos_base.hpp
```diff
@@ -128,6 +128,13 @@ class EosBase {
   struct is_raw_pointer
       : std::is_same<std::remove_reference_t<std::remove_cv_t<T>>, R *> {};
 
+  // Generic evaluator
+  template <typename Functor_t>
+  PORTABLE_INLINE_FUNCTION void Evaluate(Functor_t &f) const {
```
Yurlungur, line 133: I changed this to accept non-const `operator()`s. This lets you create functors with side-effects in case you want to, e.g., store the value of a reduction.

I think const is less error prone, but this seems more flexible to me. Willing to walk it back if people object.

test/test_eos_unit.cpp
```diff
@@ -509,6 +509,71 @@ SCENARIO("Ideal gas entropy", "[IdealGas][Entropy]") {
...
+  void operator()(const T &eos) {
+    portableReduce(
+        "MyCheckPofRE", 0, N_,
+        PORTABLE_LAMBDA(const int i, int &nw) {
+          nw += !(isClose(P_[i],
+                          eos.PressureFromDensityInternalEnergy(rho_[i], sie_[i], nullptr),
+                          1e-15));
+        },
+        nwrong);
+  }
```
Yurlungur, line 526: I thought this was cute---something you can't do with the generic vector ops is a reduction. Here we do a reduction.

test/test_eos_unit.cpp
```diff
@@ -509,6 +509,71 @@ SCENARIO("Ideal gas entropy", "[IdealGas][Entropy]") {
...
+  // must be initialized to zero because portableReduce is a simple for loop on host
+  int nwrong = 0;
+
+ private:
+  int N_;
+  Real *P_;
+  Real *rho_;
+  Real *sie_;
+};
+SCENARIO("Ideal gas vector Evaluate call", "[IdealGas][Evaluate]") {
```
Yurlungur, line 537: Story for later, but I really want to split up `test_eos_unit.cpp` into more, smaller files. Some day soon.

doc/sphinx/src/using-eos.rst
```diff
@@ -216,6 +216,99 @@ available member function.
    eos.TemperatureFromDensityInternalEnergy(density.data(), energy.data(), temperature.data(),
                                             scratch.data(), density.size());
 
+The Evaluate Method
```
Yurlungur, line 219: Took a stab at documentation

Yurlungur (approved):
Took a stab at documentation and tests. Tests triggered on Darwin.

doc/sphinx/src/using-eos.rst
jdolence: Thanks!

singularity-eos/eos/eos_base.hpp
jdolence: Yeah I agree -- `const` is probably safer, but the ability to do reductions may come in handy and I think justifies dropping the `const` requirement.

doc/sphinx/src/using-eos.rst
jhp-lanl: Can you put a quick blurb about this up in the "The parallelism model" section? I think this would be good to advertise up there.

jhp-lanl (approved):
Thanks for the documentation and test @Yurlungur ! This looks good to me other than a minor change to the docs

test/test_eos_unit.cpp
```diff
@@ -509,6 +509,76 @@ SCENARIO("Ideal gas entropy", "[IdealGas][Entropy]") {
   }
 }
 
+// A non-standard pattern where we do a reduction
+class CheckPofRE {
+ public:
+  CheckPofRE(Real *P, Real *rho, Real *sie, int N) : P_(P), rho_(rho), sie_(sie), N_(N) {}
+#pragma nv_exec_check_disable
```
dholladay00, line 516: do we need to only include these pragmas when +cuda?

Yurlungur: Yes they only need to be there when `+cuda` but I assume they don't do anything otherwise. Is there an easy way to add them only in that case? This basically to remove the `calling device function from __host__ __device__ function` warning, because `Evaluate` is decorated as `PORTABLE_INLINE_FUNCTION`, but in this case we know we're on host. Could potentially move those pragmas to inside the base class, as well, which might be cleaner.

Yurlungur: alternatively could differentiate `EvaluateHost` and `EvaluateDevice` but that seems kinda gross IMO

doc/sphinx/src/using-eos.rst
Yurlungur: Good point @jhp-lanl ! Done.

test/test_eos_unit.cpp
dholladay00: https://github.com/lanl/singularity-eos/blob/main/singularity-eos/eos/eos_eospac.hpp#L47

Yurlungur: Done.

Yurlungur: And moved it to the decorations for `Evaluate` in the base class, rather than in the functor.

Yurlungur:
Tests now pass on darwin. As soon as tests pass on github we can merge.

singularity-eos/eos/eos_base.hpp
```diff
@@ -128,6 +128,26 @@ class EosBase {
...
+  // is __host__ __device__
+  // This is desirable, because you may want to call device-side batched
+  // evaluations (or inner loop dispatch with hierarchical parallelism)
+  // or you may want to call Evaluate from the host to, e.g., launch
+  // a reduction operation or somesuch.
+  // Note that this is a little dangerous as the compiler won't tell you
+  // if you messed up and ACTUALLY called a host-side function from
+  // device. The code will just segfault.
+#if defined(__CUDACC__)
+#pragma nv_exec_check_disable
```
dholladay00, line 142: yeah this is a bit of a hammer IMO.

dholladay00: Require that if the downstream user to place the pragmas if they are calling host functions and want warning free code. I assume if one marks the `operator()` as a portable that it won't give a warning?

You could try constexpr. 

Either way, hiding a (admittedly sometimes annoying) warning that may save a developer a lot of time does not feel like the right compromise.

Yurlungur: Agreed that the pragma solution is dangerous. The constexpr solution works nicely. Thanks for the suggestion @dholladay00 !

Issue #309, opened 2023-10-12
Move to catch2 v3?


Yurlungur:
At some point it may be worth modernizing our test infrastructure by moving from v2 of catch2 to v3. v3 is not header-only, it's more a proper static library, which should speed up compilation/linking times, as the entire catch library doesn't need to be built/linked per translation unit.

My question for you all is: what are the impediments to doing so from the perspective of our tooling?
- @mauneyc-LANL you have a comment in the singularity-eos build that v2.13.7 is the version of catch that works with clang/newer compilers. Is this a problem for updating to newer catch?
- Similarly, do we need update `Spiner` and `ports-of-call` to newer catch if we want `singularity-eos` updated?
- Finally, for spack, is it enough to change the target in the spackage? Or do we need to install catch v3 on the CI shared spack? (Question for @ktsai7 or @rbberger )

ktsai7:
@Yurlungur if you change the dependency to `catch2@3.x` in the spackage, it will get pulled in and installed during CI, so it shouldn't stop your development

The CI will be slightly slower since it has to build the `catch2@3.x` inflight until the next regular deployment cycle that pulls in the updated newer spackage after the spackaage has been merged in. If this slowed down is too much, then let me know and I can trigger a manual deployment update to pull the package in.

dholladay00:
Isn't this now completed @Yurlungur ?

Yurlungur:
No this is the one thing I didn't get to when messing with the tests.

dholladay00:
This is now done correct?

Yurlungur:
Yes.

PR #310, opened 2023-10-12
Speed up and clean up unit tests

Yurlungur:
## PR Summary

This is a long-overdue PR to improve both the performance and readability of our unit tests. I do the following:
1. Split `test_eos_unit` into multiple smaller files. In particular, the tabulated EOSs are new separated into sesame  and stellar collapse files. The lower-level tests for, e.g., root finding are in a separate file now, and the tests for vector EOS and EOS builder are in separate files. This dramatically improves readability, but it actually made compile times worse (for reasons will explain below).
2. I split the `eos_unit_tests` executable into 3 executables, which are all registered with `ctest`. I try to organize around theme (such as analytic EOS, or infrastructure). This reduces link times, as we can link 3 different executables in parallel, all of which are smaller.
3. I turn on the `CATCH_CONFIG_FAST_COMPILE` flag. I'm actually not sure what this does... remove some error handling, I think. It improves compile times by 20% or so.
4. The reason compile times got slower when I split tests into more translation units is because each object was compiling the *entire* variant, even if the test was only testing, e.g., Gruneisen. As a final performance improvement, I selectively shrink the variant in each translation unit to contain only the models actually tested. In some cases, like testing modifiers or the EOS builder, this is still the whole variant, so the code path is exercised. But for single model tests, it's usually now that model + IdealGas.

With all these changes, I am able to shrink compile times (without closures, or fortran, or sesame) to 2m30s. This is a more than 2x improvement. I also think the code is substantially cleaner now.

There's one more improvement I'd like to make, probably in a separate PR, which I would like some input on. See issue #309 .

## PR Checklist


doc/sphinx/src/contributing.rst
```diff
@@ -245,15 +244,23 @@ test in the ``CMakeLists.txt`` file,
...
+instantiated when ``singularity-eos`` is compiled.  Therefore to
+exercise all code paths, it is best to create an ``EOS`` type
+instantiated as
 
 .. code-block:: c++
 
     #include <singularity-eos/eos/eos.hpp>
-    EOS my_eos = my_new_eos(parameter1, parameter2, ...)
+    using EOS = singularity::Variant<MyNewEOS>;``.
+    EOS my_eos = MyNewEOS(parameter1, parameter2, ...)
```
jhp-lanl, line 263: Thank you for this addition! I was going to suggest you do it, but you already had!

test/CMakeLists.txt
```diff
@@ -12,35 +12,58 @@
...
+  singularity-eos::singularity-eos)
+target_link_libraries(eos_infrastructure_tests PRIVATE Catch2::Catch2
+  singularity-eos::singularity-eos)
+target_link_libraries(eos_tabulated_unit_tests PRIVATE Catch2::Catch2
+  singularity-eos::singularity-eos)
 include(Catch)
-catch_discover_tests(eos_unit_tests PROPERTIES TIMEOUT 60)
+catch_discover_tests(eos_analytic_unit_tests PROPERTIES TIMEOUT 60)
+catch_discover_tests(eos_infrastructure_tests PROPERTIES TIMEOUT 60)
+catch_discover_tests(eos_tabulated_unit_tests PROPERTIES TIMEOUT 60)
```
jhp-lanl, line 66: This is also good because it allows us to have different timeouts for the different types of tests. I don't imagine the analytic tests taking more than a minute, but tabular could in the future.

test/eos_unit_test_helpers.hpp
```diff
@@ -71,34 +71,65 @@ inline void array_compare(int num, X &&x, Y &&y, Z &&z, ZT &&ztrue, XN xname, YN
...
-  REQUIRE(isClose(test_e.BulkModulusFromDensityTemperature(1, 1),
-                  ref_e.BulkModulusFromDensityTemperature(1, 1), 1.e-15));
-  REQUIRE(isClose(test_e.GruneisenParamFromDensityTemperature(1, 1),
-                  ref_e.GruneisenParamFromDensityTemperature(1, 1), 1.e-15));
-  REQUIRE(isClose(test_e.MinimumDensity(), ref_e.MinimumDensity(), 1.e-15));
-  REQUIRE(isClose(test_e.MinimumTemperature(), ref_e.MinimumTemperature(), 1.e-15));
+  INFO("reference T: " << ref_e.TemperatureFromDensityInternalEnergy(1, 1) << " test T: "
+                       << test_e.TemperatureFromDensityInternalEnergy(1, 1));
+  CHECK(isClose(test_e.TemperatureFromDensityInternalEnergy(1, 1),
+                ref_e.TemperatureFromDensityInternalEnergy(1, 1), 1.e-15));
```
jhp-lanl, line 84: I like this improvement as well!

jhp-lanl (approved):
Great change! Thank you for doing this!

test/eos_unit_test_helpers.hpp
Yurlungur: I think that's due to you! I just copy-pasted it out of `test_eos_unit`, which had more detailed reporting. :)

jhp-lanl: Haha I forgot I wrote that. Well either way, it's a useful change!

Issue #133, opened 2022-06-06
Refactor EOS Builder


Yurlungur:
The EOS Builder is getting cumbersome as we add more modifiers. Some care should be placed in cleaning it up.

Yurlungur:
Resolved by #311

Issue #28, opened 2021-09-22
EOS Builder refactor


Yurlungur:
To enable easy construction of nested variants, I wrote the `EosBuilder`. I believe this machinery is becoming overly complex and cumbersome. It needs a refactor.

This is potentially related to #5, as if downstream codes define which variants they consider "valid," then the builder will need to be modified substantially to support this flexibility.

Yurlungur:
Resolved by #311

PR #311, opened 2023-10-13
EOSBuilder rewrite

Yurlungur:
## PR Summary

This has been a long time coming, as referenced in #28 and #133 . The `EOSBuilder` infrastructure inflexible, hard to extend, and slow. Moreover, as we move towards a model where each downstream code has its own variant (at least for the C++ codes), the hardcoded nature of `EOSBuilder` becomes an issue.

Here I refactor this machinery to allow a downstream code to be able to iteratively build up an equation of state from a base type and a nested chain of modifiers. For example:
```C++
using namespace singularity;
EOS eos = IdealGas(gm1, cv);
if (do_shift) {
  eos = EOSBuilder::Modify<ShiftedEOS>(eos, shift);
}
if (do_scale) {
  eos = EOSBuilder::Modify<ScaledEOS>(eos, scale);
}
```

This machinery is built off of some tricky templating nonsense. In particular, the `Modify` method recursively walks through all types in the variant to find the current underlying type of `eos` and modify it. This must be guarded by checks that the modifier applied to the `eos` object is *still* in the variant.

Playing these games also allows me to remove the fully EOS type list from the `test_modifiers.cpp` unit tests file and use a smaller type list, speeding up unit test compile times. 

I was also able to remove `eos_builder.cpp` entirely. Just to speed up compile times, I added an `eos.cpp` file where some modifiers applied to EOS objects are declared, providing some code for the linker across translation units. This can be removed if desired, but doing so will require modifying the build system to support an entirely header-only build mode.

These kinds of games are easier in C++17 than in older standards, so it's quite likely the variatic code I wrote can be cleaned up once we jump standards. In particular `if constexpr` would be super useful here.

## PR Checklist


test/test_eos_tabulated.cpp
```diff
@@ -261,54 +259,6 @@ SCENARIO("SpinerEOS depends on rho and sie", "[SpinerEOS],[DependsRhoSie]") {
     steelEOS.Finalize();      // cleans up device memory
   }
 }
-
-SCENARIO("EOS Builder and SpinerEOS",
```
Yurlungur, line 265: Doesn't make sense to test this anymore, as there's no special code path for the builder. Testing modifiers elsewhere does the trick.

test/test_eos_modifiers.cpp
```diff
@@ -61,34 +117,14 @@ SCENARIO("EOS Builder and Modifiers", "[EOSBuilder][Modifiers][IdealGas]") {
           REQUIRE(unmod.IsType<IdealGas>());
         }
       }
-    }
-    WHEN("We use the EOSBuilder") {
```
Yurlungur, line 65: Unified with the above test case, as EOSBuilder now no longer is a special code path.

singularity-eos/eos/eos.cpp
```diff
@@ -0,0 +1,27 @@
...
+
+using namespace singularity;
+
+#define DECLARE_MODS(x)                                                                  \
+  template class ShiftedEOS<x>;                                                          \
+  template class ScaledEOS<x>;                                                           \
+  template class UnitSystem<x>
+
+// Just explicitly declare some templated classes to speed up compile
+// times. Just pick the most commonly used ones.
```
dholladay00, line 25: What was the compile time improvement? I've heard mixed reviews and observed ambiguous results when it comes to explicit template instantiation.

singularity-eos/base/variadic_utils.hpp
```diff
@@ -156,7 +181,7 @@ constexpr auto pack_size(type_list<Ts...>) {
   return sizeof...(Ts);
 }
 
-} // namespace detail
+} // namespace types
```
dholladay00, line 184: This namespace contains more than types, what is the purpose of the namespace name change?

singularity-eos/eos/eos_builder.hpp
```diff
@@ -15,129 +15,116 @@
...
-EOS makeBilinearRamp(T &&eos, Real r0, Real a, Real b, Real c) {
-  return BilinearRampEOS<T>(std::forward<T>(eos), r0, a, b, c);
+// The ModifyDispatcher function specifies which tag in tag dispatch
+// to use. It's main purpose is to check whether or not U and Mod<U>
+// are in the type list for the variant. If yes, it dispatches to the
+// ModifyHelper<Mod>(yes,...) method, otherwise, the
+// ModifyHelper<Mod>(no,...)  method.
+template <template <class...> typename Mod, typename U, typename... Rest, typename... Ts,
+          typename... Args>
+Variant<Ts...> ModifyDispatcher(const Variant<Ts...> &var,
```
dholladay00, line 62: should the modifydispatcher functions be marked constexpr?  Could these be implemented with an empty struct?

dholladay00: seems like maybe empty structs with partial specialization and std conditional in a type alias may do the trick?

singularity-eos/eos/eos.cpp
Yurlungur: To be honest, I didn't see a difference. This was mostly me being superstitious. I also couldn't entirely eliminate a source file the way the build system is now. It did not like building a library with no translation units.

singularity-eos/base/variadic_utils.hpp
Yurlungur: I wanted to expose this machinery to downstream codes so they could use the automated stuff to build up a variant. Not married to the naming, and I'm willing to change it back if you prefer.

singularity-eos/eos/eos_builder.hpp
Yurlungur: It can almost certainly be marked conatexpr. Not sure I understand how to implement this as an empty struct though. Is there a benefit of doing it that way?

singularity-eos/eos/eos.cpp
mauneyc-LANL: For the size of the template parameter lists I've seen, `singularity-eos` is quite tame compared to other codes, so compile-time speedups locally probably are not noticeable. Maybe there's some benefit downstream for heavy TMP codes depending on how `singularity-eos` is used.

Yurlungur: @mauneyc-LANL would you recommend keeping it around then, for that reason?

singularity-eos/eos/eos_builder.hpp
mauneyc-LANL: I think it's fine to use either; myself I prefer using imperative-style (i.e. functions) to do this kind of type manipulation, though you should add `constexpr` to make clear to the compiler that this is resolved at compile-time (I wouldn't object to `auto` here either).

Using structs, you're being explicit that the compiler should sort things out. If our consensus is we trust compilers to have compliant `constexpr`, then either way is fine.

singularity-eos/base/variadic_utils.hpp
```diff
@@ -18,7 +18,32 @@
...
+template <bool... Bs>
+using conjunction = std::is_same<bool_pack<true, Bs...>, bool_pack<Bs..., true>>;
+template <bool... Bs>
+struct disjunction : bool_constant<!conjunction<!Bs...>::value> {};
+
+// Checks if T is contained in the pack Ts
+template <typename T, typename... Ts>
+constexpr bool contains_v() {
+  return disjunction<std::is_same<T, Ts>::value...>::value;
+}
```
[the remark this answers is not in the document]
mauneyc-LANL: Let's develop separately, then see how they compare.

singularity-eos/eos/eos_builder.hpp
Yurlungur: ok---it's easiest to just mark it `constexpr` , and I find a function style easier to interpret/reason about, so I'll do that. Will try adding `auto` too. I think experimented with that, and the compiler couldn't deduce types properly, but I may be mistaken and that was caused by some other issue.

singularity-eos/eos/eos.cpp
mauneyc-LANL: hmmm..I wouldn't recommend *not* keeping them around (?). I don't *know* there's any benefit, but there's minimal downside. I guess I don't object to them being there, but I can't affirm it's necessary or useful.

Having said that, I've gotten lost for days in macros/template compiler errors, which can be really nasty just to identify before you even get to resolving it. Granted those codes had **a lot** more of this kind of pattern, which I don't think would be an issue with this snippet.

If pressed, I would say unless we can show it helps then probably best to avoid any pitfalls.

singularity-eos/eos/eos_builder.hpp
Yurlungur: Done.

singularity-eos/eos/eos_builder.hpp
```diff
@@ -15,129 +15,116 @@
...
+// object. If both (a) and (b) are true, it returns a new variant
+// containing Mod<u>. Otherwise, it moves to the next type.
+template <template <class...> typename Mod, typename U, typename... Rest, typename... Ts,
+          typename... Args>
+Variant<Ts...> ModifyHelper(std::true_type, const Variant<Ts...> &var,
+                            const type_list<U, Rest...> &tl, Args &&...args) {
+  bool contains_u = var.template IsType<U>();
+  if (contains_u) {
+    return Variant<Ts...>(Mod<U>(var.template get<U>(), std::forward<Args>(args)...));
+  } else {
```
mauneyc-LANL, line 82: This is related to above, but unfortunately without `if constexpr` here the function style breaks here. Probably compilers will optimize it away and get something like compile-time behavior, but as C++ code this is explicitly a run-time function.

I need to look a bit deeper. If it's intended as a run-time function, then it can't (or at least _shouldn't_) be used in `constexpr` calls. However it's not difficult to emulated `if constexpr` (you already do).

mauneyc-LANL (changes requested):
I'd like to get a bit deeper into this. We may want to go about this differently; in particular, stacking type-lists and `variant` may be adding complexity we can avoid, and with what you have it may be just a little extra is enough to dump `variant` entirely.

singularity-eos/eos/eos.cpp
dholladay00: I would say to get rid of it unless it gives noticeable speedups in compile times.

singularity-eos/eos/eos_builder.hpp
Yurlungur: @dholladay00 I can't easily get the empty struct version to compile, so I'm inclined to leave the `constexpr` function version for now until compile times with the builder become a noticeable issue. FWIW the compile time on my desktop for the unit tests excluding the fortran and closure code is now ~30s, reduced from 7m!

singularity-eos/base/variadic_utils.hpp
Yurlungur: Renamed to `variadic_utils`

singularity-eos/eos/eos.cpp
Yurlungur: Removed `eos.cpp`. Library now builds as header-only in the absence of the fortran bindings. Turns out the only thing I needed to add this:
```C++
  set_target_properties(singularity-eos PROPERTIES LINKER_LANGUAGE CXX)
```
which I guarded in a check for fortran bindings (since the proper linker is obviously not exclusively C++ in that case).

singularity-eos/eos/eos_builder.hpp
Yurlungur: @dholladay00 I added some additional functionality to help build up some struct machinery (although I still can't get the struct version to work). Primarily, I added an empty `IsModifiable` struct which should help.

singularity-eos/eos/eos.cpp
Yurlungur: Oops I lied. That fixes local tests, but codes downstream that rely on `singularity-eos` being installed still look for `a libsingularity.a` file and they die at the linker phase. I put `eos.cpp` back in, but it's just an empty file with a long comment explaining what I did.

Yurlungur: Comment transcribed below:
```
// JMM: This is a dummy file to provide a source file for cmake to
// compile when built without fortran.
//
// The main reason to do this is that when we build without fortran
// bindings, the library is completely header only. However WITH
// fortran bindings, it is not.
//
// cmake supports a build mode for header only libraries by marking
// those libraries INTERFACE. Without this flag, a few things will go
// wrong:
// 1. CMake will be unable to infer a proper linker
//    so one must be specified by hand.
// 2. no dynamic library file (e.g., "*.a" will be generated,
//    but dependencies that link against it, such as another library,
//    or in our case, tests, will look for it. This will lead to
//    a failure of the downstream library at link time.
// 
// On the other hand, a library with implementation files
// CANNOT be marked INTERFACE, as otherwise source files will
// not be compiled.
//
// Unfortunately, switching the INTERFACE tag on and off in cmake
// is very cumbersome, as if a library is marked INTERFACE
// all its dependencies must ALSO be slurped in with an INTERFACE
// flag. This introduces significant branching in the cmake code
// and a lot of builer plate. For now, then, I simply include this
// empty source file.
//
// In the future, we could include centralized code here that we
// always want compiled, such as template instantiations or
// convenience functions.
```

singularity-eos/eos/singularity_eos.cpp
```diff
@@ -20,6 +20,56 @@
 
 using namespace singularity;
 
+// TODO: Replace these with the new Modify method in EOSBuilder.
+// NOTE: The new EOSBuilder machinery will likely be slower than these.
+template <typename T>
+EOS applyShiftAndScale(T &&eos, bool scaled, bool shifted, Real scale, Real shift) {
```
Yurlungur, line 26: @dholladay00 if we end up keeping the rewrite, we should convert `singularity_eos.cpp` to use the new builder code at some point. But for now, I left the older `applyXY*` stuff, as it works without intervention. It may also be faster to compile on the full variant.

test/test_variadic_utils.cpp
Yurlungur, line 1: New tests for some of the variatic functionality

singularity-eos/eos/eos.cpp
mauneyc-LANL: Would the simpler solution here just be to have a separate target for the Fortran library?

Yurlungur: That sounds reasonable to me. But yeah, let's leave it for a later PR.

mauneyc-LANL (approved):

singularity-eos/eos/eos_builder.hpp
```diff
@@ -15,129 +15,130 @@
...
-                         time_unit, mass_unit, length_unit, temp_unit);
-  } else {
-    return UnitSystem<T>(std::move(eos), eos_units_init::thermal_units_init_tag, rho_unit,
-                         sie_unit, temp_unit);
-  }
+template <template <class...> typename Mod, typename U, typename... Ts>
+using IsModifiable =
+    conjunction<contains<U, Ts...>::value, contains<Mod<U>, Ts...>::value>;
+template <template <class...> typename Mod, typename U, typename... Ts>
+constexpr bool is_modifiable(const Variant<Ts...> &var) {
```
dholladay00, line 36: Could this function arg be replaced with a type list rather than a `Variant`?

Yurlungur: Sure---though I would suggest adding an overload with a type list version rather than getting rid of the variant version... because I was thinking maybe the variant version might be useful in userspace?

test/compare_on_data.cpp
```diff
@@ -212,16 +212,11 @@ void load_eos_spiner(const std::string &filename, const std::vector<int> &matids
...
-  EOSBuilder::params_t params;
-  params["filename"] = filename;
-  params["reproducibility_mode"] = false;
 
   for (int matid : matids) {
     if (matid > 0) { // non-analytic
-      params["matid"] = matid;
-      eos_h.push_back(EOSBuilder::buildEOS(type, params));
-      eos_d.push_back(eos_h.back().GetOnDevice());
+      eos_h.push_back(SpinerEOSDependsRhoT(filename, matid))
```
dholladay00, line 218: Missing `;`?

Yurlungur: Good catch. fixed.

singularity-eos/eos/eos_builder.hpp
Yurlungur: Done. Added an overload that takes a type list. I recommend using that one for internals, but I think the variant version is useful for userspace.

test/compare_on_data.cpp
dholladay00: is this test not run in our standard CI testing? 

Has this been run on gitlab CI?

Yurlungur: It's not run our built (or even tied into the cmake). It's one of my old testing scripts from the early days of singularity. It's designed to read in something like an xrage dump and compare eospac vs spiner on realistic datasets. I don't want to delete it, as I think it will come in handy again at some point, but it's not something we need to worry too much about maintaining.

dholladay00: Maybe we should have something like a deprecated/one-off/old directory to move some of this legacy stuff so as not to be confused with files in the test dir that are connected to cmake and commonly built/tested?

dholladay00: Maybe that movement could/should be a separate PR, if so, an issue should be made.

Yurlungur:
I finally had a chance to return to this.

@dholladay00 thanks for the suggestion of using CRTP to clean this up/make it more performant. The result of this is the following:
1. I am still using the `contains` empty struct, to check whether or not a type is in a type list.
2. However, all explicit tail recursion is gone. Since the base class knows the underlying type through CRTP, we can jump right to that spot in the variant and modify that object directly.

The public facing API is identical, plus an object-oriented API is available:
```C++
eos = eos.Modify<Modifier>(args);
```
now works, and I also provide
```C++
bool is_modifiable = eos.ModifiedInVariant<Mod>();
```
which checks that the *current* type can be modified and remain in the variant. In the variant, this code looks basically identical to all the other variant functions---just calls to visit.

Note that grossness couldn't be *entirely* removed. The problem is that `mpark::visit` requires that the visiting function always have the same return type. So in `Modify`, I need a visitor that always returns a variant with the same types. To make this happen, I needed some tag dispatch in the base class:
```C++
  // Modifies an EOS object by pulling out underlying type and modifying it
  // This one returns Mod<CRT>
  template <template <class> typename Mod, typename... Args>
  constexpr Mod<CRTP> Modify(Args &&... args) const {
    CRTP unmodified = *(static_cast<CRTP const *>(this));
    return Mod<CRTP>(std::move(unmodified), std::forward<Args>(args)...);
  }
  // These are overloads needed for the variant, as std::visit must be
  // able to return a variant of the same type every time. This lets
  // us do so, even though sometimes we don't modify the object.
  template <template <class> typename Mod, typename... Args>
  constexpr Mod<CRTP> ConditionallyModify(std::true_type, Args &&... args) const {
    return Modify<Mod>(std::forward<Args>(args)...);
  }
  template <template <class> typename Mod, typename... Args>
  constexpr CRTP ConditionallyModify(std::false_type, Args &&... args) const {
    CRTP unmodified = *(static_cast<CRTP const *>(this));
    return unmodified;
  }
  // Empty type list struct just used to capture "Ts" and "Args" simultaneously.
  template <template <class> typename Mod, typename... Ts, typename... Args>
  constexpr auto ConditionallyModify(const variadic_utils::type_list<Ts...> &tl,
                                     Args &&... args) const {
    constexpr bool do_mod = variadic_utils::contains_v<Mod<CRTP>, Ts...>();
    return ConditionallyModify<Mod>(variadic_utils::bool_constant<do_mod>(),
                                    std::forward<Args>(args)...);
  }
```
then the variant checks at runtime that you never hit `ConditionallyModify(std::false_type,...)`:
```C++
  // EOS modifier object-oriented API
  template<template<class> typename Mod>
  constexpr bool ModifiedInVariant() const {
    return mpark::visit(
        [](const auto &eos) { return eos.template ModifiedInList<Mod, EOSs...>(); },
        eos_);
  }
  template<template<class> typename Mod, typename... Args>
  constexpr auto Modify(Args &&...args) const {
    PORTABLE_ALWAYS_REQUIRE(ModifiedInVariant<Mod>(), "Modifier must be in variant");
    return mpark::visit(
        [&](const auto &eos) {
          auto modified = eos.template ConditionallyModify<Mod>(
              variadic_utils::type_list<EOSs...>(), std::forward<Args>(args)...);
          return eos_variant<EOSs...>(modified);
        },
        eos_);
  }
```

Anyway I think this is in pretty good shape now.

singularity-eos/eos/eos.cpp
```diff
@@ -0,0 +1,45 @@
...
+// Nuclear Security Administration. All rights in the program are
+// reserved by Triad National Security, LLC, and the U.S. Department of
+// Energy/National Nuclear Security Administration. The Government is
+// granted for itself and others acting on its behalf a nonexclusive,
+// paid-up, irrevocable worldwide license in this material to reproduce,
+// prepare derivative works, distribute copies to the public, perform
+// publicly and display publicly, and to permit others to do so.
+//------------------------------------------------------------------------------
+
+// JMM: This is a dummy file to provide a source file for cmake to
```
dholladay00, line 15: @Yurlungur could you create an issue on how to best address this issue and tag @rbberger in it?

Yurlungur: Done #323

singularity-eos/eos/eos_base.hpp
```diff
@@ -137,6 +138,39 @@ class EosBase {
...
+  // These are overloads needed for the variant, as std::visit must be
+  // able to return a variant of the same type every time. This lets
+  // us do so, even though sometimes we don't modify the object.
+  template <template <class> typename Mod, typename... Args>
+  constexpr Mod<CRTP> ConditionallyModify(std::true_type, Args &&...args) const {
+    return Modify<Mod>(std::forward<Args>(args)...);
+  }
+  template <template <class> typename Mod, typename... Args>
+  constexpr CRTP ConditionallyModify(std::false_type, Args &&...args) const {
+    CRTP unmodified = *(static_cast<CRTP const *>(this));
```
dholladay00, line 163: should we emit a warning that the modification was unsuccessful?

singularity-eos/eos/eos_variant.hpp
```diff
@@ -980,12 +1001,18 @@ class Variant {
 
   inline constexpr Variant UnmodifyOnce() {
     return mpark::visit(
-        [](auto &eos) { return eos_variant<EOSs...>(eos.UnmodifyOnce()); }, eos_);
+        [](auto &eos) -> eos_variant<EOSs...> {
```
dholladay00, line 1004: were there issues with implicit conversions?

Yurlungur: Uh I don't remember why I had to make this change---I think there was an error about visitor always returning the same type.

singularity-eos/eos/eos_base.hpp
Yurlungur: We emit an error when it's called from the variant. We can emit a warning here too if you prefer.

dholladay00: That's sufficient.

singularity-eos/eos/eos_builder.hpp
```diff
@@ -15,129 +15,57 @@
...
-  if (shifted) {
-    ShiftedEOS<T> sh_eos(std::forward<T>(eos), shift);
-    return W<ShiftedEOS<T>>(std::move(sh_eos), args...);
-  }
-  if (scaled) {
-    ScaledEOS<T> sc_eos(std::forward<T>(eos), scale);
-    return W<ScaledEOS<T>>(std::move(sc_eos), args...);
-  }
-  return W<T>(std::forward<T>(eos), args...);
+// Modifies the eos contained in the var object with the modifier Mod,
```
dholladay00, line 48: what is going on with this commented code? The git diff makes it a bit confusing for me to figure out.

Yurlungur: That's example intended usage. This is documented in the docs now. I'll just remove the comment.

Yurlungur:
@dholladay00 says he's happy and tests pass. I'm pulling the trigger.

PR #322, opened 2023-11-27
Add error check after 'eosCreateTables()' in the 'eosSafeLoad()' function

jhp-lanl:
## PR Summary

Fixes a bug when the comment table wasn't available. The `eosCreateTables()` function would return an error code, but the code wouldn't be handled. This change adds an early return to skip the `eosLoadTables()` call if `eosCreateTables()` fails.

EDIT - An additional check was needed for when the comment table is created but when the error code returned is okay. The comments table could still be invalid and so the length needs to be checked before trying to extract comments.

I _think_ this should help address a bug found by @annapiegra where a comments table length would be negative when a comments table didn't exist, and would cause issues for the comments vector later on when the `resize()` member is called.

## PR Checklist

### DIFF
--- a/eospac-wrapper/eospac_wrapper.cpp
+++ b/eospac-wrapper/eospac_wrapper.cpp
@@ -77,24 +77,26 @@ void eosGetMetadata(int matid, SesameMetadata &metadata, Verbosity eospacWarn) {
   EOS_INTEGER errorCode =
       eosSafeLoad(1, matid, commentsType, commentsHandle, {"EOS_Comments"}, eospacWarn);
   EOS_INTEGER eospacComments = commentsHandle[0];
+  bool commentTableCreated = (errorCode == EOS_OK);
 
-  if (errorCode == EOS_OK) {
-    std::vector<EOS_CHAR> comments;
-    EOS_REAL commentLen;
+  EOS_REAL commentLen = -1;
+  if (commentTableCreated) {
     EOS_INTEGER commentItem = EOS_Cmnt_Len;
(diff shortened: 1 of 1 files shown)


annapiegra:
This doesn't solve the issue since commentLen is returned  from eosSafeTableInfo. Initializing commentLen to 0 would be a possible solution (in the current implementation if the info returns an error commentLen keeps a garbage value)

jhp-lanl:
> This doesn't solve the issue since commentLen is returned from eosSafeTableInfo. Initializing commentLen to 0 would be a possible solution (in the current implementation if the info returns an error commentLen keeps a garbage value)

@annapiegra let me know if this fixes it

eospac-wrapper/eospac_wrapper.cpp
```diff
@@ -104,6 +106,12 @@ void eosGetMetadata(int matid, SesameMetadata &metadata, Verbosity eospacWarn) {
     metadata.name = "No name for matid " + matid_str;
     metadata.comments = "Comment unavailable for matid " + matid_str;
   }
+
+  // Note: table could be created even if the commentLen is nonsense, so we
+  // need to separate this block from the above logic
+  if (commentTableCreated) {
+    eosSafeDestroy(1, commentsHandle, eospacWarn);
+  }
```
Yurlungur, line 114: :+1: good catch

Yurlungur (approved):
Thanks for fixing this @jhp-lanl this one is an interesting edge case.

jhp-lanl:
These were minor changes, but it still feels good to have the format check pass without having to run the format tool 😁

annapiegra (approved):
Works now. Thanks!

PR #330, opened 2023-12-08
Piecewise grids for spiner EOS

Yurlungur:
## PR Checklist


Yurlungur:
@jhp-lanl @dholladay00 @chadmeyer this is ready for review. I think there's more one can do in this space, but IMO this is a nice start. In particular, energy and pressure near normal density are significantly better resolved with this method. 

In anticipation of enabling spiner tables to load data from another eos, I move the `Bounds` object out of `sesame2spiner` into `singularity-eos/base`, and add tests to it.

I also found and fixed the issue @pdmullen found in riot regarding extrapolation. It was a bizarre interaction between fast logs and eospac in `sesame2spiner`. Now fixed.

Yurlungur:
Tests triggered on re-git.

Yurlungur:
The core piece of this machinery lives in spiner. So the changes in singularity are just to turn that machinery on and set up reasonable grids in `sesame2spiner`.

chadmeyer (commented):
Overall, this is a good improvement.  Most of my comments are re: documentation.  

General observations:
* Is this backwards compatible with previously generated .sp5 files or not?  I assume a newly generated .sp5 file with piecewise grids will not be compatible with older versions, but I'm not sure whether that will be true the other direction because of potential metadata changes, etc.  If there is any incompatibility like that, there will need to be some kind of error telling you what happened.
* 6 regions (2T/e x 3 rho) is probably a good balance between performance and where to put resolution to cover our needs, but is it sufficiently flexible?

doc/sphinx/src/models.rst
```diff
@@ -1491,16 +1490,82 @@ key-value pairs. For exampe the following input deck is for air:
...
+piecewise grids are active, the density grid gets split into three
+pieces, a piece of diameter ``rhoFineDiameterDecades`` (in log space)
+around the material's normal density and a piece above and below. The
+``numrho/decade`` parameter sets the number of points per decade in
+this central refined region. The regions at lower and higher density
+have ``rhoCoarseFactorLo`` and ``rhoCoarseFactorHi`` fewer points per
+decade respectively compared to the finer region.
+
+The temperature grid has two regions, a more finely spaced region at
+low temperatures and a more finely spaced region at high
```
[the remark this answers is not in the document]
Yurlungur: :+1:

doc/sphinx/src/models.rst
```diff
@@ -1491,16 +1490,82 @@ key-value pairs. For exampe the following input deck is for air:
...
+the temperature grid, with the energy split point corresponding to the
+temperature split point. The coarser high-temperature temperature and
+energy grids are coarsened by a factor of ``TCoarseFactor`` and
+``sieCoarseFactor`` respectively.
+
+Thus the input block for piecewise grid might look like this:
+
+.. code-block::
+
+  # defaults are true
```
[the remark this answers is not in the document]
Yurlungur: only the booleans. Comment now written to disambiguate.

doc/sphinx/src/models.rst
```diff
@@ -1491,16 +1490,82 @@ key-value pairs. For exampe the following input deck is for air:
...
+  piecewiseSie = true
+
+These options may be true or false. The default is true. When
+piecewise grids are active, the density grid gets split into three
+pieces, a piece of diameter ``rhoFineDiameterDecades`` (in log space)
+around the material's normal density and a piece above and below. The
+``numrho/decade`` parameter sets the number of points per decade in
+this central refined region. The regions at lower and higher density
+have ``rhoCoarseFactorLo`` and ``rhoCoarseFactorHi`` fewer points per
+decade respectively compared to the finer region.
```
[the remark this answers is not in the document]
Yurlungur: > This whole description would benefit from an image or a diagram.

That's fair. Added.

> Also, is this the best way of inputting this? Why not something like a rhominfine->rhomaxfine? I know that the same effect can be specified by adjusting quantities specified below, and maybe this is the best way for it to be internally represented, but it might be reasonable to permit a different kind of input. Just a thought there.

I agree. I chose the model I chose here because it seemed reasonable to me in the moment. Obviously `rhominfine/rhomaxfine` is a bit more flexible. I added the ability to set `rhoMinFine` and `rhoMaxFine`.

doc/sphinx/src/models.rst
```diff
@@ -1491,16 +1490,82 @@ key-value pairs. For exampe the following input deck is for air:
...
+  # energy has the split point sie(rhonormal, TSplitPoint)
+  # but we may still specify the resolution
+  numSie/decade = 100
+  sieCoarseFactor = 1.5
+
+.. note::
+
+  For all grid types, the only required value in an input file is the
+  matid. All other values will be inferred from the original sesame
+  database if possible and if no value in the input file is provided.
```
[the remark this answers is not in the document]
Yurlungur: Table bounds and normal density are inferred. Other options have sensible-ish defaults, which are now listed.

sesame2spiner/examples/aluminum.dat
```diff
@@ -0,0 +1,21 @@
...
+#  granted for itself and others acting on its behalf a nonexclusive,
+#  paid-up, irrevocable worldwide license in this material to reproduce,
+#  prepare derivative works, distribute copies to the public, perform
+#  publicly and display publicly, and to permit others to do so.
+# ======================================================================
+
+matid=3720
+name=aluminum
+rhomin = 1e-5
+Tmin = 1
```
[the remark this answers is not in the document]
Yurlungur: Mainly it's here because I tested aluminum. Also I found that the minimum density/temperatures provided by the database are frequently unnecessarily low.

sesame2spiner/generate_files.cpp
```diff
@@ -273,34 +274,88 @@ void getMatBounds(int i, int matid, const SesameMetadata &metadata, const Params
...
+
+  Real ppd_factor_rho_lo = params.Get("rhoCoarseFactorLo", COARSE_FACTOR_DEFAULT_RHO_LO);
+  Real ppd_factor_rho_hi = params.Get("rhoCoarseFactorHi", COARSE_FACTOR_DEFAULT_RHO_HI);
+  Real ppd_factor_T = params.Get("TCoarseFactor", COARSE_FACTOR_DEFAULT_T);
+  Real ppd_factor_sie = params.Get("sieCoarseFactor", COARSE_FACTOR_DEFAULT_T);
+  Real rho_fine_diameter =
+      params.Get("rhoFineDiameterDecades", RHO_FINE_DIAMETER_DEFAULT);
+  Real TSplitPoint = params.Get("TSplitPoint", T_SPLIT_POINT_DEFAULT);
+
+  Real rho_fine_center = rhoAnchor;
```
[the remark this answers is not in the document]
Yurlungur: Done.

Yurlungur:
@chadmeyer I believe I have addressed all your inline comments. A few comments regarding your observations:

>     * Is this backwards compatible with previously generated .sp5 files or not?  I assume a newly generated .sp5 file with piecewise grids will not be compatible with older versions, but I'm not sure whether that will be true the other direction because of potential metadata changes, etc.  If there is any incompatibility like that, there will need to be some kind of error telling you what happened.

It's not backwards compatible, but this is a good point. I think I can make it backwards compatible, but this is a change in *spiner* not singularity-eos. I'll submit an MR there. I'd like this to go independently of that, however.

>     * 6 regions (2T/e x 3 rho) is probably a good balance between performance and where to put resolution to cover our needs, but is it sufficiently flexible?

No idea. The code infrastructure supports more regions under the hood---it's fully general---but I had to pick something to get started and adding a fully general *input* infrastructure feels like too much for a first pass.

Yurlungur:
> > * Is this backwards compatible with previously generated .sp5 files or not?  I assume a newly generated .sp5 file with piecewise grids will not be compatible with older versions, but I'm not sure whether that will be true the other direction because of potential metadata changes, etc.  If there is any incompatibility like that, there will need to be some kind of error telling you what happened.
> > ```
> 
> It's not backwards compatible, but this is a good point. I think I can make it backwards compatible, but this is a change in _spiner_ not singularity-eos. I'll submit an MR there. I'd like this to go independently of that, however.

@chadmeyer see this MR on spiner https://github.com/lanl/spiner/pull/92

doc/sphinx/src/models.rst
```diff
@@ -1491,16 +1492,114 @@ key-value pairs. For exampe the following input deck is for air:
...
+  # it's usually wise to to not let
+  # temperature get too small in log space if you do this
+  Tmin = 1
+  # The coarser region (above the split point) is 50 percent less refined
+  TCoarseFactor = 1.5
+
+  # energy has the split point sie(rhonormal, TSplitPoint)
+  # but we may still specify the resolution
+  numSie/decade = 100
+  sieCoarseFactor = 1.5
```
jhp-lanl, line 1583: So I take it that if for example we want to increase to say 10 grid regions in some dimension, we'll need to rework how this input syntax works, yes?

doc/sphinx/src/models.rst
```diff
@@ -1491,16 +1492,114 @@ key-value pairs. For exampe the following input deck is for air:
...
+
+.. note::
+
+  For all grid types, the only required value in an input file is the
+  matid. Table bounds and normal density will be inferred from teh
```
jhp-lanl, line 1588:
```suggestion
  matid. Table bounds and normal density will be inferred from the
```
typo

sesame2spiner/generate_files.cpp
```diff
@@ -273,34 +274,110 @@ void getMatBounds(int i, int matid, const SesameMetadata &metadata, const Params
...
+  int ppdSie = params.Get("numSie/decade", PPD_DEFAULT_T);
+  int numSieDefault = Bounds::getNumPointsFromPPD(sieMin, sieMax, ppdSie);
 
   int numRho = params.Get("numrho", numRhoDefault);
   int numT = params.Get("numT", numTDefault);
   int numSie = params.Get("numsie", numSieDefault);
 
-  Real rhoAnchor = metadata.normalDensity;
-  Real TAnchor = 298.15;
+  const Real TAnchor = 298.15;
```
jhp-lanl, line 290:
```suggestion
  constexpr Real TAnchor = 298.15;
```
Any reason this can't be `constexpr`? Good addition with the `const` though!

sesame2spiner/generate_files.cpp
```diff
@@ -273,34 +274,110 @@ void getMatBounds(int i, int matid, const SesameMetadata &metadata, const Params
...
+  Real rhoAnchor = params.Get("rho_fine_center", metadata.normalDensity);
+  if (std::isnan(rhoAnchor) || rhoAnchor <= 0 || rhoAnchor > 1e8) {
```
jhp-lanl, line 292: Do we encounter `NaN`s in the `normalDensity` parameter?

sesame2spiner/generate_files.cpp
```diff
@@ -273,34 +274,110 @@ void getMatBounds(int i, int matid, const SesameMetadata &metadata, const Params
...
+    std::cerr << "WARNING [" << matid << "] "
+              << "normal density ill defined. Setting it to a sensible default."
+              << std::endl;
+    rhoAnchor = 1;
+  }
+
+  // Piecewise grids stuff
+  bool piecewiseRho = params.Get("piecewiseRho", true);
+  bool piecewiseT = params.Get("piecewiseT", true);
+  bool piecewiseSie = params.Get("piecewiseSie", true);
```
jhp-lanl, line 302: Can these be `const`?

sesame2spiner/generate_files.cpp
```diff
@@ -273,34 +274,110 @@ void getMatBounds(int i, int matid, const SesameMetadata &metadata, const Params
...
+
+  Real ppd_factor_rho_lo = params.Get("rhoCoarseFactorLo", COARSE_FACTOR_DEFAULT_RHO_LO);
+  Real ppd_factor_rho_hi = params.Get("rhoCoarseFactorHi", COARSE_FACTOR_DEFAULT_RHO_HI);
+  Real ppd_factor_T = params.Get("TCoarseFactor", COARSE_FACTOR_DEFAULT_T);
+  Real ppd_factor_sie = params.Get("sieCoarseFactor", COARSE_FACTOR_DEFAULT_T);
+  Real rho_fine_diameter =
+      params.Get("rhoFineDiameterDecades", RHO_FINE_DIAMETER_DEFAULT);
+  Real TSplitPoint = params.Get("TSplitPoint", T_SPLIT_POINT_DEFAULT);
+
+  Real rho_fine_center = rhoAnchor;
```
jhp-lanl, line 312: I might be missing changes below, but it looks like all of these could be `const`.

There are other reasons to advocate for lots of `const`, but I think from a readability perspective it's helpful to understand which variables may change and which are never going to change.

sesame2spiner/generate_files.cpp
```diff
@@ -273,34 +274,110 @@ void getMatBounds(int i, int matid, const SesameMetadata &metadata, const Params
...
+    {
+      eosSafeLoad(NT, matid, tableType, tableHandle, {"EOS_Ut_DT"}, Verbosity::Quiet);
+      EOS_INTEGER eospacEofRT = tableHandle[0];
+      rho[0] = rho[1] = densityToSesame(rhoAnchor);
+      T[0] = temperatureToSesame(TAnchor);
+      T[1] = temperatureToSesame(TSplitPoint);
+      eosSafeInterpolate(&eospacEofRT, nXYPairs, rho, T, sie, dx, dy, "EofRT",
+                         Verbosity::Quiet);
+      eosSafeDestroy(NT, tableHandle, Verbosity::Quiet);
+    }
```
jhp-lanl, line 368: Does it make sense to utilize the `EOSPAC` class here directly for this lookup?

sesame2spiner/io_eospac.hpp
```diff
@@ -31,83 +27,16 @@
...
-       << "[N,dx] = [" << b.grid.nPoints() << ", " << b.grid.dx() << "]"
-       << "\n";
-    return os;
-  }
-
- public:
-  RegularGrid1D grid;
-  Real offset;
-};
+constexpr int NGRIDS = 3;
```
jhp-lanl, line 36: I assume it's appropriate to set this separate from the restrictions on the `spiner` end, right? I.e. if `spiner` supports 10 grids, we'd need significant code modification to support that, right?

singularity-eos/base/table_bounds.hpp
```diff
@@ -0,0 +1,281 @@
...
+#include <iostream>
+#include <limits>
+#include <string>
+#include <type_traits>
+#include <vector>
+
+#include <ports-of-call/portability.hpp>
+#include <ports-of-call/portable_errors.hpp>
+#include <singularity-eos/base/fast-math/logs.hpp>
+#include <spiner/databox.hpp>
```
jhp-lanl, line 29: Do you actually use `DataBox` here?

singularity-eos/base/table_bounds.hpp
```diff
@@ -0,0 +1,281 @@
...
+#include <spiner/interpolation.hpp>
+
+namespace singularity {
```
jhp-lanl, line 32: Would it make sense for this to be in its own namespace? My thinking is along a few lines:
* We do this in other utility functions such as the root finder.
* `Bounds` is a generic enough name that I'm hesitant for it to live naked in the `singularity` namespace
* These utilities only apply to tabular EOS, and specifically to those tabular EOS based on `spiner`, right?

jhp-lanl: You could also rename the file `spiner_table_utils` and create a `spiner_utils` namespace.

singularity-eos/base/table_bounds.hpp
```diff
@@ -0,0 +1,281 @@
...
+  Bounds() : offset(0), piecewise(false), linmin_(0), linmax_(0) {}
+
+  Bounds(Real min, Real max, int N, Real offset)
+      : grid(Grid_t(std::vector<RegularGrid1D>{RegularGrid1D(min, max, N)})),
+        offset(offset), piecewise(false), linmin_(min), linmax_(max) {}
+  Bounds(OneGrid, Real min, Real max, int N, Real offset) : Bounds(min, max, N, offset) {}
+
+  Bounds(Real min, Real max, int N, bool convertToLog = false, Real shrinkRange = 0,
+         Real anchor_point = std::numeric_limits<Real>::signaling_NaN())
+      : offset(0), piecewise(true), linmin_(min), linmax_(max) {
```
jhp-lanl, line 55: It looks like several of these arguments can be `const`.

Rather than commenting on each constructor, can you make the other constructors have `const` arguments where appropriate?

test/test_bounds.cpp
```diff
@@ -0,0 +1,130 @@
...
+constexpr Real T_min = 1;
+constexpr Real T_max = 1e9;
+constexpr Real T_split = 1e4;
+constexpr int N_per_decade_fine = 200;
+constexpr Real N_factor = 5;
+
+SCENARIO("Bounds can compute number of points from points per decade", "[Bounds]") {
+  WHEN("We compute the number of points from points per decade") {
+    int np = Bounds::getNumPointsFromPPD(T_min, T_max, N_per_decade_fine);
+    THEN("We get the right number") { REQUIRE(np == 1800); }
```
jhp-lanl, line 44: Where does `1800` come from? Would it make sense to calculate it instead?

test/test_bounds.cpp
```diff
@@ -0,0 +1,130 @@
...
+  }
+}
+
+SCENARIO("Logarithmic, single-grid bounds in the bounds object", "[Bounds]") {
+  WHEN("We compute logarithmic single-grid bounds") {
+    int np = Bounds::getNumPointsFromPPD(rho_min, rho_max, N_per_decade_fine);
+    Bounds lRhoBounds(rho_min, rho_max, np, true, 0.0, rho_normal);
+    THEN("The lower and upper bounds are right") {
+      REQUIRE(std::abs(lRhoBounds.grid.min() - singularity::FastMath::log10(rho_min)) <=
+              1e-12);
```
jhp-lanl, line 68: You use `1e-12` in a lot of places for a tolerance. Two comments:
1. I would at least like this to be a `constexpr` defined parameter since it's repeated so much
2. It seems like the value might be predicated on the type of `Real`. With the idea of making things flexible, would it make sense to define
    ```c++
    constexpr REAL_TOL = std::numeric_limits<Real>::epsilon * 1e3
    ```
    instead?

I'll let this be a blanket comment for the entire file rather than pointing out every instance.

test/test_bounds.cpp
```diff
@@ -0,0 +1,130 @@
...
+    Bounds bnds(Bounds::ThreeGrids(), rho_min, rho_max, rho_normal, 0.5,
+                N_per_decade_fine, N_factor, N_factor, true);
+    THEN("The bounds are right") {
+      Real lrmin = singularity::FastMath::log10(rho_min);
+      Real lrmax = singularity::FastMath::log10(rho_max);
+      REQUIRE(std::abs(bnds.grid.min() - lrmin) <= 1e-12);
+      REQUIRE(std::abs(bnds.grid.max() - lrmax) <= 1e-12);
+      REQUIRE(bnds.grid.nGrids() == 3);
+      AND_THEN(
+          "The total number of points is less than a uniform fine spacing would imply") {
```
jhp-lanl, line 94: nice check!

jhp-lanl (approved):
Approved with comments and requested changes (all pretty minor).

I think the existing tests cover this, but please ensure that the tabular EOS changes are properly covered by our existing tests.

doc/sphinx/src/models.rst
jhp-lanl: I wrote this at the beginning of the review. The answer is yes, but I think this is a good step in the right direction.

sesame2spiner/generate_files.cpp
Yurlungur: Nope it can be constexpr!

sesame2spiner/generate_files.cpp
Yurlungur: Probably not but I am concerned about what happens if it's not defined for a given material. For example, a custom material someone tries to use. That's just me being cautious.

sesame2spiner/generate_files.cpp
Yurlungur: Yep :+1:

sesame2spiner/generate_files.cpp
Yurlungur: Hmm... I'm not sure. The EOSPAC class is a lot heavier than just what I load up here. But it would be less code. Would prefer to leave this for now if you don't mind.

There might be a rewrite down the line when `sesame2spiner` becomes a lot lighter and leverages the eospac class. So, let's wait for that.

sesame2spiner/io_eospac.hpp
Yurlungur: Under the hood, `Spiner::DataBox` supports any number of grids, but it must be set at compiel time. `SpinerEOS` and `sesame2spiner` can have different `NGRIDS` without fear, so long as the `NGRIDS` in `SpinerEOS` is greater than or equal to the number in `sesame2spiner`. I plan to unify this more down the road.

sesame2spiner/generate_files.cpp
Yurlungur: Agreed. I'll add const tags.

singularity-eos/base/table_bounds.hpp
Yurlungur: Nope. Removed.

singularity-eos/base/table_bounds.hpp
Yurlungur: I am reluctant to name this `spiner_utils` as I don't think that's really what this is. I'll stick it in a `table_utils` namespace. How's that?

singularity-eos/base/table_bounds.hpp
Yurlungur: Most of these arguments **cannot** be `const`. However, I think that is a strong reason to name the ones that are explicitly const.

test/test_bounds.cpp
Yurlungur: N_PER_DECADE_FINE = 200 and that T covers 9 decades. Calculated now.

test/test_bounds.cpp
Yurlungur: Done.

Yurlungur:
Thanks, @jhp-lanl !

> I think the existing tests cover this, but please ensure that the tabular EOS changes are properly covered by our existing tests.

Yep already on that. The existing tests do cover this on re-git. So it's very important those get run before final merge.

Yurlungur:
@dholladay00 please review. @chadmeyer please re-review or click approve depending.

Yurlungur:
(tests now all passing on re-git.)

sesame2spiner/generate_files.cpp
jhp-lanl: That's fine with me then. I just hate having to deal with the EOSPAC interface

singularity-eos/base/table_bounds.hpp
jhp-lanl: They only apply to tabular EOS built off of spiner though, right?

Yurlungur: Currently that's true... But that would put this in `SpinerEOS` as it's a util for that, not for `Spiner` as a library. 

That said, I think this is more broadly useful for, e.g., computing log bounds and things like that.

singularity-eos/base/table_bounds.hpp
```diff
@@ -0,0 +1,285 @@
...
+      os << "\t[ig,dx] = [" << ig << ", " << b.grid.dx(ig) << "]"
+         << "\n";
+    }
+    return os;
+  }
+
+  // This uses real logs
+  template <typename T>
+  static int getNumPointsFromPPD(Real min, Real max, const T ppd) {
+    constexpr Real epsilon = std::numeric_limits<float>::epsilon();
```
dholladay00, line 160: Why specifically `float` epsilon?

singularity-eos/base/table_bounds.hpp
```diff
@@ -0,0 +1,285 @@
...
+    max = toLog_(max, offset);
+
+    Real delta = max - min;
+    min += 0.5 * shrinkRange * delta;
+    max -= 0.5 * shrinkRange * delta;
+  }
+
+  static void adjustForAnchor_(const Real min, Real &max, int &N,
+                               const Real anchor_point) {
+    if (min < anchor_point && anchor_point < max) {
```
dholladay00, line 208: What are the implications if this condition isn't met, would it make sense to error?

singularity-eos/eos/eos_stellar_collapse.hpp
```diff
@@ -220,6 +220,20 @@ class StellarCollapse : public EosBase<StellarCollapse> {
...
+   public:
+    PORTABLE_INLINE_FUNCTION
+    LogT(const DataBox &field, const Real Ye, const Real lRho)
+        : field_(field), Ye_(Ye), lRho_(lRho) {}
+    PORTABLE_INLINE_FUNCTION Real operator()(const Real lT) const {
+      return field_.interpToReal(Ye_, lT, lRho_);
+    }
+
+   private:
+    const DataBox &field_;
```
dholladay00, line 233: I see that this is mostly a copy/paste move, but be warned of potentially cryptic compilation errors due to const class members.

test/compare_to_eospac.cpp
```diff
@@ -47,10 +47,13 @@
 
 #include <singularity-eos/eos/eos.hpp>
 
-using namespace Spiner;
+// using namespace Spiner;
```
dholladay00, line 50: remove commented out code if truly unnecessary.

singularity-eos/base/table_bounds.hpp
```diff
@@ -0,0 +1,285 @@
...
+      : Bounds(min, max, N, offset) {}
+
+  Bounds(Real min, Real max, int N, const bool convertToLog = false,
+         const Real shrinkRange = 0,
+         Real anchor_point = std::numeric_limits<Real>::signaling_NaN())
+      : offset(0), piecewise(true), linmin_(min), linmax_(max) {
+    if (convertToLog) {
+      convertBoundsToLog_(min, max, shrinkRange);
+      if (!(std::isnan(anchor_point))) {
+        anchor_point += offset;
```
dholladay00, line 60: is offset non-zero here?

singularity-eos/base/table_bounds.hpp
Yurlungur: That's a bug. Should be double.

singularity-eos/base/table_bounds.hpp
```diff
@@ -0,0 +1,285 @@
...
+ private:
+  Real toLog_(Real val, const Real offset) {
+    val += offset;
+    return singularity::FastMath::log10(std::abs(val));
+  }
+
+  void convertBoundsToLog_(Real &min, Real &max, const Real shrinkRange = 0) {
+    // Log scales can't handle negative numbers or exactly zero. To
+    // deal with that, we offset.
+    constexpr Real epsilon = std::numeric_limits<float>::epsilon();
```
dholladay00, line 191: is this a bug as well @Yurlungur ?

Yurlungur: Yep. Will fix.

singularity-eos/base/table_bounds.hpp
Yurlungur: If the condition isn't met, it will just not adjust the number of points or the max. The table will still be valid, but there won't be a grid point exactly at the anchor point. I'd like `sesame2spiner` to basically always run... so I'd rather this not error out. But I'll add a warning.

singularity-eos/eos/eos_stellar_collapse.hpp
Yurlungur: In this case it should be ok, and it was a copy/paste, but point taken. Will remove the `const`s.

test/compare_to_eospac.cpp
Yurlungur: Removed.

Yurlungur:
Thanks @dholladay00 I think I've addressed all your comments. Can you click approve if you agree? (I will not merge until post-change tests all pass on re-git)

singularity-eos/base/table_bounds.hpp
Yurlungur: Also renamed file.

singularity-eos/eos/eos_stellar_collapse.hpp
Yurlungur: This actually has to be `const` due to the `const` correctness elsewhere in the eos object.

chadmeyer (approved):
I think it looks good now for pushing.  Good Job @Yurlungur!

Yurlungur:
Thanks @chadmeyer @jhp-lanl @dholladay00 for your reviews! I now plan to merge as soon as the latest commit passes tests on re-git. Speak now if you have objections. :)

test/test_bounds.cpp
```diff
@@ -0,0 +1,134 @@
...
+#include <singularity-eos/base/fast-math/logs.hpp>
+#include <singularity-eos/base/spiner_table_bounds.hpp>
+
+#ifndef CATCH_CONFIG_FAST_COMPILE
+#define CATCH_CONFIG_FAST_COMPILE
+#include <catch2/catch_test_macros.hpp>
+#endif
+
+constexpr int NGRIDS = 3;
+using Bounds = singularity::table_utils::Bounds<NGRIDS>;
```
jhp-lanl, line 28: @Yurlungur since the `Bounds` class depends on `spiner`, can you wrap all of this in a `SINGULARITY_USE_SPINER` `#ifdef`?

I'm failing to build the tests when `SINGULARITY_USE_SPINER` is off.

I'll create an issue too.

PR #334, opened 2024-01-11
Add plugin infrastructure

Yurlungur:
## PR Summary

This PR adds a plugin infrastructure to `singularity-eos` so that at configure time, a user may request additional directories to be incorporated into singularity, which may contain additional equations of state, modifiers, C++ or fortran code, and tests. The `singularity` build system should be robust enough to incorporate these and build them and include them as tests.

I would like feedback on the current implementation if you get the chance. @dholladay00 @mauneyc-LANL . More to come soon.

This is currently a work in progress. Checklist:
- [x] Enable plugins to register header, source files and tests
- [x] Example plugin builds, with tests.
- [x] Enable plugins to specify the EOS variant to use
- [x] Thread machinery through spack
- [x] Check edge cases

## PR Checklist


Yurlungur:
@dholladay00 @mauneyc-LANL @jhp-lanl this is ready for review. The only missing piece now is the spack variant, which I will look at soon.

Yurlungur:
@gopsub this machinery will be used for the eos-extras stuff we discussed elsewhere. I will ping you offline to explain.

Yurlungur:
@rbberger I have added the plugin machinery to spack. Could you take a look and let me know what you think?

spack-repo/packages/singularity-eos/package.py
```diff
@@ -186,6 +193,14 @@ def cmake_args(self):
...
         if "+kokkos+cuda" in self.spec:
             args.append(self.define("CMAKE_CXX_COMPILER", self.spec["kokkos"].kokkos_cxx))
+        
+        if self.spec.satisfies("+plugins"):
+            PLUGIN_PATHS=os.environ.get("SINGULARITY_PLUGINS", "")
+            SINGULARITY_VARIANT=os.envirn.get("SINGULARITY_VARIANT", "")
+            if len(PLUGIN_PATHS) > 0:
+                args.append(self.define("SINGULARITY_PLUGINS", PLUGIN_PATHS))
+            if len(SINGULARITY_VARIANT) > 0:
+                args.append(self.define("SINGULARITY_VARIANT", SINGULARITY_VARIANT))
```
rbberger, line 203: Who sets the environment variables and when? Where does the plugins source code live? Instead of a Boolean variant, I would also imagine `plugins` to be something similar to `patches`.

```bash
spack install singularity-eos plugins=dir1:dir2:dir3 variant=some_path/to/another/file
```

I'm not sold on the idea to have this generalized in the public spackage. For the particular use case in mind, what's the downside of just having a inherited Spack package for singularity-eos and manipulating the CMake build and source directory for a particular plugin combo?

As in, Project XYZ has its own `singularity-eos` spackage that inherits from the upstream one, but grabs more source files, puts it in the right place and uses the CMake machinery to make it show up in the Singularity-EOS build.

Yurlungur: Do variants accept arbitrary strings? The reason I went with the environment variable path is that it wasn't clear from the spack documentation that this is possible. It is possible?

> I'm not sold on the idea to have this generalized in the public spackage. For the particular use case in mind, what's the downside of just having a inherited Spack package for singularity-eos and manipulating the CMake build and source directory for a particular plugin combo?

This is fine with me. The advantage of doing it here is that we only have to maintain one spackage. But maybe that's basically the case with the inheritence version as well? Since presumably the inherited package doesn't contain much?

cmake/plugins.cmake
```diff
@@ -0,0 +1,68 @@
...
+macro(export_source)
+  set(EOS_HEADERS ${EOS_HEADERS} PARENT_SCOPE)
+  set(_install_headers ${_install_headers} PARENT_SCOPE)
+  set(EOS_SRCS ${EOS_SRCS} PARENT_SCOPE)
+  set(PLUGIN_TESTS ${PLUGIN_TESTS} PARENT_SCOPE)
+  set(PLUGIN_INCLUDE_PATHS ${PLUGIN_INCLUDE_PATHS} PARENT_SCOPE)
+endmacro()
+macro(export_plugin)
+  list(APPEND PLUGIN_INCLUDE_PATHS ${CMAKE_CURRENT_SOURCE_DIR})
+  set(EOS_HEADERS ${EOS_HEADERS} PARENT_SCOPE)
```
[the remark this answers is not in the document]
Yurlungur: :+1: Done

spack-repo/packages/singularity-eos/package.py
Yurlungur: Ok I will go the spackage inheritence path for now.

Yurlungur:
@rbberger @dholladay00 @mauneyc-LANL @junghans  please re-review. I believe this is ready. As Richard suggests, I will attempt (as a first pass) to do plugins with spackage inheritence, rather than adding it to the spackage directly. However, I would like feedback on this approach. Is it better to add plugins as a variant? Or better to do spackage via inheritence?

CMakeLists.txt
```diff
@@ -396,6 +413,13 @@ endif() # SINGULARITY_USE_FORTRAN
 target_include_directories(
   singularity-eos PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>
                          $<INSTALL_INTERFACE:include>)
+target_include_directories(
+  singularity-eos PUBLIC $<BUILD_INTERFACE:${CMAKE_BINARY_DIR}/generated>)
+
+get_property(plugin_include_paths GLOBAL PROPERTY PLUGIN_INCLUDE_PATHS)
+foreach(path ${plugin_include_paths})
+  target_include_directories(singularity-eos PUBLIC $<BUILD_INTERFACE:${path}>)
+endforeach()
```
mauneyc-LANL, line 423: Shouldn't we have an install interface here? I think it'd just be the same as the `$<INSTALL_INTERFACE:include>`, but there's other scaffolding to manage (`INSTALL()` and `EXPORT()`)

mauneyc-LANL (commented):
I left some comments, but I need to give it a test drive on my side. We also need to being thinking about how spack integrates into this

CMakeLists.txt
```diff
@@ -359,27 +367,36 @@ if(SINGULARITY_BUILD_TESTS)
...
-foreach(_src ${EOS_SRCS})
-  list(APPEND _srcs singularity-eos/${_src})
-endforeach()
+# TODO(JMM): Kind of nice to have?
+get_property(eos_headers GLOBAL PROPERTY EOS_HEADERS)
+get_property(eos_srcs GLOBAL PROPERTY EOS_SRCS)
+if (SINGULARITY_VERBOSE_CONFIG)
+  message(STATUS "EOS Headers:\n\t${eos_headers}")
+  message(STATUS "EOS Sources:\n\t${eos_srcs}")
+endif()
```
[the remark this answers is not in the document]
Yurlungur: Oh cool. I will change this to integrate with the standard cmake version.

CMakeLists.txt
Yurlungur: I just hooked it into your current install machinery. If you look above, I add headers to the install headers list object. And for the source files, they all get linked into a single library anyway. That additional `foreach` that you comment on here is just to make sure in-tree/submodule mode builds work.

CMakeLists.txt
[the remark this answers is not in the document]
Yurlungur: Done

Yurlungur:
> I left some comments, but I need to give it a test drive on my side.

Cool please let me know at your earliest convenience. There is starting to be a need for the extras repo, so I want this plugin infrastructure merged as quickly as possible.

>  We also need to being thinking about how spack integrates into this

Yes---did you see the discussion with @rbberger above? We should discuss at the standup meeting this week. It seems that there are some differences of opinion about the right path forward.

spack-repo/packages/singularity-eos/package.py
rbberger: I'm not aware of a limitation of variants, they can be arbitrary strings. That does not mean they should be arbitrary complex strings. From my point of view, the main issue here is that spack covers runtime plugin systems (e.g. python/pypi), not compile time plugin systems.

Yurlungur: :+1: Either way is fine with me, so long as we have something that is useful broadly.

junghans (approved):

mauneyc-LANL:
Pushed a small change for install to work

Yurlungur:
> Pushed a small change for install to work

Thanks for the fix!

Yurlungur:
Got verbal approval from @mauneyc-LANL so I'm going to click the button now. @rbberger I think we can start on the spackage at this point.

rbberger:
Ok. On it.

Yurlungur:
correction. clicking button after tests pass

doc/sphinx/src/customization
```diff
@@ -0,0 +1,143 @@
...
+.. code-block:: cpp
+
+  #include <singularity-eos/eos.hpp>
+  using namespace singularity;
+  
+  using MyEOS_t = eos_variant<IdealGas, Gruneisen>;
+
+This will create a new type, ``MyEOS_t`` which contains only the
+``IdealGas`` and ``Gruneisen`` classes. (All of these live under the
+``singularity`` namespace.)
```
jhp-lanl, line 22: Do we want to add anything here about the helper functions @dholladay00 created for assembling complex type lists? I'd be fine with a small blurb to just go look at `eos.hpp` for examples.

doc/sphinx/src/customization
```diff
@@ -0,0 +1,143 @@
...
+the ``singularity-eos`` testing infrastructure.
+
+The ``CMakeLists.txt`` file registers the plugin with the build system
+via several custom ``cmake`` functions provided by
+``singularity-eos``. To register a header file (and most files in
+``singularity-eos`` should be header files) we use (for example)
+
+.. code-block::
+
+  register_headers(PLUGIN dust dust.hpp dust_variant.hpp)
```
jhp-lanl, line 80: I really like this

jhp-lanl (commented):
Sorry for the feedback after merge. My comments only concern documentation and can be wrapped into an issue for later.

Other than my comment on creating the type list, we should change the instructions for adding a new EOS that _specifically_ reference `eos.hpp` which has been renamed to `default_variant.hpp`. This should be a pretty easy grep and replace in the documentation.

PR #336, opened 2024-01-22
PowerMG EOS implemented and documented.

aematts:
## PR Summary

The Power Mie-Gruneisen EOS created at SNL by Robinson, adequate for high pressure modeling, is implemented and documented in relation to the linear Us-up Mie-Gruneisen implemented previously. This EOS is a full EOS with correct temperature. It is the least restrictive EOS of Mie-Gruneisen form and similar in spirit to the Extended Vinet EOS.

## PR Checklist


singularity-eos/eos/eos_powermg.hpp
```diff
@@ -0,0 +1,440 @@
...
+  if (_K0toK40[0] < 0.0) {
+    PORTABLE_ALWAYS_THROW_OR_ABORT("Required PowerMG model parameter K0 < 0");
+  }
+  if (_Pmin >= 0.0) {
+    _Pmin = -1000 * _K0toK40[0];
+#ifndef NDEBUG
+    PORTABLE_WARN(
+        "PowerMG model parameter Pmin not set or positive. Reset to default (-1000*K0)");
+#endif // NDEBUG
+  }
```
Yurlungur, line 165: This strikes me as a little odd... does the fact that this model supports negative pressures imply that it supports tension? I'm not familiar with PowerMG but usually Mie-Gruneisen does not support tension, right?

singularity-eos/eos/eos_powermg.hpp
```diff
@@ -0,0 +1,440 @@
...
+
+  // The PressureCoeffsK0toK40Size (=41) allowed K0 to K40 coefficients
+  // for the pressure reference curve vs rho
+  // are seldom all used so did not want to crowd the argument list with them.
+  // Instead I ask the host code to send me a pointer to this array so that I can
+  // copy it here. Not used coeffs should be set to 0.0 (of course).
+  for (int ind = 0; ind < PressureCoeffsK0toK40Size; ind++) {
+    _K0toK40[ind] = K0toK40input[ind];
+    if (_K0toK40[ind] != 0.0) _M = ind;
+  }
```
Yurlungur, line 178: `_M` is zero-initialized by default, obviously... but I think it would help for clarity to set it to zero above this loop.

singularity-eos/eos/eos_powermg.hpp
```diff
@@ -0,0 +1,440 @@
...
+}
+
+PORTABLE_INLINE_FUNCTION Real PowerMG::_HugPressureFromDensity(Real rho) const {
+  Real eta = 1.0 - robust::ratio(_rho0, rho);
```
Yurlungur, line 182: Looks like `eta` is fixed for the lifetime of the function from here, so I think it's nicer and more clear to mark it `const`.
```suggestion
  const Real eta = 1.0 - robust::ratio(_rho0, rho);
```
Actually you might consider implementing a private helper function for `eta` just to write less code. Something like (in private scope)
```C++
PORTABLE_INLINE_FUNCTION Real _GetEta(Real rho) const { return 1.0 - robust::ratio(_rho0, rho); }
```
Since other quantities are updated in the loop, marking the ones that aren't modified as `const` I think helps with clarity.

singularity-eos/eos/eos_powermg.hpp
```diff
@@ -0,0 +1,440 @@
...
+}
+PORTABLE_INLINE_FUNCTION Real PowerMG::_HugTemperatureFromDensity(Real rho) const {
+  Real eta = 1.0 - robust::ratio(_rho0, rho);
+  Real gamma2 = _SN2Mp2(_G0 * eta);
+  Real sum = _M * _K0toK40[_M] * gamma2;
+  for (int ind = _M - 1; ind >= 1; ind--) {
+    gamma2 = _G0 * eta / (ind + 2) * gamma2 + 1.0 / (ind + 2);
+    sum = eta * sum + ind * _K0toK40[ind] * gamma2;
+  }
+  Real temp = _K0toK40[0] / _Cv0 / 2.0 / _rho0 * eta * eta * eta * sum;
```
Yurlungur, line 198: Another minor clarity thing... Since `eta*eta*eta*sum` presumably is in the numerator here, I think it would help to put it before the divisions.

singularity-eos/eos/eos_powermg.hpp
```diff
@@ -0,0 +1,440 @@
...
+    ind = ind + 1;
+    ak = ak * x / ind;
+    ek = pf * ak;
+    sum = sum + ak;
+  }
+  if (ind >= maxind) {
+#ifndef NDEBUG
+    PORTABLE_WARN("Part of Hugoniot Temperature not converged");
+#endif // NDEBUG
+  }
```
Yurlungur, line 219: You don't need to change this (or maybe ever), but what are the implications of the temperature not converging? What kind of output will be produced?

singularity-eos/eos/eos_powermg.hpp
```diff
@@ -0,0 +1,440 @@
...
+  Real ak = 1.0 / ind;
+  Real sum = ak;
+  Real pf = exp(x) - 1.0;
+  Real ek = pf * ak;
+  while ((ek > sum * 1.e-15) && (ind < maxind)) {
+    ind = ind + 1;
+    ak = ak * x / ind;
+    ek = pf * ak;
+    sum = sum + ak;
+  }
```
Yurlungur, line 214: I'm not sure I understand what's going on with this function... Is this constructing Taylor series coefficients?

singularity-eos/eos/eos_powermg.hpp
```diff
@@ -0,0 +1,440 @@
...
+  PORTABLE_INLINE_FUNCTION Real SpecificHeatFromDensityInternalEnergy(
+      const Real rho, const Real sie, Real *lambda = nullptr) const {
+    return _Cv0;
+  }
+  // added for testing AEM Dec 2023
+  PORTABLE_INLINE_FUNCTION Real AllHugPressureFromDensity(const Real rho) const;
+  PORTABLE_INLINE_FUNCTION Real AllHugInternalEnergyFromDensity(const Real rho) const;
+  PORTABLE_INLINE_FUNCTION Real AllHugTemperatureFromDensity(const Real rho) const;
+  // Thermal Bulk Modulus added AEM Dec 2022
+  PORTABLE_INLINE_FUNCTION Real TBulkModulusFromDensityTemperature(
```
Yurlungur, line 75: No need to change this for now---but worth noting. If we ever expose thermal bulk mod, thermal expansion, etc., we may want to rename these. (Not on you. That's just me making a note for future reference.)

singularity-eos/eos/eos_powermg.hpp
```diff
@@ -0,0 +1,440 @@
...
+         _G0 * _G0 * _Cv0 * _rho0 * robust::ratio(_rho0, rho) * temp;
+}
+PORTABLE_INLINE_FUNCTION Real
+PowerMG::_compBulkModulusFromDensityTemperature(const Real rho, const Real temp) const {
+  Real eta = 1.0 - robust::ratio(_rho0, rho);
+  Real value = (1.0 - _G0 * eta / 2.0) * _K0toK40[0];
+  Real sum = 0.0;
+  for (int ind = _M; ind >= 1; ind--) {
+    sum = sum * eta + ind * _K0toK40[ind] * eta;
+  }
```
Yurlungur, line 289: It seems like this exact calculation is repeated a few times. Would there be benefit in making a helper function which computes this?

singularity-eos/eos/eos_powermg.hpp
```diff
@@ -0,0 +1,440 @@
...
+}
+// AEM: Give error since function is not well defined
+PORTABLE_INLINE_FUNCTION void
+PowerMG::DensityEnergyFromPressureTemperature(const Real press, const Real temp,
+                                              Real *lambda, Real &rho, Real &sie) const {
+  EOS_ERROR("PowerMG::DensityEnergyFromPressureTemperature: "
+            "Not implemented.\n");
+}
+// AEM: We should add entropy and Gruneissen parameters here so that it is complete
+// If we add also alpha and BT, those should also be in here.
```
Yurlungur, line 389: :+1: this is a good suggestion

singularity-eos/eos/eos_powermg.hpp
```diff
@@ -0,0 +1,440 @@
...
+  }
+  if (output & thermalqs::temperature)
+    temp = TemperatureFromDensityInternalEnergy(rho, sie);
+  if (output & thermalqs::specific_internal_energy) sie = sie;
+  if (output & thermalqs::pressure) press = PressureFromDensityInternalEnergy(rho, sie);
+  if (output & thermalqs::specific_heat)
+    cv = SpecificHeatFromDensityInternalEnergy(rho, sie);
+  if (output & thermalqs::bulk_modulus)
+    bmod = BulkModulusFromDensityInternalEnergy(rho, sie);
+}
```
Yurlungur, line 413: In Vinet, if I recall, you computed the free energy and then got everything from that, so the `FillEos` model could just compute the sum over coefficients once for all outputs. Is there a similar optimization you could do here? Where, e.g., you compute the sum over the coefficients once, and then get everything from that?

Yurlungur (approved):
Thanks for this, @aematts ! Assuming tests pass, this looks good to me. I have a few minor comments, but they are suggestions, and I consider them non-blocking. Feel free to push back on them.

singularity-eos/eos/eos_powermg.hpp
aematts: Tension is tricky and it seems more or less everyone has their own way of dealing with it. This is exactly how it is done in ALEGRA and in the Robinson write-up. See the documentation.

singularity-eos/eos/eos_powermg.hpp
aematts: Agree. Will change.

singularity-eos/eos/eos_powermg.hpp
aematts: Agree.

singularity-eos/eos/eos_powermg.hpp
aematts: Agree. Why I did it like this is because _K0toK40[0] / _Cv0 / 2.0 / _rho0 is a common refactor that shows up other places.

singularity-eos/eos/eos_powermg.hpp
aematts: Why I only do a WARN here is because if this happens I think a lot of other things would have broken as well. But since I set maxind myself, I think it is best to make sure it is sufficient. But I also do not want it to go on forever if something goes wrong.

aematts: So basically it indicates to the user that something is not right here. But maybe the not correct things is that *I* put in a limit that was too small. What do you think I should do?

singularity-eos/eos/eos_powermg.hpp
aematts: OK. This is described in Robinson's paper. It is to make it fast and accurate.
ek is the maximum of the reminder of the infinite sum so we cut the sum off when that is less that machine precision from the sum itself.

singularity-eos/eos/eos_powermg.hpp
aematts: I don't think so. Helper functions for something that fits on 3-4 lines I think is too much. But I am happy to hear your arguments.

singularity-eos/eos/eos_powermg.hpp
aematts: refactor = prefactor

singularity-eos/eos/eos_powermg.hpp
aematts: But as usual, something that needs to be changed in many places...

singularity-eos/eos/eos_powermg.hpp
aematts: OK. You did not like that ;).
I did not even think about doing it in that way. I basically followed how Robinson wrote it in the report. (referenced in the documentation). You want me to look into it?

singularity-eos/eos/eos_powermg.hpp
aematts: Done

singularity-eos/eos/eos_powermg.hpp
aematts: You have a few other things I do not like in that region, like negative bulk modulus (isothermal).

singularity-eos/eos/eos_powermg.hpp
aematts: Done

singularity-eos/eos/eos_powermg.hpp
aematts: This is the only major calculation for the temperature. The rest of the coifs are calculated from this via recursion relations. Really cool.

singularity-eos/eos/eos_powermg.hpp
Yurlungur: That makes sense. Maybe put the prefactor in parentheses? That should make it clear to a casual reader what's going on.

aematts:
OK. I have fixed a few small things (initiating _M, and adding const in front of Real eta = ....) and of course forgot to rerun the formatting.

singularity-eos/eos/eos_powermg.hpp
Yurlungur: This is a personal preference thing. I tend to add helper functions for everything, as that way if I make an error (or want to tweak or debug something) I only have to change one place in the not, e.g., 5. 

However, it's a trade-off of many small things... lines of code added, amount the reader needs to jump around, etc. You don't need to change this if you like it as is.

singularity-eos/eos/eos_powermg.hpp
Yurlungur: Yeah---this is something for later as it's a relatively big refactor.

singularity-eos/eos/eos_powermg.hpp
Yurlungur: I didn't like it in the smaller API calls (like pressure from density temperature) because computing everything for that call implies wasted work. However, it's a win in `FillEos` because you compute things once and then leverage them for all the return values `FillEos` can provide. That's actually why `FillEos` exists---so one can optimize like that if desired.

My thinking was just that math operations, e.g., summing over the Taylor series could be reused in `FIllEos` for a performance gain. But if it's not easy to do, don't bother now. It's not worth your time.

singularity-eos/eos/eos_powermg.hpp
Yurlungur: :+1: It's all fine as long as the behavior is documented and is clear.

aematts:
Last comments for today:
1) I could go through and make this more efficient by using one routine for everything, as I did in Vinet (Vinet_F_DT_func) this would avoid calculating the same thing several times in FillEos (as Jonah pointed out). But this is not how Robinson has written it in the report, so it would take some work. But I actually think it is a god idea for eos implementation since it exposes everything that is not consistent. Maybe we can put it up as a future wish.
2) The temperature convergence: in _SN2Mp2 I have set maxind to 200 so that the while loop would never be able to go on forever. I am pretty sure that this is sufficient for everything, but I put in a PORTBLE_WARN in case it was not. Is this an OK solution or should we discuss it more?
3) Tension: Always a tricky question and the solution Robinson has in his report is as good as anything, in my view. But it does give negative bulk modulus and such things. Maybe we should discuss this more, but then maybe in general terms to compare different solutions different people have to this.
That's it for today.

Yurlungur:
>     1. I could go through and make this more efficient by using one routine for everything, as I did in Vinet (Vinet_F_DT_func) this would avoid calculating the same thing several times in FillEos (as Jonah pointed out). But this is not how Robinson has written it in the report, so it would take some work. But I actually think it is a god idea for eos implementation since it exposes everything that is not consistent. Maybe we can put it up as a future wish.

Wishlist item is good. Don't worry about it for now. :)

>     2. The temperature convergence: in _SN2Mp2 I have set maxind to 200 so that the while loop would never be able to go on forever. I am pretty sure that this is sufficient for everything, but I put in a PORTBLE_WARN in case it was not. Is this an OK solution or should we discuss it more?

This is an OK solution for now. We can revisit if we actually hit the limit. You could move the limit to a global const (e.g., `static constexpr int MAXITER=200` in the class definition) if you like, that could serve as a useful reminder to look at it if we hit problems.

>     3. Tension: Always a tricky question and the solution Robinson has in his report is as good as anything, in my view. But it does give negative bulk modulus and such things. Maybe we should discuss this more, but then maybe in general terms to compare different solutions different people have to this.

This I'm not sure about and I'm curious what @jhp-lanl thinks. I worry a bit about the EOS returning things like negative bulk modulus, as a hydro code that receives something like that might go completely bonkers. I don't think this necessarily has to be solved now---it could be resolved, for example, with friendly testing from a host code, if we have people interested in using the EOS.

aematts:
Tension: Note that the Pmin is only used at very large tension. The expansion first go through a part that follows the isentrope. 
image

aematts:
In fact if you set a large negative value for Pmin (as the default) it never really kicks in but follows the isentrope all the way to infinite volume ;).

Yurlungur:
That figure looks totally reasonable. And it's not like our tabulated EOS's don't also have crazy stuff in tension.

singularity-eos/eos/eos_powermg.hpp
```diff
@@ -0,0 +1,441 @@
...
+#include <singularity-eos/base/robust_utils.hpp>
+#include <singularity-eos/eos/eos_base.hpp>
+
+namespace singularity {
+
+using namespace eos_base;
+using namespace robust;
+
+// COMMENT: This is a complete, thermodynamically consistent Mie-Gruneisen
+// EOS based on a linear Us-up Hugoniot, Us=Cs+s*up.
```
jhp-lanl, line 35: This EOS is a power-law series Us-up Hugoniot, right?

singularity-eos/eos/eos_powermg.hpp
jhp-lanl: Could you add a comment above the loop to the effect of what you said here? This sort of clever code behavior isn't obvious so I think it's useful for future developers if a comment indicates what the purpose is.

jhp-lanl (approved):
Some minor comments and suggested changes.

It's a shame the HTML can't render a PDF. I'm fine with the hyperlinks, so my suggestions are mainly to try to make it more obvious what the hyperlinks are doing and to separate them from the text.

singularity-eos/eos/eos_powermg.hpp
aematts: Whoops

doc/sphinx/src/models.rst
```diff
@@ -882,6 +885,10 @@ The first omitted term in the expansion inside the square brackets is :math:`\Ga
 in fact even better than the common approximation of replacing the full temperature on the Hugoniot with the temperature on the 
 isentrope, that is, the first term :math:`T_0 e^{\Gamma(\rho_0) \eta}`.
 
+.. image:: ../ApproxForTH.pdf
+  :width: 500
+  :alt: Different approximations for the temperature on the Hugoniot.
+
```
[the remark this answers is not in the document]
aematts: It renders fine in my Safari browser:
image
But I will add these suggestions anyway.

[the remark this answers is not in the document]
jhp-lanl: Ooooo that's good to know. This is how it looks in Firefox:

Screenshot 2024-01-31 at 12 17 27 PM

singularity-eos/eos/eos_powermg.hpp
aematts: Done

singularity-eos/eos/eos_powermg.hpp
aematts: Done

singularity-eos/eos/eos_powermg.hpp
aematts: Done

singularity-eos/eos/eos_powermg.hpp
aematts: Done

doc/sphinx/src/models.rst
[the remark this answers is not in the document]
aematts: Done

doc/sphinx/src/models.rst
```diff
@@ -895,6 +902,85 @@ where
 ``G0`` is :math:`\Gamma(\rho_0)`, ``Cv0`` is :math:`C_V`,
 ``E0`` is :math:`E_0`, and ``S0`` is :math:`S_0`. 
 
+Mie-Gruneisen power expansion EOS
+`````````````````````````````````
+As we noted above, the assumption of a linear :math:`U_s`- :math:`u_p` relation is simply not valid at large compressions. At 
+Sandia National Laboratories Z-pinch machine, the compression is routinely so large that a new Mie-Gruneisen EOS was developped,
+by `Robinson <PowerMG_>`_, that could handle these large compressions. The overall structure and motivation for approximations are as described above, in compression it is only the formula for :math:`P_H`, and by extension :math:`T_H`, that differ. This EOS is however modified in expansion to follow an isentrope instead of the invalid-in-expansion Hugoniot.
```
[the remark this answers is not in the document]
aematts: Done

doc/sphinx/src/models.rst
```diff
@@ -895,6 +902,85 @@ where
...
+.. math::
+    
+    K_0 &=& C_s^2 \rho_0 \ \ \ \ \ \ \ \ & \\
+    K_n &=& (n+1) s^n & \ \ \ \ \ \ n >= 1
+
+In the figure below we have used :math:`M=20` with these coefficients and show how the divergence in the MGUsup pressure at :math:`\eta = \frac{1}{s}` is avoided in the PowerMG, making it more suitable for modeling high pressures.
+
+.. image:: ../PMGvsMGUsupPress.pdf
+  :width: 500
+  :alt: Comparing Hugoniot pressure for PowerMG and MGUsup
```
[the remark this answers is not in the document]
aematts: Done

doc/sphinx/src/models.rst
```diff
@@ -895,6 +902,85 @@ where
...
+
+    P_H(\rho) = K_0 \eta \left( 1 + K_1 \eta + K_2 \eta^2 + K_3 \eta^3 + \cdots + K_M \eta^M \right)  
+
+By expanding the MGUsup Hugoniot pressure into a power series in :math:`\eta` we see that we can recover the MGUsup results by setting
+
+.. math::
+    
+    K_0 &=& C_s^2 \rho_0 \ \ \ \ \ \ \ \ & \\
+    K_n &=& (n+1) s^n & \ \ \ \ \ \ n >= 1
+
```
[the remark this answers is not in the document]
aematts: Thanks!

singularity-eos/eos/eos_powermg.hpp
aematts: Done

doc/sphinx/src/models.rst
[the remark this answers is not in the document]
aematts: Done

jhp-lanl:
I'm running a pipeline on re-git to make sure things work there

jhp-lanl:
> I'm running a pipeline on re-git to make sure things work there

Pipeline passes on re-git... merging

PR #348, opened 2024-02-22
ci: disable power9 and add roci jobs

rbberger:
## PR Summary

Closes #347 

## PR Checklist

- [ ] Adds a test for any bugs fixed. Adds tests for new features.
- [ ] Format your changes by using the `make format` command after configuring with `cmake`.
- [ ] Document any new features, update documentation for changes made.
- [ ] Make sure the copyright notice on any files you modified is up to date.
- [ ] After creating a pull request, note it in the CHANGELOG.md file.
- [ ] LANL employees: make sure tests pass both on the github CI and on the Darwin CI

If preparing for a new release, in addition please check the following:
- [ ] Update the version in cmake.
- [ ] Move the changes in the CHANGELOG.md file under a new header for the new release, and reset the categories.

PR #352, opened 2024-03-12
move the singularity_eos cpp files back into the fortran build only

Yurlungur:
## PR Summary

As @dholladay00 commented in PR #348 , the `singularity_eos.{cpp,hpp}` files should only be built when fortran is on to minimize build times for C++ codes. This reverts that change. 

## PR Checklist

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -10,7 +10,7 @@ variables:
   GIT_SUBMODULE_STRATEGY: normal
   SINGULARITY_GOLDFILES_VERSION: "goldfiles-1.8.0"
   SINGULARITY_EOS_GCC_VERSION: "10.3.0"
-  SINGULARITY_EOS_OPENMPI_VERSION: "4.1.3"
+  SINGULARITY_EOS_OPENMPI_VERSION: "4.1.1"
   SINGULARITY_EOS_CUDA_VERSION: "11.6.0"
   SINGULARITY_EOS_BUILD_TYPE: "RelWithDebInfo"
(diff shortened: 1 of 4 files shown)


dholladay00 (approved):

Yurlungur:
@rbberger please confirm this is ok. then will merge.

rbberger:
@Yurlungur do you have a re-git pipeline to show this works?

rbberger:
If there are no C++ source files in the singularity-eos_Library, we need to change it to an INTERFACE library in that case.

Yurlungur:
> @Yurlungur do you have a re-git pipeline to show this works?

No re-git was down

> If there are no C++ source files in the singularity-eos_Library, we need to change it to an INTERFACE library in that case.

This was why I had that dummy C++ file in there... should we revive that?

rbberger:
I think we can do it in a way which just avoids defining the _Library in the `NOT (CLOSURE && FORTRAN)` case. I'll need to look into that.

Yurlungur:
Sounds good. I'll hold off on merging this then. Do you just want to push to this branch?

rbberger:
Will do.

rbberger:
Pushed a change, need to wait until the CI gives us some feedback.

Yurlungur:
> Pushed a change, need to wait until the CI gives us some feedback.

When re-git comes back online do you want to push there? Or should I?

rbberger:
@Yurlungur there is a problem with this PR, since there are clients of the .cpp files other than Fortran. They are used in the unit tests.

Yurlungur:
> @Yurlungur there is a problem with this PR, since there are clients of the .cpp files other than Fortran. They are used in the unit tests.

Ah good point. What if we hide those added files/this cmake logic behind fortran *or* testing?

rbberger (approved):
This is ready from my point of view.

PR #358, opened 2024-04-12
Update closure model documentation

jhp-lanl:
## PR Summary

Updates the PTE solver documentation. Here is a summary of the additions/changes
- [x] Various minor notation adjustments to make things more clear
- [x] Reorganized things into a general discussion of the governing equations (separate from PTE) and the $`\bar{\rho}`$ formulation and the implication that volume fractions and densities are both inputs and outputs.
- [x] Reformulated the constraint equations to better reflect what's in the code. Reorganized the density-energy/density-temperature discussion to reflect the changes.
- [x] Added descriptions for the pressure/temperature known versions

## PR Checklist

- N/A Format your changes by using the `make format` command after configuring with `cmake`.
- N/A LANL employees: make sure tests pass both on the github CI and on the Darwin CI

### DIFF
--- a/doc/sphinx/src/using-closures.rst
+++ b/doc/sphinx/src/using-closures.rst
@@ -131,7 +131,7 @@ closure state.
     f_i = \frac{1}{N}
 
 Pressure-Temperature Equilibirum
-================================
+--------------------------------
 
 At present, ``singularity-eos`` focuses on several methods for finding a PTE
 solution, i.e. one where the pressures and temperatures of the individual
@@ -144,12 +144,12 @@ In essence, the PTE equations can be posed as two residual equations:
 .. math::
 
   f_\mathrm{tot} - \sum\limits_{i=0}^{N-1} f_i = 
-    \sum\limits_{i=0}^{N-1} f_i^*(x_i*, y_i*) - f_i(x_i, y_i)
+    \sum\limits_{i=0}^{N-1} f_i^*(x_i^*, y_i*) - f_i(x_i, y_i)
 
 .. math::
 
   u_\mathrm{tot} - \sum\limits_{i=0}^{N-1} u_i = 
-    \sum\limits_{i=0}^{N-1} u_i^*(x_i*, y_i*) - u_i(x_i, y_i)
+    \sum\limits_{i=0}^{N-1} u_i^*(x_i^*, y_i*) - u_i(x_i, y_i)
 
 where the superscript :math:`^*` denotes the variables at the PTE state,
 :math:`f` corresponds to the volume fractions, and :math:`u` to the energy
@@ -163,19 +163,19 @@ about the equilibrium states :math:`f_i^*(\rho_i, y_i)` and
 
 .. math::
 
-  f_\mathrm{tot} - \sum\limits_{i=0}^{N-1} f_i(\rho_i, y_i) \approx
-    \sum\limits_{i=0}^{N-1} (\rho_i^* - \rho_i)
-      \left(\frac{\partial f_i}{\partial \rho_i}\right)_{y_i}
+  f_\mathrm{tot} - \sum\limits_{i=0}^{N-1} f_i(x_i, y_i) \approx
+    \sum\limits_{i=0}^{N-1} (x_i^* - x_i)
+      \left(\frac{\partial f_i}{\partial x_i}\right)_{y_i}
     + \sum\limits_{i=0}^{N-1} (y_i^* - y_i)
-      \left(\frac{\partial f_i}{\partial y_i}\right)_{\rho_i}
+      \left(\frac{\partial f_i}{\partial y_i}\right)_{x_i}
 
 .. math::
 
-  u - \sum\limits_{i=0}^{N-1} u_i(\rho_i, y_i) \approx
-    \sum\limits_{i=0}^{N-1} (\rho_i^* - \rho_i)
-      \left(\frac{\partial u_i}{\partial \rho_i}\right)_{y_i}
+  u - \sum\limits_{i=0}^{N-1} u_i(x_i, y_i) \approx
+    \sum\limits_{i=0}^{N-1} (x_i^* - x_i)
+      \left(\frac{\partial u_i}{\partial x_i}\right)_{y_i}
     + \sum\limits_{i=0}^{N-1} (y_i^* - y_i)
-      \left(\frac{\partial u_i}{\partial y_i}\right)_{\rho_i},
+      \left(\frac{\partial u_i}{\partial y_i}\right)_{x_i},
 
 removing the dependence on the unknown equilibrium state. Minor manipulations
 are needed to recast the derivatives in terms of accessible thermodynamic
@@ -200,14 +200,44 @@ an addition :math:`N - 1` residual equations of the form
 
 .. math::
 
-  P_i(\rho_i, y_j) - P_j(\rho_j, y_j) = 0
+  P_i(\rho_i, y_i) - P_j(\rho_j, y_j) = 0,
 
-At this point, there are :math:`N + 1` equations and unknowns in the PTE sover.
-The choice of the second independent variable is discussed below and has
-implications for complexity of the numerics.
+
+which can be written as a Taylor expansion about the equilibrium state so that
+
+.. math::
+
+  P_i(\rho_i, y_j) - P_j(\rho_j, y_j)
+    = (f^*_i - f_i) \left(\frac{\partial P_i}{\partial f_i}\right)_{y_i}
+    + (y^*_i - y_i) \left(\frac{\partial P_i}{\partial y_i}\right)_{f_i} \\
+    - (f^*_j - f_j) \left(\frac{\partial P_j}{\partial f_j}\right)_{y_j}
+    - (y^*_j - y_j) \left(\frac{\partial P_j}{\partial y_j}\right)_{f_j}
+
+and typically the equations are written such that :math:`j = i + 1`.
+Additionally, the volume fractions are pure functions of density, so the volume
+contstraint equation can be rewritten simply as
+
+.. math::
+
+  f_\mathrm{tot} - \sum\limits_{i=0}^{N-1} f_i =
+    \sum\limits_{i=0}^{N-1} (f_i^* - f_i)
+
+where the other equations can also be re-written to be in terms of volume
+fractions instead of densities. Since the EOS returns derivatives in terms of
+density, these can be transformed to volume fraction derivatives via
+
+.. math::
+
+  \left(\frac{\partial Q}{\partial f_i}\right)_X 
+    = - \frac{\rho_i^2}{\rho}\left(\frac{\partial Q}{\partial \rho_i}\right)_X,
+
+where :math:`Q` and :math:`X` are arbitrary thermodynamic variables. At this
+point, there are :math:`N + 1` equations and unknowns in the PTE sover. The
+choice of the second independent variable is discussed below and has
+implications on the number of additional unknowns and equations.
 
 The Density-Energy Formulation
----------------------------------
+''''''''''''''''''''''''''''''
 
 One choice is to treat volume fractions and material energies as independent
 quantities. However, the material energies provide :math:`N - 1` additional
@@ -218,6 +248,19 @@ equations of the form
 .. math::
 
   T_i(\rho_i, \epsilon_j) - T_j(\rho_j, \epsilon_j) = 0.
+
+Again Taylor expanding about the equilibirum state, this results in a set of
+equations of the form
+
+.. math::
+
+  T_i(\rho_i, \epsilon_j) - T_j(\rho_j, \epsilon_j)
+    = (f^*_i - f_i) \left(\frac{\partial T_i}{\partial f_i}\right)_{\epsilon_i}
+    + (\epsilon^*_i - \epsilon_i) \
+        \left(\frac{\partial T_i}{\partial \epsilon_i}\right)_{f_i} \\
+    - (f^*_j - f_j) \left(\frac{\partial T_j}{\partial f_j}\right)_{\epsilon_j}
+    - (\epsilon^*_j - \epsilon_j)
+        \left(\frac{\partial T_j}{\partial \epsilon_j}\right)_{f_j}
 
 This leads to a total number of :math:`2N` equations and unknowns. The large
 matrix means that this algorithm is fairly hard to solve, and may not converge.
@@ -230,19 +273,11 @@ density-temperature formulation seems to be more stable and performant.
 In the code this is referred to as the ``PTESolverRhoU``.
 
 The Density-Temperature Formulation
-------------------------------------
+'''''''''''''''''''''''''''''''''''
 
 Another choice is to treat the temperature as an independent
 variable. Then the assumption of temperature equilibrium requires no additional
-equations, and the energy and volume fraction constraints take the form
-
-.. math::
-
-  f_\mathrm{tot} - \sum\limits_{i=0}^{N-1} f_i(\rho_i, T) \approx
-    \sum\limits_{i=0}^{N-1} (\rho_i^* - \rho_i)
-      \left(\frac{\partial f_i}{\partial \rho_i}\right)_{T}
-    + (T^* - T)\sum\limits_{i=0}^{N-1}
-      \left(\frac{\partial f_i}{\partial T}\right)_{\rho_i}
+equations, and the energy residual equation takes the form
 
 .. math::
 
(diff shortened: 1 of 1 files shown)


jhp-lanl:
See https://lanl.github.io/singularity-eos/jhp/docs/index.html to view my changes

jhp-lanl:
@Yurlungur and @aematts I think I have everything updated. Please review

doc/sphinx/src/using-closures.rst
```diff
@@ -3,126 +3,257 @@
...
+
+where the superscript :math:`^*` denotes the variables at the PTE state,
+:math:`f` corresponds to the volume fractions, and :math:`u` to the energy
+density (see the previous section for more information). In these equations,
+:math:`x` and :math:`y` represent some choice of independent thermodynamic
+variables.
+
+Then the energy and volume fraction constraint equations can be Taylor-expanded
+about the equilibrium states :math:`f_i^*(\rho_i, y_i)` and
+:math:`u_i^*(\rho_i, y_i)` so that they become
```
Yurlungur, line 163: I dunno if I agree that we are assuming a Taylor expansion. The Taylor expansion is used by the Newton-Raphson solve to compute local gradients, but we are solving for the root of a function generically. The linear solve is a *piece* of the generic solve.

doc/sphinx/src/using-closures.rst
```diff
@@ -167,6 +298,23 @@ objects for each equation of state, similar to the vector API for a
...
+
+  It bears repeating: **both the volume fractions and densities act as inputs
+  and outputs**. They are used to define the internal :math:`\bar
+  {\rho}_i` variables at the beginning of the PTE solve. The volume fractions
+  and densities at the end of the PTE solve will represent those for the new
+  PTE state.
+
+.. warning::
+
+  The PTE solvers ***require*** that all input densities and volume fractions
```
Yurlungur, line 396: One extra pair of stars
```suggestion
  The PTE solvers **require** that all input densities and volume fractions
```

Yurlungur (approved):
Thanks for adding this more detailed documentation, @jhp-lanl !

doc/sphinx/src/using-closures.rst
jhp-lanl: I don't really understand your distinction. My understanding of a Newton solve is that it is derived from the Taylor expansion of the residual function about the root.

doc/sphinx/src/using-closures.rst
jhp-lanl: I was going for italics but forgot which markdown convention Sphinx uses. However, I think that's probably not necessary. Committed your suggestion as-is

doc/sphinx/src/using-closures.rst
Yurlungur: I guess I'm concerned that this implies that the *global* root is computed using the *local* taylor expansion. However, that's not the case. We re-compute the taylor expansion at each iteration of the outer loop of the solve, so that the global solution is not limited to the local tangent space.

jhp-lanl: Ahhhh I see. Okay I'll try to reword things to better reflect that we are still solving the fully-non-linear system and that the non-linear solve is built off of the linearization via Taylor expansion.

jhp-lanl: @Yurlungur see if you like my new wording. I tried to highlight more clearly that this step is needed for the linearization of the problem in preparation for a Newton-Raphson method.  (docs may take a bit to build)

Yurlungur:
@aematts how does this look to you? If you're happy I'll merge.

aematts:
I am looking at it now. Sorry about holding you up but I was on my Frday off and took my husband to the neurologist Monday.

Yurlungur:
> I am looking at it now. Sorry about holding you up but I was on my Frday off and took my husband to the neurologist Monday.

Not at all--there's no rush!

doc/sphinx/src/using-closures.rst
```diff
@@ -3,131 +3,362 @@
...
+  P_i = F(\rho, \epsilon, \mu_1, ..., \mu_i, ..., \mu_{N-1}) \\
+  \rho_i = G(\rho, \epsilon, \mu_1, ..., \mu_i, ..., \mu_{N-1}) \\
+  \epsilon_i = H(\rho, \epsilon, \mu_1, ..., \mu_i, ..., \mu_{N-1}) \\
+  T_i = J(\rho, \epsilon, \mu_1, ..., \mu_i, ..., \mu_{N-1})
 
-The average density in a given material-occupied volume is
+where each material has its own density, :math:`\rho_i`, and specific internal
+energy, :math:`e_i`, as well as in principle its own pressure and temperature.
+Each will be a function of the bulk density, :math:`\rho`, the bulk specific
+internal energy, :math:`\epsilon`, and both the mass fraction, :math:`\mu_j`,
```
aematts, line 36: change "both" to "all"

doc/sphinx/src/using-closures.rst
```diff
@@ -3,131 +3,362 @@
...
+of all materials present.
+
+For convenience of minimizing input arguments though, the solvers in
```
aematts, line 39: Divide it up in \bar{rho} and a guess for f_i would be better I think. Are we really sure we want to stay with these two coupled variables for the future? It is much easier to rewrite the code now than later when we have connected a bunch of host codes that uses this.

doc/sphinx/src/using-closures.rst
```diff
@@ -3,131 +3,362 @@
...
+          \rho_1, ..., \rho_i, ..., \rho_{N-1}) \\
+  \epsilon_i = H(\epsilon, f_\mathrm{tot}, f_1, ..., f_i, ..., f_{N-1},
+          \rho_1, ..., \rho_i, ..., \rho_{N-1}) \\
+  T_i = J(\epsilon, f_\mathrm{tot}, f_1, ..., f_i, ..., f_{N-1},
+          \rho_1, ..., \rho_i, ..., \rho_{N-1})
+
+
+The important nuance here is the **the volume fractions and densities are both
+inputs and outputs** in the current ``singularity-eos`` formulation of the
+closure models. From physics perspective this can be confusing, but from a code
```
aematts, line 102: If you insist on these input variables you have to add how they are dependent on each other, that is that \overline{rho} is constant. It cannot be repeated enough.

doc/sphinx/src/using-closures.rst
```diff
@@ -3,131 +3,362 @@
...
+  .. math::
+
+    \sum\limits_{i=0}^{N-1} f_i = f_\mathrm{tot}.
+
+  One such assumption would be to **equipartition** the volume between all
+  materials such that
+
+  .. math::
+
+    f_i = \frac{1}{N}
```
aematts, line 132: This doesn't work for most of the test I have run. It is much more sensitive than that. For the 3-phase problem/test case I made I had to start much more close to the final results, and even the 2-phase problems/test cases I made it would have been too crude if it wasn't for the fact that the final results were close to 50/50.

Please add that the previous time step's volume fractions might be a good guess.
Or, if you have a god sense on the individual densities of the phases, putting these in as a guess and then set the volume fractions accordingly (as I showed in my write up) might be better.

It took me quite a while to find settings that worked.

doc/sphinx/src/using-closures.rst
```diff
@@ -3,131 +3,362 @@
...
 
-and thus the density averaged over all materials is
+Pressure-Temperature Equilibirum
+--------------------------------
+
+At present, ``singularity-eos`` focuses on several methods for finding a PTE
+solution, i.e. one where the pressures and temperatures of the individual
+materials are all the same. The methods presented differ in what they treat as
+independent variables, and thus what precise system of equations they solve.
+However they all share the above mathematicaly formulation.
```
aematts, line 141: correct mathematical/ly

doc/sphinx/src/using-closures.rst
```diff
@@ -3,131 +3,362 @@
...
----------------------------------
+''''''''''''''''''''''''''''''
 
-One choice is to treat volume fractions and material energies as
-independent quantities. The material energies provide :math:`N` more
-unknowns. The equality of pressures provides :math:`N-1` additional
-constraints. Additionally, the euqality of material temperatures, evaluated as
+One choice is to treat volume fractions and material energies as independent
+quantities. However, the material energies provide :math:`N - 1` additional
+unknowns. This requires that euqality of material temperatures satisfy the
```
aematts, line 259: correct euqality

doc/sphinx/src/using-closures.rst
```diff
@@ -167,6 +398,23 @@ objects for each equation of state, similar to the vector API for a
 given EOS. Here the indexers/vectors are not over cells, but
 materials.
 
+.. warning::
+
+  It bears repeating: **both the volume fractions and densities act as inputs
+  and outputs**. They are used to define the internal :math:`\overline
```
aematts, line 404: An internal variable is not necessarily constant, so add constant somewhere to underscore that THIS is how the two input variables are coupled.

doc/sphinx/src/using-closures.rst
```diff
@@ -167,6 +398,23 @@ objects for each equation of state, similar to the vector API for a
...
+  PTE state.
+
+.. warning::
+
+  The PTE solvers **require** that all input densities and volume fractions
+  are non-zero. As a result, ``nmat`` refers to the number of *participating*
+  materials. The user is encouraged to wrap their data arrays using an
+  ``Indexer`` concept where, for example, three paricipating PTE materials
+  might be indexed as 5, 7, 20 in the material arrays. This requires overloading
+  the square bracket operator to map from PTE idex to material index.
```
aematts, line 416: My question is, how do we know what the participating materials are? It is easy to say that it is when \overline{rho} is non-zero, but how do we know that \overline{rho} became non-zero in the last time step? Or that \overline{rho} became zero and a material/phase vanished? Is it \overline{rho} that is updated in the timestep? Please educate me.

aematts:
OK @jhp-lanl  and @Yurlungur. I have gone through it once now. Your turn.

doc/sphinx/src/using-closures.rst
jhp-lanl: Changed to
```markdown
and the mass fractions, :math:`\mu_j`, of all other materials present.
```

doc/sphinx/src/using-closures.rst
jhp-lanl: No I don't think we're bound to this input specification at all, and I think specifying `rho_bar` directly is an excellent idea since it clearly differentiates inputs from outputs.

I agree that it will be easier to change things sooner rather than later, but since we already have codes that depend on the current input specification, I'd imagine that we'd offer both the existing specification as well as a newer, less confusing one in the future. In fact, I'd love to rewrite the `get_sg_eos()` functions to use what you suggest in the future.

Right now I don't really have the bandwidth to tackle this though, so I'm afraid I'll have to punt on code changes here. I'll create an issue though:
https://github.com/lanl/singularity-eos/issues/360

doc/sphinx/src/using-closures.rst
jhp-lanl: I repeated the total density equation in the warning block to re-emphasize this point

doc/sphinx/src/using-closures.rst
Yurlungur: I want to think about this a little bit more before we jump on this change. I agree this suggestion is less confusing, but we should think about exactly the optimal set of inputs/outputs for a host code.

aematts:
One more thing: We might want to point out that if you use sesame tables they should NOT be inverted if you use the pressure-temperature formalism, but they SHOULD BE inverted if you use the density-energy formalism.

aematts:
It is all about what gives you the best derivatives.

aematts:
Fixed pressure should also be done on original tables (NOT inverted).
Fixed temperature doesn't matter but using the original tables would probably be better, avoiding inversion when it is not needed.

doc/sphinx/src/using-closures.rst
jhp-lanl: thanks for the feedback! I added a note about using the previous timestep's values to the top of this block and added a discussion about different assumptions that can be made.

A better assumption is just to use the mass fractions, and yet a better assumption is to weight the mass fractions by the material bulk moduli. I added this to the discussion as well.

doc/sphinx/src/using-closures.rst
jhp-lanl: fixed

doc/sphinx/src/using-closures.rst
jhp-lanl: fixed

Yurlungur:
> One more thing: We might want to point out that if you use sesame tables they should NOT be inverted if you use the pressure-temperature formalism, but they SHOULD BE inverted if you use the density-energy formalism.

These are good rules of thumb to recommend, though I'm sure there are exceptions that prove the rule.

doc/sphinx/src/using-closures.rst
jhp-lanl: Deciding which materials participate is a very important question, and one we struggle with in an Eulerian frame quite a bit. Many Eulerian codes have some sort of non-zero mass fraction limit that limits participation in the PTE solver, but with a good volume fraction guess it might be better to use a small volume fraction cutoff. In general this is left up to the host codes, but we could imagine tools in the future to aid in this decision. I've seen cases where mixed cells will drive small volume fractions of materials up to densities well over 100 g/cc only for the density to be completely reasonable in the next cycle.

To your second question, the hydro code I think is free to update as it sees fit, i.e. updating something like `$\overline{\rho}$` (which is basically equivalent to updating the masses directly), or maybe the physical densities themselves. I think adopting your suggestion to use `$\overline{\rho}$` as an input would be natural for many hydro codes.

doc/sphinx/src/using-closures.rst
aematts: Maybe you can discuss this at the point you first talk about \overline{bar}? Pointing out that holding it constant means that the density is constant but that both rho_i and f_i can change as long as their product is constant.

jhp-lanl:
> One more thing: We might want to point out that if you use sesame tables they should NOT be inverted if you use the pressure-temperature formalism, but they SHOULD BE inverted if you use the density-energy formalism.

I have a note in there about the need to invert from density-temperature space to density-energy space for the density-energy formulation. I mentioned that there are both performance and accuracy implications (I'm not sure how much I trust the inverted derivatives for example).

doc/sphinx/src/using-closures.rst
aematts: The density equation itself doesn't say that \overline{bar} is constant since only the sum of all of them need to be constant.

doc/sphinx/src/using-closures.rst
Yurlungur: I am not sure we want to be in charge of which materials participate and which don't. IMO this is a question for the host code, as different codes want to do different things. In riot, for example, we experimented with this quite a bit. We now iteratively remove the material with the smallest estimated volume fraction until PTE converges.

jhp-lanl:
I'll merge once the docs build and I see they look okay. Thanks for the review @aematts and @Yurlungur !

doc/sphinx/src/using-closures.rst
jhp-lanl: fixed

doc/sphinx/src/using-closures.rst
jhp-lanl: I'll repeat it there as well

doc/sphinx/src/using-closures.rst
aematts: So, when it comes to KPT the best would be using the mass fractions since that is what gets updated with time. 
For computational reasons we should stay at what gives the best derivatives. Nothing will improve convergence better than using the best possible derivatives.

Yurlungur: > So, when it comes to KPT the best would be using the mass fractions since that is what gets updated with time.

This makes sense---that also motivates using rhobar over rho, since that's what Eulerian codes update with time.

> For computational reasons we should stay at what gives the best derivatives. Nothing will improve convergence better than using the best possible derivatives.

This is a discussion for another time but... while this seems intuitively right, but is that always true? For example, for tables that combine multiple phases together, are the "best" derivatives not discontinuous at phase transitions almost by definition?

doc/sphinx/src/using-closures.rst
jhp-lanl: I'm not advocating for a fixed prescription, but just as we provide modifiers to EOS, I think it's reasonable to maybe in the future provide a PTE filter if you will. Something like this could potentially automate the creation of indexer objects for instance.

I'll also note that at some level we already do make this decision in `get_sg_eos()`. We just use zero mass fraction as the cutoff, which is pretty obvious and sane. However, since our PTE solver doesn't support zero mass volume fractions or densities, it may be prudent to generalize this machinery.

doc/sphinx/src/using-closures.rst
jhp-lanl: Just to be clear, I'm not proposing changing the existing inputs. I'm proposing modifying the classes to accept different multiple input specifications.

aematts (approved):
Looks good now.

Issue #57, opened 2021-12-21
spack info does not describe variants


Yurlungur:
The following command sequence 
```bash
spack repo add singularity-eos/spack-repo
spack info singularity-eos
```
produces the following meaningless help message
```bash
(base) jonahm@cn135:$ spack info singularity-eos 
CMakePackage:   singularity-eos

Description:
    None
==> Error: can only concatenate str (not "NoneType") to str
```
How do we get this to list variants, etc? @mauneyc-LANL

Issue #56, opened 2021-12-21
Re-enable in-tree builds of kokkos


Yurlungur:
I am getting very annoyed by the need to compile kokkos ahead of time. I want to re-enable in-tree builds.

Yurlungur:
One nice solution might be `cmake`'s fetch functionality. Just download kokkos if it's not available.

Yurlungur:
I was mistaken... this capability was never removed. Closing the issue.

Issue #55, opened 2021-12-21
Parser should provide useful warnings


Yurlungur:
The parser should check for mis-spelled or otherwise erroneous inputs and provide meaningful warnings.

PR #54, opened 2021-12-21
Modification w.r.t. Darwin CentOS8 Upgrade

ktsai7:
## PR Summary

- update clang-format dependency
- only do `pages_check`, not `pages` when it is triggered pipelines in CI
- use `singularity-eos` instead of `singularity-eos-deps` spackage
- some cleanups and reorg in `.gitlab-ci.yml`

## PR Checklist

- [x] Adds a test for any bugs fixed. Adds tests for new features.
- [x] Format your changes by using the `make format` command after configuring with `cmake`.
- [x] Document any new features, update documentation for changes made.

PR #361, opened 2024-04-17
1) 3 Tests for testing PTESolverRhoT versus Ann's Flag PTsolv 2) Implementing Carl Grief's mass fraction update model.

aematts:
## PR Summary

1) Created and added 3 tests for making sure PTESolverRhoT gives the same as Ann's Flag PTsolv.
2) Implementing a skeleton scheme for introducing a Kinetic PhaseTransition scheme into host codes.
In the first round we only implement ingredients one by one and assemble them in test cases. Later we might
introduce classes, move subroutines around for better structure, and add more advanced C++ features (for example Indexers) to make future expansions/variations easier to handle. The division into host and singularity-eos code is floating for the moment.

## PR Checklist

- [ x] Adds a test for any bugs fixed. Adds tests for new features.
- [ x] Document any new features, update documentation for changes made.


jhp-lanl:
Thanks for putting this up @aematts and thanks for digging into what isn't necessarily intuitive C++ in a lot of this!

> @jhp-lanl and @Yurlungur here is the WIP PR that I promised. This is a real WIP since I have several issues that I need help with. One issue is: the 2phase cpp file can use two header files pte_longtest_2phase...... and pte_test_2phase... Is it possible to make two test cases out of this without duplicating the cpp file? Then the tests seem to run using ctest -R name -v on Darwin (skylake-gold) but at least one of the tests here (Minimal) has failed (so far).

If I understand your question, I think the answer is namespaces. It looks like you use a lot of the same names in both the headers so an easy way to differentiate between the two would be to wrap the body of each header in a namespace.

So something like this in the header
```c++
namespace longtest_2phaseVinetSn {
constexpr int NMAT = 2;
constexpr int NTRIAL = 20;
constexpr int NPTS = NTRIAL * NMAT;
...
} \\ namespace 

And then when you want to use it in the actual test, you could do something like this:
```c++
longtest_2phaseVinetSn::set_eos(eos_hv.data());
```

I didn't take the time to track down what exactly you want from the header files, but this was the first example I could find where it might be different.

I think there is a lot of duplicate code between the headers as well that you may want to break out into a common header that they can all use. For example are the indexers all different? They didn't look different but it's possible I missed something.

aematts:
> Thanks for putting this up @aematts and thanks for digging into what isn't necessarily intuitive C++ in a lot of this!
> 
> > @jhp-lanl and @Yurlungur here is the WIP PR that I promised. This is a real WIP since I have several issues that I need help with. One issue is: the 2phase cpp file can use two header files pte_longtest_2phase...... and pte_test_2phase... Is it possible to make two test cases out of this without duplicating the cpp file? Then the tests seem to run using ctest -R name -v on Darwin (skylake-gold) but at least one of the tests here (Minimal) has failed (so far).
> 
> If I understand your question, I think the answer is namespaces. It looks like you use a lot of the same names in both the headers so an easy way to differentiate between the two would be to wrap the body of each header in a namespace.
> 
> So something like this in the header
> 
> ```c++
> namespace longtest_2phaseVinetSn {
> constexpr int NMAT = 2;
> constexpr int NTRIAL = 20;
> constexpr int NPTS = NTRIAL * NMAT;
> ...
> } // namespace longtest_2phaseVinetSn
> ```
> 
> And then when you want to use it in the actual test, you could do something like this:
> 
> ```c++
> longtest_2phaseVinetSn::set_eos(eos_hv.data());
> ```
> 
> I didn't take the time to track down what exactly you want from the header files, but this was the first example I could find where it might be different.
> 
> I think there is a lot of duplicate code between the headers as well that you may want to break out into a common header that they can all use. For example are the indexers all different? They didn't look different but it's possible I missed something.

Good idea with one common header file for the things that appear in both header files. It is only the data sets that differ between the header files.
But the namespace solution that you suggests for the two header files is not what I am looking for.

What I am looking for is for the compiler to create two tests: 1) one with the _test_ header file and test_pte_2phase.cpp and 2) one with the _longtest_ header file and test_pte_2phase.cpp. 

I will check on what failed in the checks.

jhp-lanl:
> What I am looking for is for the compiler to create two tests: 1) one with the test header file and test_pte_2phase.cpp and 2) one with the longtest header file and test_pte_2phase.cpp.

Ahhhhh sorry I misunderstood.

Really the best solution is to not mirror the existing PTE test since it's not really a well-written test and just use the Catch test framework we use for all the other tests. This way, you can define things common to both tests at a higher level in the test (see the Gruneisen test you wrote as an example) and then have lower levels specify the things that are different.

One solution with minimal code changes would be to change each test from having a `main()` function without arguments to some other function that takes the arguments you want to change in each test. Then this could also become a header and then the `.cpp` file would just wrap the function in a `main()` and call with the different parameter sets, maybe propagating the return value so that `main()` will return zero only if both tests pass.

test/CMakeLists.txt
```diff
@@ -99,6 +99,20 @@ if(SINGULARITY_BUILD_CLOSURE)
...
     add_test(pte test_pte)
   endif()
+  if(SINGULARITY_USE_SPINER)
+    add_executable(test_pte_2phase test_pte_2phase.cpp)
+    target_link_libraries(test_pte_2phase PRIVATE Catch2::Catch2
+                                         singularity-eos::singularity-eos)
+    target_include_directories(test_pte_2phase PRIVATE ${CMAKE_CURRENT_SOURCE_DIR})
+    add_test(pte_2phase test_pte_2phase)
+  endif()
+   if(SINGULARITY_USE_SPINER)
```
Yurlungur, line 109: I think these two if statements can be combined. :) 
```suggestion
```

test/pte_longtest_2phaseVinetSn.hpp
```diff
@@ -0,0 +1,135 @@
...
+#include <ports-of-call/portability.hpp>
+#include <ports-of-call/portable_arrays.hpp>
+#include <singularity-eos/eos/eos.hpp>
+
+constexpr int NMAT = 2;
+constexpr int NTRIAL = 20;
+constexpr int NPTS = NTRIAL * NMAT;
+constexpr int HIST_SIZE = 10;
+
+constexpr Real in_rho_tot[NTRIAL] = {
```
Yurlungur, line 29: For future us it might be useful to have a comment where these numbers come from. Are they from an equivalent test in flag or something like that?

test/pte_test_2phaseVinetSn.hpp
```diff
@@ -0,0 +1,91 @@
...
+#include <ports-of-call/portability.hpp>
+#include <ports-of-call/portable_arrays.hpp>
+#include <singularity-eos/eos/eos.hpp>
+
+constexpr int NMAT = 2;
+constexpr int NTRIAL = 5;
+constexpr int NPTS = NTRIAL * NMAT;
+constexpr int HIST_SIZE = 10;
+
+constexpr Real in_rho_tot[5] = {7.27800000, 7.27981950, 7.28163945, 7.28345986,
```
Yurlungur, line 29: Same as above... a comment about where these came from would be nice.

Yurlungur (commented):
Thanks for adding these @aematts ! 

So if I understand your question above---the issue is that `test_pte_*phase.cpp` is identical for both CPP files just with different headers that define `set_eos` is that right? And the function signature of `set_eos` is the same and `Indexer2D is the same between all the files?

I think the first step to cleaning that up a bit is to define a void function that does all the stuff in side the test cpp file between `Kokkos::initialize` and `Kokkos::finalize`. It would have a signature of something like this:

```C++
template<typename SetEos_t>
void TestPTE(const SetEos_t &set_eos) {
  /* all the stuff thats currently in there *./
}
```

Then you can call this function twice in the same cpp file with the two different `set_eos` functions defined in your headers, so long as they're named slightly differently. You can name one `set_eos_2pte` and one `set_eos_3pte`, or use namespaces as @jhp-lanl suggested. So that would look something like this:

```C++
test_pte(2phase::set_eos);

test_pte(3phase::set_eos);
```

From there, we can think about adding the catch2 stuff. But I think this gets you most of what you want. Does that seem like a good path?

aematts:
I just pushed my first cleanup attempt. My plan is to write one .cpp for all pte tests. But I have not yet fixed test_pte_{2,3}phase.cpp (and hopefully even test_pte.cpp) to be the same, I have to think a little more about it. But in the end I want to use one .hpp file (or rather namespace) for each test and line the tests up in one .cpp file like the other tests, using a common testPTE function.

aematts:
To do next: Fix text output to be outside of what will be testPTE, and use same input format (mainly trial_vfrac[NMAT][NTRIAL] and in_lambda[NMAT][NTRIAL]) for all calls to testPTE. Then it should be ready for breaking out testPTE.

aematts:
@Yurlungur , would you mind if I put test_pte into this new framework? That means changing set_state to not have t as input parameter but instead n, the trial number. But it should be the same, just setting t=600.0 as an input in the namespace and get sie from it in in the .hpp file.

aematts:
I will fix the formatting and add a comment on where the data is from before I leave this for a week or so. I will also try to figure out why I get this now:
/home/runner/work/singularity-eos/singularity-eos/test/pte_test_5phaseSesameSn.hpp:69:42: error: ‘EOSPAC’ is not a member of ‘singularity’
   69 |   singularity::EOS Snbeta = singularity::EOSPAC(SnbetaID);
      |                                          ^~~~~~

Yurlungur:
> @Yurlungur , would you mind if I put test_pte into this new framework? That means changing set_state to not have t as input parameter but instead n, the trial number. But it should be the same, just setting t=600.0 as an input in the namespace and get sie from it in in the .hpp file.

I wouldn't mind at all. I think that would be a nice cleanup!

jhp-lanl:
> I will fix the formatting and add a comment on where the data is from before I leave this for a week or so. I will also try to figure out why I get this now: /home/runner/work/singularity-eos/singularity-eos/test/pte_test_5phaseSesameSn.hpp:69:42: error: ‘EOSPAC’ is not a member of ‘singularity’ 69 | singularity::EOS Snbeta = singularity::EOSPAC(SnbetaID); | ^~~~~~

In case you haven't fixed this, this to me seems like maybe EOSPAC isn't being slurped in for the testing either because it's not being detected in the environment or because it's not enabled in the build options.

aematts:
I fixed it. Thanks.

Ann

From: Jeff Peterson ***@***.***>
Reply-To: lanl/singularity-eos ***@***.***>
Date: Thursday, May 2, 2024 at 6:43 PM
To: lanl/singularity-eos ***@***.***>
Cc: Ann Wills ***@***.***>, Mention ***@***.***>
Subject: [EXTERNAL] Re: [lanl/singularity-eos] WIP: 3 Tests for testing PTESolverRhoT versus Ann's Flag PTsolv (PR #361)


I will fix the formatting and add a comment on where the data is from before I leave this for a week or so. I will also try to figure out why I get this now: /home/runner/work/singularity-eos/singularity-eos/test/pte_test_5phaseSesameSn.hpp:69:42: error: ‘EOSPAC’ is not a member of ‘singularity’ 69 | singularity::EOS Snbeta = singularity::EOSPAC(SnbetaID); | ^~~~~~

In case you haven't fixed this, this to me seems like maybe EOSPAC isn't being slurped in for the testing either because it's not being detected in the environment or because it's not enabled in the build options.

aematts:
The three new files 
    closure/kinetic_phasetransition_models.hpp
    closure/kinetic_phasetransition_utils.hpp
    closure/kinetic_phasetransition_methods.hpp
now contains all that is needed to put together a strawman KPT framework.
The file test/test_kpt_models.cpp contains a few tests that together cover the chain from given Gibbs free energies for 5 phases to updated given mass fractions.

aematts:
1) this needs to be combined with the pte solver getting a new state for the new mass fractions and then give the Gibbs free energies for the phases in this new state. test/test_pte_2phase.cpp and test/test_pte_5phase.cpp can be used as templates for this. Jonah @Yurlungur promised to add a direct call to get Gibbs from the EOS's.

aematts:
2. I implemented a slightly modified mass fraction update. We need to discuss if this is what we want. It starts by moving mass from the highest Gibbs free energy phase to the lowest Gibbs free energy phase according to the rate and timestep. Next, if you still have mass in the highest Gibbs phase, you move mass to the next lowest phase, and so on. The you go to the next-to-the-highest Gibbs phase, and move what was there at the start of the time step and all that came from the highest Gibbs phase to lower Gibbs phases, starting with the lowest Gibbs free energy phase again. And so on.

aematts:
This allows you to move mass from phases with very high Gibbs energy to even the lowest Gibbs phase, the ground state, jumping several phases with Gibbs energy in between. This is a scenario that we have seen in some shock cases.
I am trying to figure out if there is something bad with this approach. I think it doesn't avoid slushing but it will be interesting to find out.

aematts:
@jhp-lanl @Yurlungur . Next we need to get together and 1) build a test case that goes the full circuit around, 2) teach me about Indexers, and 3) how to put this into classes and other parts of a C++ framework that I have neglected so far.

aematts:
I am updating my branch with the main and will commit it when I have checked that it runs the tests.

aematts:
Oh, I will start the documentation now as well. At least some of it as place holder. (Until we have straightened out the C++ structure).

Yurlungur:
@aematts thanks for all your hard work on this! I will have the Gibbs free energy ASAP. @jhp-lanl and I are going to read through what you have and let's find a time to meet sometime soon to discuss the C++ aspects.

singularity-eos/closure/kinetic_phasetransition_models.hpp
```diff
@@ -0,0 +1,91 @@
...
+      // always mass transport from j->k
+      Real hgj = gibbs[gibbsorder[j]];
+      Real hgk = gibbs[gibbsorder[k]];
+      // calc phase dx from hgi to hgk
+      Real dgdb = (hgj - hgk) / (b[gibbsorder[j] * num_phases + gibbsorder[k]]);
+      // First is highest level to levels below, largest gibbs energy diff first. Then
+      // follows 2nd highest to levels below.
+      logRjk[jk] =
+          log(w[gibbsorder[j] * num_phases + gibbsorder[k]] * dgdb) + dgdb * dgdb;
+      fromto[jk] = gibbsorder[j] * 10 + gibbsorder[k];
```
jhp-lanl, line 56: Is `10` supposed to be `num_phases` here?

If not, can you clarify for me where the factor of `10` comes from in the indexing?

singularity-eos/closure/kinetic_phasetransition_methods.hpp
```diff
@@ -0,0 +1,102 @@
...
+#endif // SINGULARITY_USE_KOKKOSKERNELS
+
+namespace singularity {
+
+PORTABLE_INLINE_FUNCTION void SmallStepMFUpdate(const Real logdt, const int num_phases,
+                                                const Real *massfractions,
+                                                const int *gibbsorder, const Real *logRjk,
+                                                Real *dmfs, Real *newmfs) {
+
+  Real minmassfraction = 1.e-10;
```
jhp-lanl, line 41: A couple of things:
1. Any compile time constants should be at least `const` or even better `constexpr`.
2. It appears that this constant is used in several files. I would propose moving this to `kinetic_phasetransition_util.hpp` and then define it as
```c++
constexpr Real KPT_MIN_MASS_FRACTION = 1.0e-10;
```
there. I'm using the convention that compile-time constants are all-caps.

This way we can define this parameter once and then use it in several files. If a host code needs this value for some reason, it's also exposed and easily queried. Further, if we want to expose it as a user-settable parameter in the future (maybe we don't... but who knows?), it's easier to do this if we've already extracted it from the places it is used.

singularity-eos/closure/kinetic_phasetransition_models.hpp
```diff
@@ -0,0 +1,91 @@
...
+  // to k+1 where j=0 is highest gibbs free energy state, and j=num_phases-1 is lowest
+  // gibbs free energy state. jk=0 is rate from highest (j=0) to lowest (k=num_phases-1)
+  // gibbs energy states.
+  return;
+}
+
+PORTABLE_INLINE_FUNCTION Real LogMaxTimeStep(const int num_phases, const Real *mfs,
+                                             const int *gibbsorder, const Real *logRjk) {
+
+  Real minmassfraction = 1.e-10;
```
jhp-lanl, line 70: See comment above about this parameter being located in a separate file and included rather than re-defined several times.

singularity-eos/closure/kinetic_phasetransition_models.hpp
```diff
@@ -0,0 +1,91 @@
...
+#else
+#include <Eigen/Dense>
+#endif // SINGULARITY_USE_KOKKOSKERNELS
+
+namespace singularity {
+
+PORTABLE_INLINE_FUNCTION void LogRatesCGModel(const Real *w, const Real *b,
+                                              const int num_phases, const Real *gibbs,
+                                              const int *gibbsorder, Real *logRjk,
+                                              int *fromto) {
```
jhp-lanl, line 39: I appreciate your attention to adding `const` where appropriate! Thank you!

test/CMakeLists.txt
```diff
@@ -110,6 +111,20 @@ if(SINGULARITY_BUILD_CLOSURE)
                                          singularity-eos::singularity-eos)
     target_include_directories(test_pte PRIVATE ${CMAKE_CURRENT_SOURCE_DIR})
     add_test(pte test_pte)
+#####
```
jhp-lanl, line 114: Extraneous comment?

jhp-lanl (changes requested):
Great job @aematts ! Thanks for getting this in.

I have several comments, but I think they're all fairly minor.

One general comment though: I didn't fully follow the indexing so I'm trusting that it's correct and properly tested. I did find one instance where I wasn't sure if it was correct, but otherwise I didn't really question it.

singularity-eos/closure/kinetic_phasetransition_methods.hpp
```diff
@@ -0,0 +1,102 @@
...
+      test = test + dmfs[offset - j];
+    }
+    for (int k = num_phases - 1; k > j; k--) {
+      // from low Gibb phase to high, k>j always, gibbs_j > gibbs_k always
+      // always mass transport from j->k
+      if (test < minmassfraction) {
+        dx = 0.;
+        dmfs[jk++] = dx;
+      } else {
+        dx = std::exp(logdt + logRjk[jk] + std::log(massfractions[gibbsorder[j]]));
```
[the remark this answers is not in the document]
Yurlungur: `std::exp(logdt + logRjk[jk])*massfractions` should be equilvalent right? Or is there some reason this might not work?

singularity-eos/closure/kinetic_phasetransition_methods.hpp
Yurlungur: I second all of this. Also, 1e-10 is perfectly reasonable, but I'm curious why you chose that value in particular.

singularity-eos/closure/kinetic_phasetransition_methods.hpp
```diff
@@ -0,0 +1,102 @@
...
+  // Then follows 2nd highest to levels below. logRjk[jk], with jk = (j+1)(num_phases-1) -
+  // (j-1)j/2 - k, contains rate for phase j+1 to k+1 where j=0 is highest gibbs free
+  // energy state, and j=num_phases-1 is lowest gibbs free energy state. jk=0 is rate from
+  // highest (j=0) to lowest (k=num_phases-1) gibbs energy states. dmfs[jk] is mass
+  // transport from phase j+1 to k+1, in the same way. Fill lowest level first and stop
+  // when origin level is exhausted. This will allow for jumping phases if needed. I also
+  // think it is most physical.
+  int jk = 0;
+  Real dx = 0.;
+  Real test = 0.;
```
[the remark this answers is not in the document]
Yurlungur: We could use a signaling nan
```
Real test = std::numeric_limits<Real>::signaling_NaN();
```

singularity-eos/closure/kinetic_phasetransition_methods.hpp
```diff
@@ -0,0 +1,102 @@
...
+  Real minmassfraction = 1.e-10;
+
+  // In logRjk: First is highest level to levels below, largest gibbs energy diff first.
+  // Then follows 2nd highest to levels below. logRjk[jk], with jk = (j+1)(num_phases-1) -
+  // (j-1)j/2 - k, contains rate for phase j+1 to k+1 where j=0 is highest gibbs free
+  // energy state, and j=num_phases-1 is lowest gibbs free energy state. jk=0 is rate from
+  // highest (j=0) to lowest (k=num_phases-1) gibbs energy states. dmfs[jk] is mass
+  // transport from phase j+1 to k+1, in the same way. Fill lowest level first and stop
+  // when origin level is exhausted. This will allow for jumping phases if needed. I also
+  // think it is most physical.
```
Yurlungur, line 50: Is this ordering enforced or assumed? I don't fully understand the looping logic below.

singularity-eos/closure/kinetic_phasetransition_utils.hpp
```diff
@@ -0,0 +1,59 @@
...
+      if (gibbs[gibbsorder[k]] > gibbs[gibbsorder[j]]) {
+        itmp = gibbsorder[j];
+        gibbsorder[j] = gibbsorder[k];
+        gibbsorder[k] = itmp;
+      }
+    }
+  }
+  return;
+}
+
```
Yurlungur, line 56: If I'm understanding this, `gibbs` is left alone but the ordering is specified by `gibbsorder` is that right? With essentially an insertion sort?  I think this is fine for now, but it might be worth us discussing this in more detail at some point.

In general regarding the sorting routines, I'd like to chat with you briefly so I fully understand the expectations for in and out for these various methods. And what exactly we expect as far as the number of phases. If 5 is normal, then insertion sort is fine... if we get up to, e.g., 20, we might want to think about a more performant sorting algorithm. Unfortunately, pre-packaged sorting algorithms won't be consistent on GPU so we might have to roll our own. Fortunately, I don't think that would be too hard.

I'm just dumping my thoughts here. You don't need to change anything about your sorter right now.

test/CMakeLists.txt
```diff
@@ -110,6 +111,20 @@ if(SINGULARITY_BUILD_CLOSURE)
...
+    target_include_directories(test_pte_2phase PRIVATE ${CMAKE_CURRENT_SOURCE_DIR})
+    add_test(pte_2phase test_pte_2phase)
+
+    if(SINGULARITY_USE_EOSPAC)
+      add_executable(test_pte_3phase test_pte_3phase.cpp)
+      target_link_libraries(test_pte_3phase PRIVATE Catch2::Catch2
+                                           singularity-eos::singularity-eos)
+      target_include_directories(test_pte_3phase PRIVATE ${CMAKE_CURRENT_SOURCE_DIR})
+      add_test(pte_3phase test_pte_3phase)
+    endif()
```
Yurlungur, line 127: :+1:

test/pte_longtest_2phaseVinetSn.hpp
```diff
@@ -0,0 +1,117 @@
...
+
+#include <stdlib.h>
+
+#include <ports-of-call/portability.hpp>
+#include <ports-of-call/portable_arrays.hpp>
+#include <singularity-eos/eos/eos.hpp>
+
+// This data is taken from the .ult file for a test case in Flag, MPMat_KPT_PTsolv.ult.std
+// Note that at this point in time (April 2024) the global and phase times are not alinged
+// in .ult but they will be in the near future.
```
Yurlungur, line 26: :+1:

test/pte_longtest_2phaseVinetSn.hpp
Yurlungur: Thanks for adding said comment. :)

test/test_kpt_models.cpp
```diff
@@ -0,0 +1,275 @@
...
+          1.00000000000000007e+210, 1.00000000000000007e+210, 1.00000000000000007e+210,
+          1.00000000000000007e+210, 4.99943400447804946e+05,  4.35424707359967171e+05,
+          1.07530324485867186e+04,  3.93959978909504514e+02,  3.09367756860215100e+05,
+          2.23469012616621285e+03};
+      constexpr std::array<int, num> phaseorder{0, 1, 2, 3, 4};
+      constexpr std::array<int, num> order_true{3, 4, 0, 2, 1};
+
+      WHEN("Using SortGibbs(num,gibbs,order) to give phases in order of largest to "
+           "smallest gibbs") {
+        SortGibbs(num, v_gibbs, v_order);
```
Yurlungur, line 95: Once we do the GPU stuff, this test may need to change in small ways---i.e., what things are on host, what things are on device, etc. We can discuss in person.

Yurlungur (commented):
Overall this looks great, @aematts ! I like the code organization you set up and the bite-sized pieces you identified. I second @jhp-lanl 's suggestions. Overall, I think it will be quite straightforward to add the C++/indexer/GPU bits.

singularity-eos/closure/kinetic_phasetransition_methods.hpp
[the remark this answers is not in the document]
jhp-lanl: That's what I'd think, but maybe I'm reading it wrong.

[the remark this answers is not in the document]
jhp-lanl: We discussed this today in-person and the issue is that the sum can be well-behaved while the individual portions are not. Thus this is the most numerically stable approach.

singularity-eos/closure/kinetic_phasetransition_methods.hpp
```diff
@@ -0,0 +1,102 @@
...
+  int offset = 0;
+  for (int j = 0; j < num_phases - 1; j++) {
+    // from high Gibb phase to low
+    test = massfractions[gibbsorder[j]];
+    // all phases except the higest gibbs free energy one (j=0) can get contributions from
+    // phases with higher gibbs free energy.
+    for (int k = 0; k < j; k++) {
+      offset = (k + 1) * (num_phases - 1) - (k - 1) * k / 2;
+      test = test + dmfs[offset - j];
+    }
```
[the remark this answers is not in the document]
jhp-lanl: My confusion was in exactly what `test` was. I'll make a new comment.

singularity-eos/closure/kinetic_phasetransition_methods.hpp
```diff
@@ -0,0 +1,102 @@
...
+    for (int k = num_phases - 1; k > j; k--) {
+      // from low Gibb phase to high, k>j always, gibbs_j > gibbs_k always
+      // always mass transport from j->k
+      if (test < minmassfraction) {
+        dx = 0.;
+        dmfs[jk++] = dx;
```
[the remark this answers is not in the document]
jhp-lanl: Maybe a better comment here would be
```c++
// Phase has already been depleted
```

singularity-eos/closure/kinetic_phasetransition_methods.hpp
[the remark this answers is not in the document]
jhp-lanl: More broadly, we agreed to limit the scope of these variables to where they're used. I'll make new comments to that effect at the specific places to declare them.

singularity-eos/closure/kinetic_phasetransition_models.hpp
jhp-lanl: We established that this is a label and not an index.

Can you add a comment here that the factor of 10 is just for creating a nice label?

singularity-eos/closure/kinetic_phasetransition_utils.hpp
jhp-lanl: I think we resolved this in-person. We can explore different sorting algorithms in the future.

singularity-eos/closure/kinetic_phasetransition_methods.hpp
```diff
@@ -0,0 +1,102 @@
...
+      } else {
+        dx = std::exp(logdt + logRjk[jk] + std::log(massfractions[gibbsorder[j]]));
+        newtest = test - dx;
```
jhp-lanl, line 73: Can you declare `newtest` here in addition to defining it? You can remove the declaration above. This will make it easier to see that this is a variable with limited scope only relevant to the surrounding code.

I'd also suggest that it have a different name,

```suggestion
        const Real depleted_mass_frac = test - dx;
```
to reflect that this is the mass fraction with depletion.

singularity-eos/closure/kinetic_phasetransition_methods.hpp
jhp-lanl, line 53: I would propose we rename `test` to `new_mass_frac`

jhp-lanl (changes requested):
I resolved some comments from our discussion earlier and am adding more concise changes.

aematts:
I added some documentation. I will fine tune it when I have gone trough the code and made the changes you suggests. Hopefully I can work on it tomorrow so that we can merge Thursday.

singularity-eos/closure/kinetic_phasetransition_methods.hpp
aematts: Done

singularity-eos/closure/kinetic_phasetransition_methods.hpp
aematts: OK. It is now in the documentation. Not as clearly written as I would like it to be but with a figure that I think explains a lot.

singularity-eos/closure/kinetic_phasetransition_methods.hpp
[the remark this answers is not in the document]
aematts: Done

singularity-eos/closure/kinetic_phasetransition_methods.hpp
aematts: Called in new_massfrac

aematts: Done

singularity-eos/closure/kinetic_phasetransition_methods.hpp
aematts: I called in new_trial_massfrac and put in comments Jonah suggested above.
Done

singularity-eos/CMakeLists.txt
```diff
@@ -63,6 +63,9 @@ register_headers(
     eos/singularity_eos_init_utils.hpp
     eos/variant_utils.hpp
     eos/eos_carnahan_starling.hpp
+    closure/kinetic_phasetransition_models.hpp
+    closure/kinetic_phasetransition_utils.hpp
+    closure/kinetic_phasetransition_methods.hpp
```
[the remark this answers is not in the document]
aematts: Done

singularity-eos/closure/kinetic_phasetransition_models.hpp
aematts: Fixed

singularity-eos/closure/kinetic_phasetransition_models.hpp
aematts: Done

test/CMakeLists.txt
aematts: Oops, how did that get into the file? Took it away.

aematts:
@jhp-lanl and @Yurlungur : I have addressed your comments. All tests on GitHub passed but some failed on re-git.lanl.gov, mainly Rocinante but also one gpu test and the other gpu is still running, like 1:36 hours later...

Can you let me know if I can go ahead and take away the WIP and prepare for merge...

aematts:
Rocinante has some kind of connection error resulting in lines like this:
fatal: unable to access 'https://gitlab.com/libeigen/eigen.git/': Failed to connect to gitlab.com port 443 after 15530 ms: Couldn't connect to server

aematts:
The two gpu cases do not like my Kokkos stuff in test/test_kpt_models.cpp and also complains about my writing output (std::cout).

singularity-eos/closure/kinetic_phasetransition_methods.hpp
```diff
@@ -0,0 +1,99 @@
...
+    }
+    for (int k = num_phases - 1; k > j; k--) {
+      // from low Gibb phase to high, k>j always, gibbs_j > gibbs_k always
+      // always mass transport from j->k
+      if (new_massfrac < KPT_MIN_MASS_FRACTION) {
+        // Phase depleated, nothing more to transfer
+        Real dx = 0.;
+        dmfs[jk++] = dx;
+      } else {
+        Real dx = std::exp(logdt + logRjk[jk] + std::log(massfractions[gibbsorder[j]]));
```
Yurlungur, line 68: Sorry the last minute suggestion---might be worth adding a comment above this line mentioning that this is done in log to avoid overflow. To prevent future doofuses (me) from switching to doing this in linear.

singularity-eos/closure/kinetic_phasetransition_models.hpp
Yurlungur: Can we also add a TODO comment that in the future this label should maybe changed to a struct or another more generalizable form? (No need to do this now. But would be good if we knew where to look if this comes up later.)

Yurlungur (approved):
Looks good @aematts ! A few  minor nitpicks and I think this is good to go. The documentation is much appreciated!

Yurlungur:
> All tests on GitHub passed but some failed on re-git.lanl.gov, mainly Rocinante but also one gpu test and the other gpu is still running, like 1:36 hours later...
> 
> Can you let me know if I can go ahead and take away the WIP and prepare for merge...

re-git tests currently being refactored I don't think we expect them all to pass. Don't worry about it for now. @rbberger is fixing the CI as we speak. :)

> Can you let me know if I can go ahead and take away the WIP and prepare for merge...

Remove the WIP :)

> The two gpu cases do not like my Kokkos stuff in test/test_kpt_models.cpp and also complains about my writing output (std::cout).

Wrap that whole test in `#ifdef PORTABILITY_STRATEGY_NONE` for now. And add a TODO that the GPU capability will need to be fixed/cleaned up in a subsequent MR. We can do that after we're happy with this and it's merged. Also will be helpful to have the re-git CI working for that.

doc/sphinx/src/using-kpt.rst
```diff
@@ -0,0 +1,248 @@
...
+The ``gibbs`` and ``gibbsorder`` arrays are of length :math:`N` while ``w`` and ``b`` are arrays of length :math:`N^2` representing the :math:`N \times N` matrices, row by row. 
+Note that it is *NOT* assumed that the phase transition parameters are the same when going from :math:`i \rightarrow j` and  :math:`j \rightarrow i`, that is, :math:`\{\nu,B\}_{ij} \neq \{\nu,B\}_{ji}`.
+Also note that :math:`R_{ii} = 0`.
+``logRij`` is an array of length :math:`N (N-1)/2` giving the logarithm of the non-zero mass transportation rates between phases. The ``fromto`` array 
+gives to which two phases in the ``gibbs`` array, each rate in ``logRij`` is associated: ``logRij[k]`` is the logarithm of the mass 
+transportation rate from/to phases ``fromto[k]``, with ``k`` a phase transition index according to the figure below. The integer
+in ``fromto``, "ij", is composed from the ``gibbs`` index of the "from" phase, :math:`i`, and the ``gibbs`` index of the "to" phase, 
+:math:`j`, as :math:`i*10+j`, and with a single digit integer, "x", interpreted as "0x".
+
+.. image:: ../GibbsOrder.pdf
```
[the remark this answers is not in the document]
aematts: I do not have that problem in my Safari browser but the same problem happened for Jeff some time ago, but it turns out it is a browser thing. I do not remember what Jeff decided we should do about it.

doc/sphinx/src/using-kpt.rst
```diff
@@ -0,0 +1,248 @@
...
+The ``gibbs`` and ``gibbsorder`` arrays are of length :math:`N` while ``w`` and ``b`` are arrays of length :math:`N^2` representing the :math:`N \times N` matrices, row by row. 
+Note that it is *NOT* assumed that the phase transition parameters are the same when going from :math:`i \rightarrow j` and  :math:`j \rightarrow i`, that is, :math:`\{\nu,B\}_{ij} \neq \{\nu,B\}_{ji}`.
+Also note that :math:`R_{ii} = 0`.
+``logRij`` is an array of length :math:`N (N-1)/2` giving the logarithm of the non-zero mass transportation rates between phases. The ``fromto`` array 
+gives to which two phases in the ``gibbs`` array, each rate in ``logRij`` is associated: ``logRij[k]`` is the logarithm of the mass 
+transportation rate from/to phases ``fromto[k]``, with ``k`` a phase transition index according to the figure below. The integer
+in ``fromto``, "ij", is composed from the ``gibbs`` index of the "from" phase, :math:`i`, and the ``gibbs`` index of the "to" phase, 
+:math:`j`, as :math:`i*10+j`, and with a single digit integer, "x", interpreted as "0x".
```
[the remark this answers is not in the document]
aematts: As I said, I will have to work more on it eventually but maybe it can wait for a while?

doc/sphinx/src/using-kpt.rst
```diff
@@ -0,0 +1,248 @@
...
+                  const int *gibbsorder, Real *logRij, int *fromto)
+
+where ``w`` is :math:`\nu_{ij}`, ``b`` is :math:`B_{ij}`, ``num_phases`` is
+:math:`N`, the number of phases, ``gibbs`` is :math:`G_i`, ``gibbsorder`` is an array of length :math:`N`, where the ``gibbs`` phase indices are ordered 
+from the largest Gibbs free energy phase to the lowest Gibbs free energy phase (see figure below),
+``logRij`` is :math:`\log(R_{ij})`, and ``fromto`` is a map between the phase indices :math:`i` and :math:`j` and the :math:`ij` phase transition indices in :math:`R_{ij}`. 
+The ``gibbs`` and ``gibbsorder`` arrays are of length :math:`N` while ``w`` and ``b`` are arrays of length :math:`N^2` representing the :math:`N \times N` matrices, row by row. 
+Note that it is *NOT* assumed that the phase transition parameters are the same when going from :math:`i \rightarrow j` and  :math:`j \rightarrow i`, that is, :math:`\{\nu,B\}_{ij} \neq \{\nu,B\}_{ji}`.
+Also note that :math:`R_{ii} = 0`.
+``logRij`` is an array of length :math:`N (N-1)/2` giving the logarithm of the non-zero mass transportation rates between phases. The ``fromto`` array
```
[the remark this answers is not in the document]
aematts: I decided not to, since I have not really introduced any matrices. But it is easy to see that if $R_{ij}$ is non-zero, $R_{ji}$ would be zero (because of the Heaviside step function, oh, I forgot to say that that is what it is....). Maybe I can say that. With $R_{ii}=0$ this means a triangle of size N*(N-1)/2. But there are many way to come to the same number...

singularity-eos/closure/kinetic_phasetransition_methods.hpp
aematts: OK. Will do.

doc/sphinx/src/using-kpt.rst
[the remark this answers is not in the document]
aematts: Done.

doc/sphinx/src/using-kpt.rst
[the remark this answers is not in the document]
Yurlungur: I'll try safari and see if that helps. If it's just a browser thing it's probably fine. When i click the link it still renders... it just doesn't render inline, weirdly.

doc/sphinx/src/using-kpt.rst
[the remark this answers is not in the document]
Yurlungur: Sure---it's fine for now.

jhp-lanl (approved):
I think you addressed all my comments. Approved.

aematts:
> > All tests on GitHub passed but some failed on re-git.lanl.gov, mainly Rocinante but also one gpu test and the other gpu is still running, like 1:36 hours later...
> > Can you let me know if I can go ahead and take away the WIP and prepare for merge...
> 
> re-git tests currently being refactored I don't think we expect them all to pass. Don't worry about it for now. @rbberger is fixing the CI as we speak. :)
> 
> > Can you let me know if I can go ahead and take away the WIP and prepare for merge...
> 
> Remove the WIP :)
> 
> > The two gpu cases do not like my Kokkos stuff in test/test_kpt_models.cpp and also complains about my writing output (std::cout).
> 
> Wrap that whole test in `#ifdef PORTABILITY_STRATEGY_NONE` for now. And add a TODO that the GPU capability will need to be fixed/cleaned up in a subsequent MR. We can do that after we're happy with this and it's merged. Also will be helpful to have the re-git CI working for that.

@Yurlungur : If I wrap it in `#ifdef PORTABILITY_STRATEGY_NONE` it doesn't run at all for me. Even on skylake-gold. What shall I do?

Yurlungur:
> > > All tests on GitHub passed but some failed on re-git.lanl.gov, mainly Rocinante but also one gpu test and the other gpu is still running, like 1:36 hours later...
> > > Can you let me know if I can go ahead and take away the WIP and prepare for merge...
> > 
> > 
> > re-git tests currently being refactored I don't think we expect them all to pass. Don't worry about it for now. @rbberger is fixing the CI as we speak. :)
> > > Can you let me know if I can go ahead and take away the WIP and prepare for merge...
> > 
> > 
> > Remove the WIP :)
> > > The two gpu cases do not like my Kokkos stuff in test/test_kpt_models.cpp and also complains about my writing output (std::cout).
> > 
> > 
> > Wrap that whole test in `#ifdef PORTABILITY_STRATEGY_NONE` for now. And add a TODO that the GPU capability will need to be fixed/cleaned up in a subsequent MR. We can do that after we're happy with this and it's merged. Also will be helpful to have the re-git CI working for that.
> 
> @Yurlungur : If I wrap it in `#ifdef PORTABILITY_STRATEGY_NONE` it doesn't run at all for me. Even on skylake-gold. What shall I do?

weird... try `#ifndef PORTABILITY_STRATEGY_KOKKOS` at the top and `#endif` at the bottom

aematts:
> > > > All tests on GitHub passed but some failed on re-git.lanl.gov, mainly Rocinante but also one gpu test and the other gpu is still running, like 1:36 hours later...
> > > > Can you let me know if I can go ahead and take away the WIP and prepare for merge...
> > > 
> > > 
> > > re-git tests currently being refactored I don't think we expect them all to pass. Don't worry about it for now. @rbberger is fixing the CI as we speak. :)
> > > > Can you let me know if I can go ahead and take away the WIP and prepare for merge...
> > > 
> > > 
> > > Remove the WIP :)
> > > > The two gpu cases do not like my Kokkos stuff in test/test_kpt_models.cpp and also complains about my writing output (std::cout).
> > > 
> > > 
> > > Wrap that whole test in `#ifdef PORTABILITY_STRATEGY_NONE` for now. And add a TODO that the GPU capability will need to be fixed/cleaned up in a subsequent MR. We can do that after we're happy with this and it's merged. Also will be helpful to have the re-git CI working for that.
> > 
> > 
> > @Yurlungur : If I wrap it in `#ifdef PORTABILITY_STRATEGY_NONE` it doesn't run at all for me. Even on skylake-gold. What shall I do?
> 
> weird... try `#ifndef PORTABILITY_STRATEGY_KOKKOS` at the top and `#endif` at the bottom

Then it will still run on the GPUs and complain about my writing with std::cout

Yurlungur:
Something is off here `PORTABILITY_STRATEGY_NONE` should just get automatically set for you by `cmake` for CPU runs and `PORTABILITY_STRATEGY_KOKKOS` should automatically get set for you by `cmake` for GPU runs. I'm not sure why that's not happening. 

You might need to include `<ports-of-call/portability.hpp>` at the top of the test?

aematts:
> Something is off here `PORTABILITY_STRATEGY_NONE` should just get automatically set for you by `cmake` for CPU runs and `PORTABILITY_STRATEGY_KOKKOS` should automatically get set for you by `cmake` for GPU runs. I'm not sure why that's not happening.
> 
> You might need to include `<ports-of-call/portability.hpp>` at the top of the test?

I did not do cmake .....
Sorry. I understand now, and I have pushed my changes to make sure they run. Soon ready.

aematts:
@Yurlungur I do not see any of the tests in this MR in the automatic testing. Now, I have not check it before so I am not sure they have ever been included in this pipeline. All the EOS tests I have done are in there. I use Spiner for the PTE solver test cases, but I think the KPT tests should run, but they do not.

aematts:
the re-git pipeline include the PTE solver tests but not the KPT tests:
 Start 53: Scenario: Density-Temperature PTE Solver
53/67 Test #53: Scenario: Density-Temperature PTE Solver ...........................................   Passed    0.47 sec
      Start 54: pte
54/67 Test #54: pte ................................................................................   Passed    0.04 sec
      Start 55: pte_2phase
55/67 Test #55: pte_2phase .........................................................................   Passed    0.03 sec
      Start 56: pte_3phase
56/67 Test #56: pte_3phase .........................................................................   Passed    1.3

aematts:
Hm, Jeff wanted me to move the KPT tests down to the closure section in CMakeLists, that is probably why.

Please help me figure this out. Where should we test what.

I also use EOSPAC for one pte test and that doesn't show up on GitHub either.

aematts:
Same result as before on re-git. But my kpt tests are not in there. But the GPU's choke on other things anyway, some of my pte tests but also eos_gruneisen and eos_variant. So it is not only my fault.

Yurlungur:
> Same result as before on re-git. But my kpt tests are not in there. But the GPU's choke on other things anyway, some of my pte tests but also eos_gruneisen and eos_variant. So it is not only my fault.

The automated testing is currently broken on re-git on GPUs. @rbberger is fixing it.

We should be able to manually test on CPU and GPU on Darwin though. The PTE tests only run on the yellow, not on github because we don't have access to the sesame database on the open. Let me give it a try on your branch.

aematts:
I am testing it with 
 cmake .. --preset="basic_with_testing" -DSINGULARITY_SUBMODULE_MODE=ON -DSINGULARITY_BUILD_CLOSURE=ON

on skylake-gold just now. On Darwin.

I am not sure this includes SPINER but anyway.

Yurlungur:
> cmake .. --preset="basic_with_testing" -DSINGULARITY_SUBMODULE_MODE=ON -DSINGULARITY_BUILD_CLOSURE=ON

:+1: I'll try with CPU and GPU and see if I can figure out what's going on with running/not running your tests. stay tuned.

aematts:
> I am testing it with cmake .. --preset="basic_with_testing" -DSINGULARITY_SUBMODULE_MODE=ON -DSINGULARITY_BUILD_CLOSURE=ON
> 
> on skylake-gold just now. On Darwin.
> 
> I am not sure this includes SPINER but anyway.

No, my KPT tests did not run :(.

Will continue looking at it.

aematts:
The current version now runs on GitHub and on the CPU part of re-git. And it doesn't run on the GPU part. 
I think that is enough for now. 

re-git GPU is complaining about std::cout stuff in my 2phase and 3phase pte tests. Should I put in the #ifndef PORTABILITY_STRATEGY_KOKKOS there too? Or can that wait for now (there are complaints about other tests as well so it will still not pass). 

If so I am done and you can merge.

Yurlungur:
> When I comment out #ifdef PORTABILITY_STRATEGY_NONE it works: Start 54: Scenario: First log rate test 54/68 Test #54: Scenario: First log rate test ...................................................... Passed 0.01 sec Start 55: pte 55/68 Test #55: pte ................................................................................ Passed 0.01 sec Start 56: pte_2phase 56/68 Test #56: pte_2phase ......................................................................... Passed 0.01 sec Start 57: pte_3phase 57/68 Test #57: pte_3phase ......................................................................... Passed 0.25 sec

@aematts I found the problem. `PORTABILITY_STRATEGY_NONE` is only for builds without Kokkos. But there are CPU builds with Kokkos. That means my original idea for a solution isn't so simple as I thought. That's on me. Let me see if I can get your code building on GPUs. Then I'd like @rbberger 's CI MR to go through (it's up now and should be merged soon.) Then I'll click merge on yours. Thanks for all the hard work on this!

Yurlungur:
There were a few minor issues, now resolved:
1. Some minor things with GPU builds and, e.g., cout vs printf.
2. `EOSPAC` doesn't currently build in a way that it can be called on device with the closure tooling so I disabled the 3phase test for device builds. The test compiles but doesn't do anything.
3. GPU builds were very slow because the variant was being rebuilt in each test. I know choose m minimal variant. Compile times are much shorter.

Yurlungur:
Will now merge as soon as CI passes on re-git and github.

Yurlungur:
@aematts I've hit a small snag on the CI. The material IDs used for the 5phase sesame test are in a sesame data base curated by you. This doesn't look like it's available on all CI machines, such as Rocinante. I'm looping in @rbberger . Let's continue the conversation by email.

Yurlungur:
Tests now passing on re-git. Will merge as soon as tests pass on github.

jhp-lanl:
Thanks for shepherding this to completion @Yurlungur !

PR #370, opened 2024-05-01
Fix bulk modulus calculation in spiner EOS

pdmullen:
Fixes a (potential) bug in Spiner EOS bulk modulus calculation.

## PR Summary

Despite the bulk modulus being a required table in a SpinerDependsRhoT or SpinerDependsRhoSie EOS file, it is overwritten by functions `fixBulkModulus_` and `calcBMod_`, respectively.  Inside these functions, the bulk modulus formulae assume `dpdrho` to be $\partial P / \partial \rho \\; \vert_{T}$.  However, I believe (please check me) that spiner expects that the tables provided correspond to $\partial P / \partial \rho \\; \vert_{e}$.  This PR changes the bulk modulus formula to reflect this.  There is this trick that `eos_spiner` plays wherein if $\partial P / \partial e \\; \vert_{\rho} < 0$, it modifies the bulk modulus calculation.  I tried to preserve this behavior by defining $\partial P / \partial \rho \\; \vert_{T}$ in terms of thermodynamic derivatives (please double check me...).

## PR Checklist

### DIFF
--- a/singularity-eos/eos/eos_spiner.hpp
+++ b/singularity-eos/eos/eos_spiner.hpp
@@ -873,12 +873,12 @@ inline void SpinerEOSDependsRhoT::fixBulkModulus_() {
       Real DPDR = dPdRho_.interpToReal(lRho, lT);
       Real DPDE = dPdE_.interpToReal(lRho, lT);
       Real DEDR = dEdRho_.interpToReal(lRho, lT);
-      Real DTDE = dTdE_.interpToReal(lRho, lT);
+      Real DPDR_T = DPDR + DPDE * DEDR;
       Real bMod;
       if (DPDE > 0.0 && rho > 0.0) {
-        bMod = rho * DPDR + DPDE * (press / rho - rho * DEDR);
+        bMod = rho * DPDR + DPDE * (press / rho);
       } else if (rho > 0.0) {
-        bMod = std::max(rho * DPDR, 0.0);
+        bMod = std::max(rho * DPDR_T, 0.0);
       } else {
         bMod = 0.0;
       }
@@ -1640,12 +1640,12 @@ inline void SpinerEOSDependsRhoSie::calcBMod_(SP5Tables &tables) {
       Real DPDR = tables.dPdRho(j, i);
       Real DPDE = tables.dPdE(j, i);
       Real DEDR = tables.dEdRho(j, i);
-      Real DTDE = tables.dTdE(j, i);
+      Real DPDR_T = DPDR + DPDE * DEDR;
       Real bMod;
       if (DPDE > 0.0 && rho > 0.0) {
-        bMod = rho * DPDR + DPDE * (press / rho - rho * DEDR);
+        bMod = rho * DPDR + DPDE * (press / rho);
       } else if (rho > 0.0) {
-        bMod = std::max(rho * DPDR, 0.0);
+        bMod = std::max(rho * DPDR_T, 0.0);
       } else {
         bMod = 0.0;
       }
--- a/singularity-eos/eos/eos_spiner.hpp
+++ b/singularity-eos/eos/eos_spiner.hpp
@@ -875,7 +875,7 @@ inline void SpinerEOSDependsRhoT::fixBulkModulus_() {
       Real DEDR_T = dEdRho_.interpToReal(lRho, lT);
       Real DPDR_T = DPDR_E + DPDE_R * DEDR_T;
       Real bMod;
-      if (DPDE > 0.0 && rho > 0.0) {
+      if (DPDE_R > 0.0 && rho > 0.0) {
         bMod = rho * DPDR_E + DPDE_R * (press / rho);
       } else if (rho > 0.0) {
         bMod = std::max(rho * DPDR_T, 0.0);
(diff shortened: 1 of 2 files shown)


Yurlungur:
Thanks for catching this @pdmullen ! I'm going to trigger tests on Darwin.

Yurlungur (approved):
I want to hear from @jhp-lanl @dholladay00 and/or @chadmeyer just to confirm. But I think this fix is correct. And tests pass on Darwin.

chadmeyer:
Looking at this now.  I think this might be OK, but I want to make sure (and perhaps also ensure that this is what we want).  

Let's agree that $`c^2 = \left.\frac{\partial P}{\partial \rho}\right|_{e}+\frac{P}{\rho^2}\left.\frac{\partial P}{\partial e}\right|_{\rho}`$.  Is that what is being done in both of these cases?  I ask because one of these cases is `DependsRhoT` and the other one is `DependsRhoSie`, so I'm not sure what was tabulated (is it the same between them?). 

Is that what it is?  I'm thinking so, but I might need to dig into `Sesame2Spiner` to be sure.

Yurlungur:
> Is that what it is? I'm thinking so, but I might need to dig into `Sesame2Spiner` to be sure.

I had the same confusion. I thought `Sesame2Spiner` was tabulating `$dP/d\rho |_T$ but it turns out it's tabulating $dP/d\rho |_e$.

singularity-eos/eos/eos_spiner.hpp
```diff
@@ -873,12 +873,12 @@ inline void SpinerEOSDependsRhoT::fixBulkModulus_() {
       Real DPDR = dPdRho_.interpToReal(lRho, lT);
       Real DPDE = dPdE_.interpToReal(lRho, lT);
       Real DEDR = dEdRho_.interpToReal(lRho, lT);
-      Real DTDE = dTdE_.interpToReal(lRho, lT);
+      Real DPDR_T = DPDR + DPDE * DEDR;
```
jhp-lanl, line 876: while we're here would it make sense to explicitly change the variable names to make sure they include what should be held constant? In other words I'd advocate for `DPDR` being changed to `DPDR_E`.

Otherwise it's difficult to tell what exactly is being done here.

jhp-lanl, line 876: Provided that 
- `DPDR` is $`\left(\frac{\partial P}{\partial \rho}\right)_e`$
- `DEDR` is $`\left(\frac{\partial e}{\partial \rho}\right)_T`$
- `DPDE` is $`\left(\frac{\partial P}{\partial e}\right)_\rho`$
Then yes I think the math is correct

singularity-eos/eos/eos_spiner.hpp
```diff
@@ -873,12 +873,12 @@ inline void SpinerEOSDependsRhoT::fixBulkModulus_() {
...
       Real bMod;
       if (DPDE > 0.0 && rho > 0.0) {
-        bMod = rho * DPDR + DPDE * (press / rho - rho * DEDR);
+        bMod = rho * DPDR + DPDE * (press / rho);
```
jhp-lanl, line 879: This is bizarre... good  catch!

jhp-lanl (approved):

jhp-lanl:
> > Is that what it is? I'm thinking so, but I might need to dig into `Sesame2Spiner` to be sure.
> 
> I had the same confusion. I thought `Sesame2Spiner` was tabulating dP/dρ|T but it turns out it's tabulating dP/dρ|e.

I think the another issue is that `Sesame2Spiner` gets density-energy derivatives for density-temperature tables via thermodynamic identities and then uses another thermodynamic identity on them to get the bulk modulus. I would propose (for the future) to just ask for the bulk modulus directly via the `EOS_BSt_DT` table.

This would be advantageous for two reasons: 1) it avoids roundoff due to unnecessary multiplications and additions and 2) it can leverages the effort that the EOSPAC folks put into making sure that EOSPAC returns the most accurate bulk modulus possible.

singularity-eos/eos/eos_spiner.hpp
pdmullen: thanks for double checking 😸

pdmullen:
> > > Is that what it is? I'm thinking so, but I might need to dig into `Sesame2Spiner` to be sure.
> > 
> > 
> > I had the same confusion. I thought `Sesame2Spiner` was tabulating dP/dρ|T but it turns out it's tabulating dP/dρ|e.
> 
> I think the another issue is that `Sesame2Spiner` gets density-energy derivatives for density-temperature tables via thermodynamic identities and then uses another thermodynamic identity on them to get the bulk modulus. I would propose (for the future) to just ask for the bulk modulus directly via the `EOS_BSt_DT` table.
> 
> This would be advantageous for two reasons: 1) it avoids roundoff due to unnecessary multiplications and additions and 2) it can leverages the effort that the EOSPAC folks put into making sure that EOSPAC returns the most accurate bulk modulus possible.

I'd add that it feels a bit weird for SpinerEOS to demand a bulk modulus field in tables when it actually just recomputes them from thermodynamic derivatives.  Maybe a future MR could optionally enroll "bulk modulus correction/fixing".  That way if a user wants to use bulk moduli as directly computed from EOSPAC, they could set `correct_bulk_mod = false` (or similar).

jhp-lanl:
I think this looks ready for merging

Yurlungur:
@jhp-lanl @pdmullen  agreed regarding using the bulk modulus from eospac. I'll make an issue.

Issue #260, opened 2023-05-22
Add feature to automatically set reference energy for Davis Products EOS


jhp-lanl:
The Davis Products EOS is most often used in combination with a reactants EOS to model reactive flows. For a detonation wave to be self-propagating, the Rayleigh line extending from the initial conditions of the reactants needs to be tangent to the release isentrope of the products at the CJ point. Most often, we want this CJ isentrope to the reference isentrope for the Davis EOS, so it makes sense to be able to automatically set the reference energy of the products EOS automatically in order to achieve this tangency condition.

This will require an iterative procedure at setup that will could look something like this:
1. Find bounding points on the reference isentrope where the slope of the Rayleigh line is larger than and smaller than the slope of the isentrope.
2. Use a bounded root finding method to iteratively find the point where the slope of the Rayleigh line equals the slope of the reference isentrope
3. Calculate the energy of the CJ point given the Rayleigh line and then subtract the energy of the reference isentrope at the CJ point to calculate the appropriate energy offset.

This procedure _should_ be done with an arbitrary reactants reference energy.

jhp-lanl:
This procedure requires two additional parameters: the reference pressure and reference density/volume for the reacting material. When these parameters are given, these would obviate the need for specifying `E0`.

Open question: should these parameters _replace_ E0 (i.e. overload the constructor so that there is _one_ more input) or should all three (`E0`, `P0`, and `rho0`) be optional parameters?

The former places the burden on the host code to determine what the user is trying to specify while the latter pushes that logic into singularity. I'm leaning towards overloading since I think that would make it more clear in the host code what singularity is doing under the hood.

Yurlungur:
IMO different host codes may want to do things differently, so I am in favor of the former---the overload approach.

jhp-lanl:
@Yurlungur brought up

> 👍 I'm in favor of the helper being `static` too.

A static member function would in addition to `P0` and `rho0` need to take `vc`, `k`, `n`, and `a` (i.e. all but two parameters for the EOS, excepting `E0`).

I suppose if this is all just part of the constructor though it shouldn't be an issue. Originally I was thinking maybe it shouldn't be static, but now that I think about it, I think I agree with you.

jhp-lanl:
Now that the Davis Product EOS no longer takes E0, it seems prudent for this to _NOT_ be static. That way the energy shift modifier can be used with something like
```c++
DavisProducts eos = DavisProducts(...);
[...]

PR #372, opened 2024-05-10
Remove E0 from Davis Products EOS

jhp-lanl:
## PR Summary

We were taking the `E0` parameter in the Davis Products EOS but not actually using it anywhere.

This MR simply subtracts the `E0` from the normal reference isentrope, which I _believe_ should be sufficient and consistent in the rest of the EOS since `es` is involved in all the energy expressions.

I also added a comment basically repeating the need described in https://github.com/lanl/singularity-eos/issues/260 for an iterative procedure to find the CJ energy from a reference state.

NOTE: This is really just a quick fix so I haven't added any unit tests. I hope to add a more complete set of tests later with a bit more of a revamp to the EOS.

## PR Checklist

### DIFF
--- a/singularity-eos/eos/eos_davis.hpp
+++ b/singularity-eos/eos/eos_davis.hpp
@@ -286,8 +286,8 @@ class DavisProducts : public EosBase<DavisProducts> {
   static inline unsigned long max_scratch_size(unsigned int nelements) { return 0; }
   PORTABLE_INLINE_FUNCTION void PrintParams() const {
     static constexpr char s1[]{"DavisProducts Params: "};
-    printf("%sa:%e b:%e k:%e\nn:%e vc:%e pc:%e\nCv:%e E0:%e\n", s1, _a, _b, _k, _n, _vc,
-           _pc, _Cv);
+    printf("%sa:%e b:%e k:%e\nn:%e vc:%e pc:%e\nCv:%e \n", s1, _a, _b, _k, _n, _vc, _pc,
+           _Cv);
   }
   inline void Finalize() {}
   static std::string EosType() { return std::string("DavisProducts"); }
--- a/singularity-eos/eos/eos_davis.hpp
+++ b/singularity-eos/eos/eos_davis.hpp
@@ -177,8 +177,8 @@ class DavisProducts : public EosBase<DavisProducts> {
   DavisProducts() = default;
   PORTABLE_INLINE_FUNCTION
   DavisProducts(const Real a, const Real b, const Real k, const Real n, const Real vc,
-                const Real pc, const Real Cv, const Real E0)
-      : _a(a), _b(b), _k(k), _n(n), _vc(vc), _pc(pc), _Cv(Cv), _E0(E0) {}
+                const Real pc, const Real Cv)
+      : _a(a), _b(b), _k(k), _n(n), _vc(vc), _pc(pc), _Cv(Cv), {}
   DavisProducts GetOnDevice() { return *this; }
   template <typename Indexer_t = Real *>
   PORTABLE_INLINE_FUNCTION Real TemperatureFromDensityInternalEnergy(
@@ -287,7 +287,7 @@ class DavisProducts : public EosBase<DavisProducts> {
   PORTABLE_INLINE_FUNCTION void PrintParams() const {
     static constexpr char s1[]{"DavisProducts Params: "};
     printf("%sa:%e b:%e k:%e\nn:%e vc:%e pc:%e\nCv:%e E0:%e\n", s1, _a, _b, _k, _n, _vc,
-           _pc, _Cv, _E0);
+           _pc, _Cv);
   }
   inline void Finalize() {}
   static std::string EosType() { return std::string("DavisProducts"); }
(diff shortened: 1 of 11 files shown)


jhp-lanl:
Updated the documentation to reflect the change and explain how to calculate `E0` from the CJ state

singularity-eos/eos/eos_davis.hpp
```diff
@@ -292,6 +292,7 @@ class DavisProducts : public EosBase<DavisProducts> {
   inline void Finalize() {}
   static std::string EosType() { return std::string("DavisProducts"); }
   static std::string EosPyType() { return EosType(); }
+  // TODO (JHP): Create helper function to find the CJ state given the reference state
```
Yurlungur, line 295: :+1: I'm in favor of the helper being `static` too.

Yurlungur (approved):
Looks like there's still some TODOs in here. And I'd like @chadmeyer 's eyes on it. We should maybe also add a unit test, and trigger it on Darwin.

That said I'm going to approve now, in case you need to get this in quickly for EAP integration work.

jhp-lanl:
> Looks like there's still some TODOs in here. And I'd like @chadmeyer 's eyes on it. We should maybe also add a unit test, and trigger it on Darwin.

I intend to address the TODO in a future MR. While it won't be overly complicated to implement the CJ state solver, it is a bit more work than I _need_ to get things working in EAP. Philosophically, it's also different from the changes here; this is a fix since `E0` was irrelevant before, whereas the CJ state solver is a new feature.

@chadmeyer this should be pretty fast to review

chadmeyer:
There is a potential issue here.  I was talking with Andrew Henrick, and for consistency, there might need to be other changes.  For instance, $P_s = \frac{e_s}{V}(k-1+F(V))$.  The point here is that if we shift the isentrope, it might mess with the pressure in unexpected ways.

jhp-lanl:
> There is a potential issue here. I was talking with Andrew Henrick, and for consistency, there might need to be other changes. For instance, Ps=esV(k−1+F(V)). The point here is that if we shift the isentrope, it might mess with the pressure in unexpected ways.

We don't use that functional form in the code as it exists right now. We calcualte the isentrope pressure as
```c++
PORTABLE_INLINE_FUNCTION Real Ps(const Real rho) const {
    const Real vvc = 1 / (rho * _vc);
    return _pc * std::pow(0.5 * (std::pow(vvc, _n) + std::pow(vvc, -_n)), _a / _n) /
           std::pow(vvc, _k + _a) * (_k - 1.0 + F(rho)) / (_k - 1.0 + _a);
  }
```
I haven't gone through the math to prove to myself that this is just the derivative of the energy along the isentrope, but it should be right? And if I'm just adding a constant offset to the energy, then the derivative shouldn't change.

chadmeyer:
> I haven't gone through the math to prove to myself that this is just the derivative of the energy along the isentrope, but it should be right? And if I'm just adding a constant offset to the energy, then the derivative shouldn't change.

In principle I think this is all right.  I just went through the whole code again, and it looks like this will be right.  But that begs the question of whether we should have an `E0` at all here... If we are correct, then it's just an EOS shift, and we already have machinery for that and maybe we just want to remove the option of doing it custom for this EOS?

jhp-lanl:
> > I haven't gone through the math to prove to myself that this is just the derivative of the energy along the isentrope, but it should be right? And if I'm just adding a constant offset to the energy, then the derivative shouldn't change.
> 
> In principle I think this is all right. I just went through the whole code again, and it looks like this will be right. But that begs the question of whether we should have an `E0` at all here... If we are correct, then it's just an EOS shift, and we already have machinery for that and maybe we just want to remove the option of doing it custom for this EOS?

Ah interesting point... that might be the better solution. I'll see if I can get that working in my tests

jhp-lanl:
I got the energy shift working in my tests and I think that's the most clear way to make the shift happen. I've removed the `E0` from the Davis Products EOS and put in the documentation that the user should explicitly use our energy shift modifier to accomplish the task of properly connect the products EOS to the desired reference state.

chadmeyer:
> I've removed the `E0` from the Davis Products EOS and put in the documentation that the user should explicitly use our energy shift modifier to accomplish the task of properly connect the products EOS to the desired reference state.

I approve of this improvement to the implementation!

chadmeyer (approved):

singularity-eos/eos/eos_davis.hpp
Yurlungur: Given that we now use the shift to set the energy of the CJ state, I am doubly convinced we should be providing a method to compute it, so that the shift can be straightforwardly generated. I suppose it doesn't need to be static necessarily... since you could generate an EOS, compute the energy, then shift it.

Yurlungur (approved):

singularity-eos/eos/eos_davis.hpp
jhp-lanl: Yeah... since the modifier takes an initialized object, it seems most prudent to have it _not_ be static so that you don't have to copy/paste parameters and risk an error.

We might want to discuss what this should look like on the Fortran side though. I'll make a comment in the issue.

jhp-lanl:
Merging now that I stomped out all of the instances of `E0` in the python bindings

PR #374, opened 2024-05-11
Various robustness additions to the Davis EOS

jhp-lanl:
## PR Summary
Three primary additions to the Davis Reactants and Products EOS to make them more robust:
1. Added the `robust::ratio` function when dividing by density
2. Added early returns for non-positive densities. Since the products EOS should exhibit ideal gas behavior at zero density, it makes sense to return zeros for the reference values.
3. Maxed the density with zero in the reactants EOS. With the robust ratio this basically just returns a low density limit. I don't think there's an overflow issue, but I could be wrong.

## PR Checklist

### DIFF
--- a/singularity-eos/eos/eos_davis.hpp
+++ b/singularity-eos/eos/eos_davis.hpp
@@ -379,12 +379,14 @@ PORTABLE_INLINE_FUNCTION Real DavisReactants::Es(const Real rho) const {
   } else {
     e_s = -y - (1.0 - std::exp(b4y)) / (4.0 * _B);
   }
-  return _e0 + _P0 * (1.0 / _rho0 - robust::ratio(1.0, std::max(rho, 0.))) + phat / _rho0 * e_s;
+  return _e0 + _P0 * (1.0 / _rho0 - robust::ratio(1.0, std::max(rho, 0.))) +
+         phat / _rho0 * e_s;
 }
 PORTABLE_INLINE_FUNCTION Real DavisReactants::Ts(const Real rho) const {
   if (rho >= _rho0) {
     const Real y = 1 - robust::ratio(_rho0, std::max(rho, 0.));
-    return _T0 * std::exp(-_Z * y) * std::pow(robust::ratio(_rho0, std::max(rho, 0.)), -_G0 - _Z);
+    return _T0 * std::exp(-_Z * y) *
+           std::pow(robust::ratio(_rho0, std::max(rho, 0.)), -_G0 - _Z);
   } else {
     return _T0 * std::pow(robust::ratio(_rho0, rho), -_G0);
   }
@@ -414,8 +416,9 @@ PORTABLE_INLINE_FUNCTION Real DavisReactants::BulkModulusFromDensityInternalEner
                  3 * y / pow<4>(1 - y) + 4 * pow<2>(y) / pow<5>(1 - y))
           : -phat * 4 * _B * _rho0 * std::exp(b4y);
   const Real gammav = (rho >= _rho0) ? _Z * _rho0 : 0.0;
-  return -(psv + (sie - Es(rho)) * rho * (gammav - gamma * std::max(rho, 0.)) - robust::ratio(gamma * std::max(rho, 0.) * esv),
-         rho);
+  return -(psv + (sie - Es(rho)) * rho * (gammav - gamma * std::max(rho, 0.)) -
+               robust::ratio(gamma * std::max(rho, 0.) * esv),
+           rho);
 }
 
 template <typename Indexer_t>
--- a/singularity-eos/eos/eos_davis.hpp
+++ b/singularity-eos/eos/eos_davis.hpp
@@ -22,6 +22,7 @@
 #include <singularity-eos/base/constants.hpp>
 #include <singularity-eos/base/math_utils.hpp>
 #include <singularity-eos/base/root-finding-1d/root_finding.hpp>
+#include <singularity-eos/base/robust_utils.hpp>
 #include <singularity-eos/eos/eos_base.hpp>
 
 namespace singularity {
--- a/singularity-eos/eos/eos_davis.hpp
+++ b/singularity-eos/eos/eos_davis.hpp
@@ -18,6 +18,7 @@
 #include <cmath>
 #include <cstdio>
 
+#include <ports-of-call/portable_errors.hpp>
 #include <singularity-eos/base/constants.hpp>
 #include <singularity-eos/base/math_utils.hpp>
 #include <singularity-eos/base/root-finding-1d/root_finding.hpp>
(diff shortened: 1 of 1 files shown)


jhp-lanl:
@chadmeyer please review these changes and see if you agree with my approach

singularity-eos/eos/eos_davis.hpp
```diff
@@ -328,7 +354,7 @@ class DavisProducts : public EosBase<DavisProducts> {
 
 PORTABLE_INLINE_FUNCTION Real DavisReactants::Ps(const Real rho) const {
   using namespace math_utils;
-  const Real y = 1.0 - _rho0 / rho;
+  const Real y = 1.0 - robust::ratio(_rho0, rho);
```
chadmeyer, line 357: Also see comment below.  If rho is slightly negative, y will tend to $+\infty$.  Is this a problem?  If rho is zero, then the value should be `-phat`, but it will be slightly not that (I doubt it matters unless there is some kind of numerical integration scheme being used on this routine).

singularity-eos/eos/eos_davis.hpp
```diff
@@ -342,7 +368,7 @@ PORTABLE_INLINE_FUNCTION Real DavisReactants::Ps(const Real rho) const {
   }
 }
 PORTABLE_INLINE_FUNCTION Real DavisReactants::Es(const Real rho) const {
-  const Real y = 1 - _rho0 / rho;
+  const Real y = 1 - robust::ratio(_rho0, std::max(rho, 0.));
```
chadmeyer, line 371: Is this (max(rho,0)) desired?  Is it necessary to guard against negative rho?  If so, we should be consistent between above and here.

singularity-eos/eos/eos_davis.hpp
```diff
@@ -354,19 +380,21 @@ PORTABLE_INLINE_FUNCTION Real DavisReactants::Es(const Real rho) const {
...
+  return _e0 + _P0 * (1.0 / _rho0 - robust::ratio(1.0, std::max(rho, 0.))) +
+         phat / _rho0 * e_s;
 }
 PORTABLE_INLINE_FUNCTION Real DavisReactants::Ts(const Real rho) const {
   if (rho >= _rho0) {
-    const Real y = 1 - _rho0 / rho;
-    return _T0 * std::exp(-_Z * y) * std::pow(_rho0 / rho, -_G0 - _Z);
+    const Real y = 1 - robust::ratio(_rho0, std::max(rho, 0.));
+    return _T0 * std::exp(-_Z * y) *
+           std::pow(robust::ratio(_rho0, std::max(rho, 0.)), -_G0 - _Z);
```
chadmeyer, line 390: In this branch rho is always > 0.

singularity-eos/eos/eos_davis.hpp
```diff
@@ -354,19 +380,21 @@ PORTABLE_INLINE_FUNCTION Real DavisReactants::Es(const Real rho) const {
...
   } else {
-    return _T0 * std::pow(_rho0 / rho, -_G0);
+    return _T0 * std::pow(robust::ratio(_rho0, rho), -_G0);
```
chadmeyer, line 392: In this branch where rho could go to zero, it might make sense to make sure rho is nonnegative.  In the end, you might just want to store `robust::ratio(_rho0.,std::max(rho,0.0))` at the beginning of this routine since it is re-used multiple times.

singularity-eos/eos/eos_davis.hpp
```diff
@@ -389,15 +417,18 @@ PORTABLE_INLINE_FUNCTION Real DavisReactants::BulkModulusFromDensityInternalEner
                  3 * y / pow<4>(1 - y) + 4 * pow<2>(y) / pow<5>(1 - y))
           : -phat * 4 * _B * _rho0 * std::exp(b4y);
   const Real gammav = (rho >= _rho0) ? _Z * _rho0 : 0.0;
-  return -(psv + (sie - Es(rho)) * rho * (gammav - gamma * rho) - gamma * rho * esv) /
-         rho;
+  const Real numerator =
+      -(psv + (sie - Es(rho)) * rho * (gammav - gamma * std::max(rho, 0.)) -
+        gamma * std::max(rho, 0.) * esv);
+  return robust::ratio(numerator, rho);
```
chadmeyer, line 423: Is there a reason for some of the rhos having a max and some not?

singularity-eos/eos/eos_davis.hpp
```diff
@@ -72,8 +79,11 @@ class DavisReactants : public EosBase<DavisReactants> {
   template <typename Indexer_t = Real *>
   PORTABLE_INLINE_FUNCTION Real MinInternalEnergyFromDensity(
       const Real rho, Indexer_t &&lambda = static_cast<Real *>(nullptr)) const {
-    MinInternalEnergyIsNotEnabled("DavisReactants");
-    return 0.0;
+    // Minimum enegy is when the returned temperature is zero. This only happens
+    // when the base to the exponent is zero (see T(rho, e) equation)
+    const Real es = Es(rho);
+    const Real ts = Ts(rho);
+    return es - (1 + _alpha) / (_Cv0 * ts);
```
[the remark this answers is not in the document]
jhp-lanl: Ah good point. Thanks for the catch. I flipped that in my math. I'll change shortly

singularity-eos/eos/eos_davis.hpp
```diff
@@ -100,8 +110,12 @@ class DavisReactants : public EosBase<DavisReactants> {
   PORTABLE_INLINE_FUNCTION Real SpecificHeatFromDensityInternalEnergy(
       const Real rho, const Real sie,
       Indexer_t &&lambda = static_cast<Real *>(nullptr)) const {
-    return _Cv0 / std::pow((1 + _alpha) / (Ts(rho) * _Cv0) * (sie - Es(rho)) + 1,
-                           -_alpha / (1 + _alpha));
+    const Real power_base = (1 + _alpha) / (Ts(rho) * _Cv0) * (sie - Es(rho)) + 1;
```
Yurlungur, line 113: Since this base is now being explicitly pulled out and checked there's maybe some opportunity to unify it with the other places it's called.

singularity-eos/eos/eos_davis.hpp
Yurlungur: You could enforce `rho` positivity by doing `robust::ratio(_rho0, std::max(0, rho));` or something similar

singularity-eos/eos/eos_davis.hpp
```diff
@@ -418,6 +449,7 @@ PORTABLE_INLINE_FUNCTION void DavisReactants::FillEos(Real &rho, Real &temp, Rea
                                                       Real &press, Real &cv, Real &bmod,
                                                       const unsigned long output,
                                                       Indexer_t &&lambda) const {
+  // BROKEN: This can only handle density-energy inputs! MUST BE CHANGED
```
Yurlungur, line 452: Good catch. Yes we should fix that.

Yurlungur (approved):
I defer to @chadmeyer 's comments. But other than what chad picked out I'm happy. Thanks for the fixes.

singularity-eos/eos/eos_davis.hpp
jhp-lanl: Yes that is the intent. I'll correct the above

singularity-eos/eos/eos_davis.hpp
jhp-lanl: For this comment and the previous, I had eliminated negative density branches in favor of just forcing a positive density. I'll do that here too

singularity-eos/eos/eos_davis.hpp
jhp-lanl: No... I was just probably working too fast. I've added `max` to all `rho` now here

singularity-eos/eos/eos_davis.hpp
jhp-lanl: I broke it out into a function

jhp-lanl:
@chadmeyer I think I addressed all your comments. I'll give things another scan and make sure I didn't miss anything else silly, but I think this is good now.

jhp-lanl:
FYI - I added a few more robustness additions. Mainly I added an error to the `DensityEnergyFromPressureTemperature()` functions if the resultant density is negative (adding a max here probably will just cause the function to not converge and produce a less informative error).

I also added a `robust::ratio` to where we divide by `Ts`. It is possible for this to underflow which _might_ be zero.

Issue #318, opened 2023-10-27
Provide default functionality for MinInternalEnergyFromDensity()


jhp-lanl:
It might make sense to just use the zero K isotherm lookup as the minimum energy at least for thermodynamically consistent EOS.

jhp-lanl:
Well, technically this is really resolved by https://github.com/lanl/singularity-eos/pull/375. The current default functionality is an error whereas that MR changes it to be in line with what I proposed here.

I'm not pedantic enough to re-open this though 😉

PR #375, opened 2024-05-11
Create minimum energy modifier

jhp-lanl:
## PR Summary

Creates a minimum energy modifier and unifies EOS behind a default behavior of returning the zero K isotherm as the minimum possible internal energy.

Some notes:
1. I duplicated @rbberger 's approach with the shifted EOS using scratch memory. Similar to populating a shift, I populate a minimum energy and then max that with the input energy to then pass to whatever function needs the internal energy.
2. I am assuming that ***all*** energies are valid when they are outputs rather than inputs. I feel this is reasonable since the EOS should be self-consistent enough to not return energies off the EOS surface. If there is an issue, it might point to badly bounded inputs (i.e. negative temperatures or densities).
3. Initially this is very rough and is guaranteed to not compile. I'll need to make tests and make sure that the Gruneisen EOS can properly call the base class version within its own version and that all the other classes can properly see both the vector and scalar overloads from the base class.

Fixes https://github.com/lanl/singularity-eos/issues/318

## PR Checklist


jhp-lanl:
Wow it compiled first time!? The modifier isn't compiled yet, but the default minimum energy stuff should be compiling, which might mean what I wrote works?

I'd still like to make an explicit unit test something along the lines of asserting that the energy is the zero temperature energy.

dholladay00:
Would this play nicely with the options that @annapiegra added for eospac @jhp-lanl ? It sounds similar though I don't recall the details. 

I assume it matters which order an energy shift is applied so that will need to be taken into account when constructing the modifier list.

jhp-lanl:
> Would this play nicely with the options that @annapiegra added for eospac @jhp-lanl ? It sounds similar though I don't recall the details.
> 
> I assume it matters which order an energy shift is applied so that will need to be taken into account when constructing the modifier list.

This is the exact function that @annapiegra added, but when it was added it was basically disabled (would throw an error) for any EOS other than EOSPAC. This basically changes that so that _every_ EOS will have some notion of minimum energy by default.

singularity-eos/eos/modifiers/minimum_energy.hpp
```diff
@@ -0,0 +1,378 @@
...
+  PORTABLE_FUNCTION Real TemperatureFromDensityInternalEnergy(
+      const Real rho, const Real sie, Indexer_t &&lambda = nullptr) const {
+    const Real min_sie = t_.MinInternalEnergyFromDensity(rho);
+    return t_.TemperatureFromDensityInternalEnergy(rho, std::max(sie, min_sie), lambda);
+  }
+  template <typename Indexer_t = Real *>
+  PORTABLE_FUNCTION Real InternalEnergyFromDensityTemperature(
+      const Real rho, const Real temperature, Indexer_t &&lambda = nullptr) const {
+    return t_.InternalEnergyFromDensityTemperature(rho, temperature, lambda);
+  }
```
Yurlungur, line 80: We may wish to compute t of emin or something for rho e based EOS's. Not sure that's worth thinking about here though.

Yurlungur (commented):
Very clean how this came out. I wonder if we should make this more "monolithic" and just provide bounds on all independent variables.

singularity-eos/eos/modifiers/minimum_energy.hpp
```diff
@@ -0,0 +1,378 @@
...
+#include <ports-of-call/portability.hpp>
+#include <singularity-eos/base/eos_error.hpp>
+#include <singularity-eos/eos/eos_base.hpp>
+
+namespace singularity {
+
+using namespace eos_base;
+
+template <typename T>
+class MinimumEnergy : public EosBase<MinimumEnergy<T>> {
```
jhp-lanl, line 31: I'm open to renaming this if someone has a more descriptive name

singularity-eos/eos/modifiers/minimum_energy.hpp
jhp-lanl: Are you talking about maxing the input temperature _before_ outputting the energy?

I think more generally though if the energy output here violates the minimum given by the `MinInternalEnergyFromDensity()` function, then it's a reflection of an unphysical EOS (i.e. a negative heat capacity). As a result, I'm not sure we want to guard against this.

That said, it's certainly possible to have a negative heat capacity in some fringe cases, but I'm not sure if this is the appropriate way to guard against this. But then again, maybe it's better to guarantee that the output is always bounded by the minimum.

jhp-lanl:
> Very clean how this came out. I wonder if we should make this more "monolithic" and just provide bounds on all independent variables.

I assume that would mean expanding this to enforce a minimum density and temperature as well? That seems pretty straight-foward and might be pretty easy to implement.

I don't know if I'd want to do anything with upper bounds though. I feel that an extrapolation approach might be a superior way to handle upper-bounds.

Speaking of extrapolation, I suppose that if a table has a non-zero lower temperature bound we might want to apply an extrapolation layer first that extends the minimum energy/temperature to 0 K first and then this would lie on top of it.

singularity-eos/eos/modifiers/minimum_energy.hpp
Yurlungur: Maybe `FlooredEnergy` or something?

singularity-eos/eos/modifiers/minimum_energy.hpp
Yurlungur: > Are you talking about maxing the input temperature before outputting the energy?

That was what I was thinking but actually it probably makes more sense to max it after. Either way it may not be worth worrying about here.

Yurlungur:
> > Very clean how this came out. I wonder if we should make this more "monolithic" and just provide bounds on all independent variables.
> 
> I assume that would mean expanding this to enforce a minimum density and temperature as well? That seems pretty straight-foward and might be pretty easy to implement.

I think it might be nice to impose energy and temperature floors together, yeah. :+1:

> I don't know if I'd want to do anything with upper bounds though. I feel that an extrapolation approach might be a superior way to handle upper-bounds.

That's fair... but is that always going to be valid going forward? What about an EOS that is for, e.g., a single phase as @aematts is planning to use?

> Speaking of extrapolation, I suppose that if a table has a non-zero lower temperature bound we might want to apply an extrapolation layer first that extends the minimum energy/temperature to 0 K first and then this would lie on top of it.

True! I actually think what you've implemented here is a nice template for that kind of thing.

jhp-lanl:
> That's fair... but is that always going to be valid going forward? What about an EOS that is for, e.g., a single phase as @aematts is planning to use?

That's a good point, but I think there's an important difference: I'm constraining the _input_ to be on the EOS surface whereas I think @aematts is going to want to constrain the _output_ such that it's clear when a phase EOS isn't valid.

From my limited understanding of kinetic phase transitions, you might want to set a domain of validity for a phase EOS and not _constrain_ yourself to that domain, but instead maybe provide a flag to indicate that you're not in the range of validity anymore. For example, you could return an infinite Gibbs free energy (or equivalent) when you're outside the domain of validity (which is basically a constant extrapolation). That way the rate at which that phase is created can be set to identically zero.

Either way, I anticipate that we'll have to create additional modifiers to handle these use scenarios.

singularity-eos/eos/modifiers/minimum_energy.hpp
jhp-lanl: I like that better. I'll change it when I get back

Yurlungur:
> > That's fair... but is that always going to be valid going forward? What about an EOS that is for, e.g., a single phase as @aematts is planning to use?
> 
> That's a good point, but I think there's an important difference: I'm constraining the _input_ to be on the EOS surface whereas I think @aematts is going to want to constrain the _output_ such that it's clear when a phase EOS isn't valid.
> 
> From my limited understanding of kinetic phase transitions, you might want to set a domain of validity for a phase EOS and not _constrain_ yourself to that domain, but instead maybe provide a flag to indicate that you're not in the range of validity anymore. For example, you could return an infinite Gibbs free energy (or equivalent) when you're outside the domain of validity (which is basically a constant extrapolation). That way the rate at which that phase is created can be set to identically zero.
> 
> Either way, I anticipate that we'll have to create additional modifiers to handle these use scenarios.

Yeah fair enough. I may be over-engineering here.

jhp-lanl:
@Yurlungur I think this is ready for final review (sorry for the long delay)

singularity-eos/eos/modifiers/floored_energy.hpp
```diff
@@ -0,0 +1,374 @@
...
+  using EosBase<FlooredEnergy<T>>::MinInternalEnergyFromDensity;
+  using EosBase<FlooredEnergy<T>>::EntropyFromDensityTemperature;
+  using EosBase<FlooredEnergy<T>>::EntropyFromDensityInternalEnergy;
+  using EosBase<FlooredEnergy<T>>::SpecificHeatFromDensityTemperature;
+  using EosBase<FlooredEnergy<T>>::SpecificHeatFromDensityInternalEnergy;
+  using EosBase<FlooredEnergy<T>>::BulkModulusFromDensityTemperature;
+  using EosBase<FlooredEnergy<T>>::BulkModulusFromDensityInternalEnergy;
+  using EosBase<FlooredEnergy<T>>::GruneisenParamFromDensityTemperature;
+  using EosBase<FlooredEnergy<T>>::GruneisenParamFromDensityInternalEnergy;
+  using EosBase<FlooredEnergy<T>>::FillEos;
```
Yurlungur, line 50: There are now some more general macros that can replace these lines defined in `eos_base.hpp`

Yurlungur (approved):
Non-blocking request. But overall looks good!

singularity-eos/eos/modifiers/floored_energy.hpp
jhp-lanl: Yes! I'll change that

jhp-lanl:
Oh and I still need to test this code. I'll try to do that here shortly.

singularity-eos/eos/modifiers/floored_energy.hpp
jhp-lanl: Fixed

jhp-lanl:
@jonahm-LANL It looks like `SpinerEOSDependsRhoSie` doesn't have the same cold curve infrastructure that `SpinerEOSDependsRhoT` does so I left that function disabled. It's probably faster for you to add the required functionality than it is for me.

This is ready for final review (really just the test was added, but there are a few more changes)

jhp-lanl:
CI is failing on LLNL systems. DO NOT MERGE

Yurlungur:
> @jonahm-LANL It looks like `SpinerEOSDependsRhoSie` doesn't have the same cold curve infrastructure that `SpinerEOSDependsRhoT` does so I left that function disabled. It's probably faster for you to add the required functionality than it is for me.

Sounds good. I'll add it in a subsequent MR.

> This is ready for final review (really just the test was added, but there are a few more changes)

doc/sphinx/src/modifiers.rst
```diff
@@ -301,6 +301,42 @@ parameter via the lambda. For example:
...
+   the lowest isotherm's energy in place of a cold curve. If a given table does
+   not extend to zero K, then an extrapolaiton is *not* performed. This results
+   in an energy floor from the lowest temperature on the table, **not** from
+   an extrapolated energy at zero K.
+
+.. note::
+
+  This modifier is only compatible with EOS where the
+  ``MinInternalEnergyFromDensity`` function is properly defined. Where it isn't
+  defined, an error will occur at runtime.
```
Yurlungur, line 337: :+1: for careful documentation here

singularity-eos/eos/modifiers/floored_energy.hpp
```diff
@@ -0,0 +1,359 @@
...
+    return t_.MinimumTemperature();
+  }
+
+  inline constexpr bool IsModified() const { return true; }
+
+  inline constexpr T UnmodifyOnce() { return t_; }
+
+  inline constexpr decltype(auto) GetUnmodifiedObject() {
+    return t_.GetUnmodifiedObject();
+  }
```
Yurlungur, line 351:
```suggestion
  PORTABLE_FORCEINLINE_FUNCTION Real MaximumDensity() const {
    return t_.MaximumDensity();
  }
  PORTABLE_FORCEINLINE_FUNCTION Real MinimumPressure() const {
    return t_.MinimumPressure();
  }
  PORTABLE_FORCEINLINE_FUNCTION Real MaximumPressureAtTemperature(const Real temp) const {
    return t_.MaximumPressureAtTemperature(temp);
  }

  template <typename Indexer_t = Real *>
  PORTABLE_INLINE_FUNCTION Real MeanAtomicMassFromDensityTemperature(
      const Real rho, const Real temperature,
      Indexer_t &&lambda = static_cast<Real *>(nullptr)) const {
    return t_.MeanAtomicMassFromDensityTemperature(rho, temperature, lambda);
  }
  template <typename Indexer_t = Real *>
  PORTABLE_INLINE_FUNCTION Real MeanAtomicNumberFromDensityTemperature(
      const Real rho, const Real temperature,
      Indexer_t &&lambda = static_cast<Real *>(nullptr)) const {
    return t_.MeanAtomicNumberFromDensityTemperature(rho, temperature, lambda);
  }

  SG_ADD_MODIFIER_METHODS(T, t_);
  SG_ADD_MODIFIER_MEAN_METHODS(t_)
```

test/eos_unit_test_helpers.hpp
```diff
@@ -82,6 +83,26 @@ inline void array_compare(int num, X &&x, Y &&y, Z &&z, ZT &&ztrue, XN xname, YN
...
+}
+
+// Helper function to call Finalize() on each eos in an array (host side)
+template <typename EOSArrT>
+void finalize_eos_arr(EOSArrT eos_arr) {
+  // Call Finalize on each EOS on the host
+  for (auto eos : eos_arr) {
+    eos.Finalize();
+  }
+}
```
Yurlungur, line 104: nice

Yurlungur (approved):
Looks good. One more suggestion caught. I think I fixed the problems on re-git in #473. But we should double check that before merging this.

singularity-eos/eos/modifiers/floored_energy.hpp
jhp-lanl: Ah my bad, I forgot how old this had become. I'll add those

test/test_modifier_floored_energy.cpp
```diff
@@ -50,8 +50,8 @@ using EOS =
 // temperature for the lookup is T_lookup. The energy for the lookup is given by
 // e(rho, T_lookup) - e_offset
 PORTABLE_INLINE_FUNCTION
-auto diff_pressures(const int n_eos, EOS *v_EOS, const Real T_lookup,
-                    const Real e_offset = 0., const Real rho_factor = 1.2) {
+std::vector<Real> diff_pressures(const int n_eos, EOS *v_EOS, const Real T_lookup,
```
dholladay00, line 53: should this be a device callable function?

Yurlungur:
@jhp-lanl is this ready for merge?

test/test_modifier_floored_energy.cpp
jhp-lanl: Nope! I just removed it 🙂

jhp-lanl:
> @jhp-lanl is this ready for merge?

Almost...

jhp-lanl:
Annoyingly, I didn't realize that a parent function for a device lambda can't have a deduced return type. In my case, my function was calling a `portableFor()` with a `PORTABLE_LAMBDA`, and even though the parent function created a `std::vector` on the host and returned it.

I think that fixes the issue on CUDA platforms. The pipeline should be passing on re-git in the morning, and I can merge then unless you beat me to it.

jhp-lanl:
All tests pass on LANL machines

PR #377, opened 2024-05-23
Separate out potentially useful type lists and variant utility functi…

dholladay00:
…ons to separate header files that could be used downstream w/o needing the variant definition itself.

## PR Summary

@Yurlungur @jhp-lanl @gopsub @jdolence 

This PR includes changes that separates potentially useful type lists and variant utility functions into separate header files that could be used by downstream codes without needing to include the variant type itself. It allows one to start with the final monolithic list of eos types (or any of the intermediate type lists) and build from there using concat, transform etc. Then the final variant's type can be constructed using the variant utility functions. 

## PR Checklist

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -40,8 +40,6 @@ default:
   id_tokens:
     SITE_ID_TOKEN:
       aud: https://asc-git.lanl.gov
-    CI_JOB_JWT:
-      aud: https://asc-git.lanl.gov
 
 ##########################
 # Build Matrix Variables #
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -40,6 +40,8 @@ default:
   id_tokens:
     SITE_ID_TOKEN:
       aud: https://asc-git.lanl.gov
+    CI_JOB_JWT:
+      aud: https://asc-git.lanl.gov
 
 ##########################
 # Build Matrix Variables #
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Current develop
 
 ### Added (new features/APIs/variables/...)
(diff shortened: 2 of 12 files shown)


Yurlungur (approved):
I approve of exposing this machinery. Just need to make sure tests pass. Probably merits a changelog entry too.

jhp-lanl (approved):
Looks good to me! And good timing! I think we'll probably be using this here soon (i.e. months) in some of the integration work

.gitlab-ci.yml
```diff
@@ -40,6 +40,8 @@ default:
   id_tokens:
     SITE_ID_TOKEN:
       aud: https://asc-git.lanl.gov
+    CI_JOB_JWT:
```
dholladay00, line 43: @rbberger is this change still needed?

dholladay00:
after comments are addressed by @rbberger we can merge @Yurlungur

.gitlab-ci.yml
rbberger: I have no other pipeline or project that needed that change.

dholladay00: That was the change  that the CI was asking for on other projects.

Yurlungur: I'm confused---do we need this change or no?

rbberger: No. I don't think so. My theory is that re-git switched over to the new version (the thing that they were warning for many months), which deprecated `CI_JOB_JWT` and that there was some outdated runners still floating around while me and Karen were out. The outdated runner would then complain about not getting the no longer provided token. All it needed was to get that runner updated to accept the new SITE_ID_TOKEN, which seems to have eventually happened.

dholladay00:
This is passing gitlab ci, clicking the button.

PR #380, opened 2024-06-18
Add zeroing out of material internal energy if its mass fraction is s…

dholladay00:
…ufficiently low so as not to be participating in the solve. This fixes issues in the event the passed in internal energies are uninitialized.

## PR Summary

This PR fixes issues seen with uninitialized data. If internal energies that are passed are uninitialized coming into the get_sg_eos_function, they may be remain so and cause downstream checks to fail. 

This is not the full changeset I had for debugging. As of now, the get_sg_eos function will always return 0. In the past, the value passed back was the number of failed PTE solves, though this should not necessarily result in a crash. That said, I had also added checks for nans in outputs, and if that occurs we may want to cause a hard stop on execution. 

Should that be part of a different PR @jhp-lanl ? Thoughts? This should be enough to get you going again though if you want to continue with debugging / vetting efforts. 

## PR Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 - [[PR#357]](https://github.com/lanl/singularity-eos/pull/357) Added support for C++17 (e.g., needed when using newer Kokkos).
 
 ### Fixed (Repair bugs, etc)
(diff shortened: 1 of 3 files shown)


jhp-lanl (approved):
Looks good to me. We can address the error return code in a separate MR

dholladay00:
Alright I'll pull the trigger on this when the checks pass @jhp-lanl.

Yurlungur (approved):

PR #382, opened 2024-06-26
Add introspection to `get_sg_eos` interface

jhp-lanl:
## PR Summary

When the library is built with debug flags, this adds introspection functions to the `init` and `final` functors in the `get_sg_eos` interface.

These functions check to make sure the values being passed are either zero or normal. I'm using slightly more complicated logic than just `is_finite` so that I can detect underflow values as well. For some quantities, I'm also including checks to make sure the returned value is strictly positive or non-negative where appropriate.

## PR Checklist

- N/A Update the version in cmake.
- N/A Move the changes in the CHANGELOG.md file under a new header for the new release, and reset the categories.

### DIFF
--- a/singularity-eos/base/error_utils.hpp
+++ b/singularity-eos/base/error_utils.hpp
@@ -1,8 +1,8 @@
 #ifndef SINGULARITY_EOS_BASE_ERROR_UTILS_HPP_
 #define SINGULARITY_EOS_BASE_ERROR_UTILS_HPP_
 
-#include <type_traits>
 #include <cmath>
+#include <type_traits>
 
 #include <ports-of-call/portability.hpp>
 
@@ -17,8 +17,8 @@ struct is_normal_or_zero {
   template <typename valT>
   bool PORTABLE_FORCEINLINE_FUNCTION operator()(valT value) const {
     static_assert(std::is_floating_point_v<valT>);
-    return (value == valT{0}) || (std::isnormal(_NORMAL_FACTOR * value) 
-                                   && std::isnormal(value));
+    return (value == valT{0}) ||
+           (std::isnormal(_NORMAL_FACTOR * value) && std::isnormal(value));
   }
 };
 
@@ -53,17 +53,17 @@ PORTABLE_FORCEINLINE_FUNCTION bool violates_condition(valT&& value, condT&& cond
 
 // Create specialized functions using the above conditions
 template <typename valT, typename nameT>
-PORTABLE_FORCEINLINE_FUNCTION bool bad_value(valT&& value, nameT&& var_name) {
+PORTABLE_FORCEINLINE_FUNCTION bool bad_value(valT &&value, nameT&& var_name) {
   return violates_condition(std::forward<valT>(value), is_normal_or_zero{},
                             std::forward<nameT>(var_name));
 }
 template <typename valT, typename nameT>
-PORTABLE_FORCEINLINE_FUNCTION bool non_positive_value(valT&& value, nameT&& var_name) {
+PORTABLE_FORCEINLINE_FUNCTION bool non_positive_value(valT &&value, nameT&& var_name) {
   return violates_condition(std::forward<valT>(value), is_strictly_positive{},
                             std::forward<nameT>(var_name));
 }
 template <typename valT, typename nameT>
-PORTABLE_FORCEINLINE_FUNCTION bool negative_value(valT&& value, nameT&& var_name) {
+PORTABLE_FORCEINLINE_FUNCTION bool negative_value(valT &&value, nameT&& var_name) {
   return violates_condition(std::forward<valT>(value), is_strictly_positive{},
                             std::forward<nameT>(var_name));
 }
--- a/singularity-eos/base/error_utils.hpp
+++ b/singularity-eos/base/error_utils.hpp
@@ -1,0 +1,74 @@
+#ifndef SINGULARITY_EOS_BASE_ERROR_UTILS_HPP_
+#define SINGULARITY_EOS_BASE_ERROR_UTILS_HPP_
+
+#include <type_traits>
+#include <cmath>
+
+#include <ports-of-call/portability.hpp>
+
+namespace singularity {
+namespace error_utils {
+
+using PortsOfCall::printf;
+
+constexpr double _NORMAL_FACTOR = 1.0e10;
+
+struct is_normal_or_zero {
+    template <typename valT>
+    bool PORTABLE_FORCEINLINE_FUNCTION operator()(valT value) const {
+        static_assert(std::is_floating_point_v<valT>);
+        return (value == valT{0}) || (std::isnormal(_NORMAL_FACTOR * value) 
+                                       && std::isnormal(value));
+    }
+};
+
+struct is_strictly_positive {
+    template <typename valT>
+    bool PORTABLE_FORCEINLINE_FUNCTION operator()(valT value) const {
+        static_assert(std::is_arithmetic_v<valT>);
+        return value > valT{0};
+    }
+};
+
+struct is_non_negative {
+    template <typename valT>
+    bool PORTABLE_FORCEINLINE_FUNCTION operator()(valT value) const {
+        static_assert(std::is_arithmetic_v<valT>);
+        return value >= valT{0};
+    }
+};
+
+// Checks whether a value obeys some sort of provided condition. If not, returns true and
+// prints the provided error message, variable name, and value (but does not abort!)
+template <typename valT, typename condT, typename nameT>
+PORTABLE_FORCEINLINE_FUNCTION bool violates_condition(valT&& value, condT&& condition,
+                                                      nameT&& var_name) {
+    const bool good = condition(std::forward<valT>(value));
+    if (!good) {
+        printf("### ERROR: Bad singularity-eos value\n  Var: %s\n Value: %.15e\n",
+               var_name, value);
+    }
+    return !good;
+}
+
+// Create specialized functions using the above conditions
+template <typename valT, typename nameT>
+PORTABLE_FORCEINLINE_FUNCTION bool bad_value(valT&& value, nameT&& var_name) {
+    return violates_condition(std::forward<valT>(value), is_normal_or_zero{},
+                              std::forward<nameT>(var_name));
+}
+template <typename valT, typename nameT>
+PORTABLE_FORCEINLINE_FUNCTION bool non_positive_value(valT&& value, nameT&& var_name) {
+    return violates_condition(std::forward<valT>(value), is_strictly_positive{},
+                              std::forward<nameT>(var_name));
+}
+template <typename valT, typename nameT>
+PORTABLE_FORCEINLINE_FUNCTION bool negative_value(valT&& value, nameT&& var_name) {
+    return violates_condition(std::forward<valT>(value), is_strictly_positive{},
+                              std::forward<nameT>(var_name));
+}
+
+} // namespace error_utils
+} // namesapce singularity
+
+#endif // #ifndef SINGULARITY_EOS_BASE_ERROR_UTILS_HPP_
--- a/singularity-eos/base/error_utils.hpp
+++ b/singularity-eos/base/error_utils.hpp
@@ -41,34 +41,34 @@ struct is_non_negative {
 // Checks whether a value obeys some sort of provided condition. If not, returns true and
 // prints the provided error message, variable name, and value (but does not abort!)
 template <typename valT, typename condT, typename nameT>
-PORTABLE_FORCEINLINE_FUNCTION bool violates_condition(valT&& value, condT&& condition,
-                                                    nameT&& var_name) {
+PORTABLE_FORCEINLINE_FUNCTION bool violates_condition(valT &&value, condT &&condition,
+                                                      nameT &&var_name) {
   const bool good = condition(std::forward<valT>(value));
   if (!good) {
-    printf("### ERROR: Bad singularity-eos value\n  Var: %s\n Value: %.15e\n",
-           var_name, value);
+    printf("### ERROR: Bad singularity-eos value\n  Var: %s\n Value: %.15e\n", var_name,
+           value);
   }
   return !good;
 }
 
 // Create specialized functions using the above conditions
 template <typename valT, typename nameT>
-PORTABLE_FORCEINLINE_FUNCTION bool bad_value(valT &&value, nameT&& var_name) {
+PORTABLE_FORCEINLINE_FUNCTION bool bad_value(valT &&value, nameT &&var_name) {
   return violates_condition(std::forward<valT>(value), is_normal_or_zero{},
                             std::forward<nameT>(var_name));
 }
 template <typename valT, typename nameT>
-PORTABLE_FORCEINLINE_FUNCTION bool non_positive_value(valT &&value, nameT&& var_name) {
+PORTABLE_FORCEINLINE_FUNCTION bool non_positive_value(valT &&value, nameT &&var_name) {
   return violates_condition(std::forward<valT>(value), is_strictly_positive{},
                             std::forward<nameT>(var_name));
 }
 template <typename valT, typename nameT>
-PORTABLE_FORCEINLINE_FUNCTION bool negative_value(valT &&value, nameT&& var_name) {
+PORTABLE_FORCEINLINE_FUNCTION bool negative_value(valT &&value, nameT &&var_name) {
   return violates_condition(std::forward<valT>(value), is_strictly_positive{},
                             std::forward<nameT>(var_name));
 }
 
 } // namespace error_utils
-} // namesapce singularity
+} // namespace singularity
 
 #endif // #ifndef SINGULARITY_EOS_BASE_ERROR_UTILS_HPP_
--- a/singularity-eos/base/error_utils.hpp
+++ b/singularity-eos/base/error_utils.hpp
@@ -1,3 +1,17 @@
+//------------------------------------------------------------------------------
+// © 2021-2024. Triad National Security, LLC. All rights reserved.  This
+// program was produced under U.S. Government contract 89233218CNA000001
+// for Los Alamos National Laboratory (LANL), which is operated by Triad
+// National Security, LLC for the U.S.  Department of Energy/National
+// Nuclear Security Administration. All rights in the program are
+// reserved by Triad National Security, LLC, and the U.S. Department of
+// Energy/National Nuclear Security Administration. The Government is
+// granted for itself and others acting on its behalf a nonexclusive,
+// paid-up, irrevocable worldwide license in this material to reproduce,
+// prepare derivative works, distribute copies to the public, perform
+// publicly and display publicly, and to permit others to do so.
+//------------------------------------------------------------------------------
+
 #ifndef SINGULARITY_EOS_BASE_ERROR_UTILS_HPP_
 #define SINGULARITY_EOS_BASE_ERROR_UTILS_HPP_
 
@@ -45,8 +59,8 @@ PORTABLE_FORCEINLINE_FUNCTION bool violates_condition(valT &&value, condT &&cond
                                                       nameT &&var_name) {
   const bool good = condition(std::forward<valT>(value));
   if (!good) {
-    printf("### ERROR: Bad singularity-eos value\n  Var: %s\n Value: %.15e\n", var_name,
-           value);
+    printf("### ERROR: Bad singularity-eos value\n  Var:   %s\n  Value: %.15e\n",
+           var_name, value);
   }
   return !good;
 }
--- a/singularity-eos/base/error_utils.hpp
+++ b/singularity-eos/base/error_utils.hpp
@@ -14,58 +14,58 @@ using PortsOfCall::printf;
 constexpr double _NORMAL_FACTOR = 1.0e10;
 
 struct is_normal_or_zero {
-    template <typename valT>
-    bool PORTABLE_FORCEINLINE_FUNCTION operator()(valT value) const {
-        static_assert(std::is_floating_point_v<valT>);
-        return (value == valT{0}) || (std::isnormal(_NORMAL_FACTOR * value) 
-                                       && std::isnormal(value));
-    }
+  template <typename valT>
+  bool PORTABLE_FORCEINLINE_FUNCTION operator()(valT value) const {
+    static_assert(std::is_floating_point_v<valT>);
+    return (value == valT{0}) || (std::isnormal(_NORMAL_FACTOR * value) 
+                                   && std::isnormal(value));
+  }
 };
 
 struct is_strictly_positive {
-    template <typename valT>
-    bool PORTABLE_FORCEINLINE_FUNCTION operator()(valT value) const {
-        static_assert(std::is_arithmetic_v<valT>);
-        return value > valT{0};
-    }
+  template <typename valT>
+  bool PORTABLE_FORCEINLINE_FUNCTION operator()(valT value) const {
+    static_assert(std::is_arithmetic_v<valT>);
+    return value > valT{0};
+  }
 };
 
 struct is_non_negative {
-    template <typename valT>
-    bool PORTABLE_FORCEINLINE_FUNCTION operator()(valT value) const {
-        static_assert(std::is_arithmetic_v<valT>);
-        return value >= valT{0};
-    }
+  template <typename valT>
+  bool PORTABLE_FORCEINLINE_FUNCTION operator()(valT value) const {
+    static_assert(std::is_arithmetic_v<valT>);
+    return value >= valT{0};
+  }
 };
 
 // Checks whether a value obeys some sort of provided condition. If not, returns true and
 // prints the provided error message, variable name, and value (but does not abort!)
 template <typename valT, typename condT, typename nameT>
 PORTABLE_FORCEINLINE_FUNCTION bool violates_condition(valT&& value, condT&& condition,
-                                                      nameT&& var_name) {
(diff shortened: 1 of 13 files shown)


jhp-lanl:
@Yurlungur @dholladay00 is going to be a bit unavailable for the next few days so I might need your review here.

Also please check me on whether you agree with what I'm requiring to be non-negative or strictly positive.

jhp-lanl:
FYI - I did some reading on `and` vs `&&` and it seems like the consensus is that because `&&` has been used traditionally, it is preferred over `and`. Coming from Fortran and python, I find `and` more readable, but I'm happy to change them to the more commonly-used C++ symbols.

Yurlungur:
> FYI - I did some reading on `and` vs `&&` and it seems like the consensus is that because `&&` has been used traditionally, it is preferred over `and`. Coming from Fortran and python, I find `and` more readable, but I'm happy to change them to the more commonly-used C++ symbols.

I strongly prefer `&&`. I wasn't even aware `and` was valid C++.

jhp-lanl:
> > FYI - I did some reading on `and` vs `&&` and it seems like the consensus is that because `&&` has been used traditionally, it is preferred over `and`. Coming from Fortran and python, I find `and` more readable, but I'm happy to change them to the more commonly-used C++ symbols.
> 
> I strongly prefer `&&`. I wasn't even aware `and` was valid C++.

Apparently it's a holdover from the ancient times when some keyboards had limited symbols believe it or not. Again, I just liked it because coming from a non-C++ background it seems more readable. I'll change it

singularity-eos/eos/get_sg_eos_functors.hpp
```diff
@@ -71,31 +73,89 @@ struct init_functor {
...
+    // Populate the inputs with consistent values.
+    // NOTE: the volume fractions and densities need to be consistent with the
+    // total specific volume since they are used to calculate internal
+    // quantities for the PTE solver
     for (int mp = 0; mp < npte; ++mp) {
       const int m = pte_mats(tid, mp);
-      rho_pte(tid, mp) = npte / spvol_v(i) * frac_mass_v(i, m);
-      vfrac_pte(tid, mp) = 1.0 / npte;
+      // Need to guess volume fractions
+      vfrac_pte(tid, mp) = frac_mass_v(i, m);
```
jhp-lanl, line 101: FYI - I'm changing the initial guess for volume fractions from equipartitioning of volume irrespective of mass fractions. This guess is equivalent to assuming that all material densities are the same.

This got me further in my test problem than assuming that all volume fractions are the same. In the future I'll see if I can work on getting the volume fractions from the previous cycle in here and then requiring that they be well-defined.

singularity-eos/eos/get_sg_eos_functors.hpp
```diff
@@ -71,31 +73,89 @@ struct init_functor {
...
     return;
   }
+
+ private:
+  // Check to make sure returned values are normal or zero
+  template <typename valT>
+  PORTABLE_FORCEINLINE_FUNCTION int bad_val(valT const value,
+                                            const char *const var_name) const {
+    const bool good =
+        (value == valT{0}) || (std::isnormal(1e10 * value) && std::isnormal(value));
```
mauneyc-LANL, line 117: Is `valT` expected to always be floating point? That can be enforced, but I'd be okay with a comment.

singularity-eos/eos/get_sg_eos_functors.hpp
```diff
@@ -71,31 +73,89 @@ struct init_functor {
...
       press_pte(tid, mp) = press_v(i) * p_mult;
       sie_pte(tid, mp) = sie_v(i) * frac_mass_v(i, m) * s_mult;
     }
     return;
   }
+
+ private:
+  // Check to make sure returned values are normal or zero
+  template <typename valT>
+  PORTABLE_FORCEINLINE_FUNCTION int bad_val(valT const value,
```
mauneyc-LANL, line 114: At a glance, this function can be `constexpr` (and similar `constexpr` variables and `if constexpr`)

singularity-eos/eos/get_sg_eos_functors.hpp
```diff
@@ -216,12 +277,144 @@ struct final_functor {
...
     for (int m = 0; m < nmat; ++m) {
       frac_mass_v(i, m) *= mass_sum;
     }
     return;
   }
+
+ private:
+  // Check to make sure returned values are normal or zero
+  template <typename valT>
+  PORTABLE_FORCEINLINE_FUNCTION int bad_val(valT const value,
```
mauneyc-LANL, line 291: What is the purpose of duplicating this function across objects? If it's common then it should be moved up scope and have a single definition.

singularity-eos/eos/get_sg_eos_functors.hpp
```diff
@@ -216,12 +277,144 @@ struct final_functor {
...
+    }
+    return negative;
+  }
+  // Only run these checks when built with debug flags
+  PORTABLE_FORCEINLINE_FUNCTION
+  void check_all_vals(int const i, int const npte, int const tid,
+                      bool const pte_converged) const {
+#ifndef NDEBUG
+    int n_bad_vals = 0;
+    for (int mp = 0; mp < npte; ++mp) {
```
mauneyc-LANL, line 332: This is a broader style point than just these lines; I'm a fan of AAA - Almost Always Auto. `auto n_bad_vals = 0;` , `for (auto mp; mp < npte, ++mp)`, so on. 
If there isn't type restrictions due to passing values to Fortran (where this style is probably bad), I'd consider AAA as the preferred C++ style.

singularity-eos/eos/get_sg_eos_functors.hpp
```diff
@@ -216,12 +277,144 @@ struct final_functor {
...
+  }
+  // Check for prrecluding zero and negative (non-positive) values
+  template <typename valT>
+  PORTABLE_FORCEINLINE_FUNCTION int non_positive_val(valT const value,
+                                                     const char *const var_name) const {
+    const bool not_positive = value <= valT{0};
+    if (not_positive) {
+      using PortsOfCall::printf;
+      printf("### ERROR: Bad singularity-eos output\n  Var: %s\n  Value: %e\n", var_name,
+             value);
```
mauneyc-LANL, line 310: Having the `using` just to bring `printf` into scope is a bit clunky. Either have `using PortsOfCall::printf;` at object or namespace scope, or just directly call `PortsOfCall::printf( /*...*/ );`

singularity-eos/eos/get_sg_eos_functors.hpp
```diff
@@ -216,12 +277,144 @@ struct final_functor {
...
+    return not_positive;
+  }
+  // Check for prrecluding negative values
+  template <typename valT>
+  PORTABLE_FORCEINLINE_FUNCTION int negative_val(valT const value,
+                                                 const char *const var_name) const {
+    const bool negative = value < valT{0};
+    if (negative) {
+      using PortsOfCall::printf;
+      printf("### ERROR: Bad singularity-eos output\n  Var: %s\n  Value: %e\n", var_name,
```
mauneyc-LANL, line 321: Not to hammer on this point too much, but reducing code duplication is a Good Thing. All these "check" functions can be simplified:
```c++
// let's start using c++17! std::string_view and all
// if we roll a constexpr `isnormal` function, we replace the lambda with a constexpr eval
template<class Cond, class VT>
PORTABLE_FORCEINLINE_FUNCTION auto check_val (Cond&& C, VT&& value, std::string_view vname) const
{
  auto pass = Cond(std::forward<VT>(value));
  if (Cond(std::forward(value))) {
    PortsOfCall::printf("%s %e\n", var_name.data(), value);
  }
  return pass;
}

auto negative_val = [](auto value, auto name) { return check_val ([](auto v) { return v < 0; }, name); };
// and so on;
```
(spitballing a bit, probably a cleaner approach).

mauneyc-LANL (approved):
Approve, with comments.

singularity-eos/eos/get_sg_eos_functors.hpp
mauneyc-LANL: Ah, `std::isnormal` needs another C++ iteration to be `constexpr`. Could roll our own `constexpr` equivalent.

singularity-eos/eos/get_sg_eos_functors.hpp
```diff
@@ -71,31 +73,89 @@ struct init_functor {
     for (int m = 0; m < nmat; ++m) {
       frac_mass_v(i, m) /= mass_sum;
     }
-    /* set inputs */
+    check_all_vals(i);
+    // count the number of participating materials and zero the inputs
     npte = 0;
     for (int m = 0; m < nmat; ++m) {
       if (frac_mass_v(i, m) > 1.e-12) {
```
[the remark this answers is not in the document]
jhp-lanl: Yes! In fact I want to pass that as a function parameter. I had forgotten that I wanted to do this, but I can certainly address that here

singularity-eos/eos/get_sg_eos_functors.hpp
jhp-lanl: I suppose `isnormal` doesn't really make sense for anything that isn't floating point, right? I can static assert here or I can SFINAE it away, whichever you prefer

singularity-eos/eos/get_sg_eos_functors.hpp
jhp-lanl: The print message is different between the two definitions in the two functors. I can probably generalize it better though and lift the functionality up in scope

singularity-eos/eos/get_sg_eos_functors.hpp
jhp-lanl: I'm not sure I'm 100% bought in to AAA, but I can see how it could be helpful in these two places. The index is pretty easy to reason about since it is constrained either by the indexing behavior of the containers or the mapping `pte_mats()`. I don't anticipate the compiler doing anything weird with the type here.

For `n_bad_vals`, I suppose the code works with both a boolean type or an integer type so I'm being too constrictive. I was just worried that maybe there was some aspect of boolean math I didn't understand that might bite me (for example if true doesn't necessarily have to be 1). So I specifically chose an `int` here so that it was easier to reason about what exactly was going to happen at each step (I did at least make sure that the `bool` -> `int` conversion was guaranteed to be defined).

If I'm being overly cautious though, I can be persuaded to change both to be auto.

singularity-eos/eos/get_sg_eos_functors.hpp
jhp-lanl: I was intentionally limiting the scope that it was used in, but I suppose there's no reason to limit the scope below the object scope. I'll move that

singularity-eos/eos/get_sg_eos_functors.hpp
jhp-lanl: I'll roll those into a single function. I just got lazy and added those negative checks at the end.

I'm also still in C++ 14 mode so I haven't started using string views.

singularity-eos/eos/get_sg_eos_functors.hpp
mauneyc-LANL: I don't want to hold things up, I just get flustered seeing duplicated code. It can be changed later.

singularity-eos/eos/get_sg_eos_functors.hpp
mauneyc-LANL: the reference says `isnormal` has integer-type overloads, with the confusing message `which are treated as double`. I think `static_assert` is fine.

singularity-eos/eos/get_sg_eos_functors.hpp
Yurlungur, line 114: Is there a reason for `bad_val` to be an integer, rather than a boolean? Also should this be more general and moved into, e.g., our robustness utils header?

singularity-eos/eos/get_sg_eos_functors.hpp
```diff
@@ -71,31 +73,89 @@ struct init_functor {
...
   }
+
+ private:
+  // Check to make sure returned values are normal or zero
+  template <typename valT>
+  PORTABLE_FORCEINLINE_FUNCTION int bad_val(valT const value,
+                                            const char *const var_name) const {
+    const bool good =
+        (value == valT{0}) || (std::isnormal(1e10 * value) && std::isnormal(value));
+    if (not good) {
```
Yurlungur, line 118: I didn't realize `not` was valid C++ but I would prefer more canonical C-like style
```suggestion
    if (!good) {
```

singularity-eos/eos/get_sg_eos_functors.hpp
```diff
@@ -71,31 +73,89 @@ struct init_functor {
...
+      using PortsOfCall::printf;
+      printf("### ERROR: Bad singularity-eos input\n  Var: %s\n  Value: %e\n", var_name,
+             value);
+    }
+    return not good;
```
Yurlungur, line 123:
```suggestion
    return !good;
```

singularity-eos/eos/get_sg_eos_functors.hpp
```diff
@@ -71,31 +73,89 @@ struct init_functor {
...
+  void check_all_vals(int const i) const {
+#ifndef NDEBUG
+    int n_bad_vals = 0;
+
+    for (int m = 0; m < nmat; ++m) {
+      n_bad_vals += bad_val(frac_mass_v(i, m), "frac_mass");
+    }
+    n_bad_vals += bad_val(press_v(i), "pres");
+    n_bad_vals += bad_val(sie_v(i), "sie");
+    n_bad_vals += bad_val(spvol_v(i), "spvol");
```
Yurlungur, line 136: Ah I see. This is why you use `int` instead of `bool` in the return type. `bool` will actually work here though, as the compiler will automatically properly cast. I would prefer `bad_val` as a `bool` still I think if possible.

singularity-eos/eos/get_sg_eos_functors.hpp
```diff
@@ -71,31 +73,89 @@ struct init_functor {
...
+
+    if (n_bad_vals > 0) {
```
Yurlungur, line 138: In fact, since you're just checking for *one* bad value, `bad_val` could return a boolean and then `n_bad_vals` could be `any_bad_vals = any_bad_vals || bad_val`

singularity-eos/eos/get_sg_eos_functors.hpp
Yurlungur: I would also prefer this be more generic. IMO something like this should live in a utility header.

singularity-eos/eos/get_sg_eos_functors.hpp
Yurlungur, line 310: Do we have a `PORTABLE_WARN` method? If so this would be an optimal use-case. Also I think these check functions belong in global scope in a utility header.

singularity-eos/eos/get_sg_eos_functors.hpp
Yurlungur: I don't think string views are necessary but I agree with @mauneyc-LANL that these checks should be unified and maybe moved to shared scope.

singularity-eos/eos/get_sg_eos_functors.hpp
Yurlungur: I'm going to push back on @mauneyc-LANL here. Types are a form of documentation IMO and I lean *against* auto unless it explicitly generalizes the code.

Regarding bools vs ints I actually think all this logic about bad values should be boolean. I don't see why we need ints at all here. I make a comment elsewhere in the review describing what I would do as far as that is concerned.

singularity-eos/eos/get_sg_eos_functors.hpp
Yurlungur, line 117: Is 1e10 the right number to multiply by here? Might want to make that a global named constant also.

Yurlungur (changes requested):
Overall this is a nice (and necessary) improvement. I have some nitpicks below, though.

singularity-eos/eos/get_sg_eos_functors.hpp
jhp-lanl: Interesting... good to know!

singularity-eos/eos/get_sg_eos_functors.hpp
jhp-lanl: No reason in particular. I might follow @mauneyc-LANL 's advice and just use `auto` unless there's a good reason not to

singularity-eos/eos/get_sg_eos_functors.hpp
jhp-lanl: I can make it a named constant. I probably won't make it a function argument like the mass fraction cutoff though

singularity-eos/eos/get_sg_eos_functors.hpp
jhp-lanl: I'm resistant to using the portable warn for two reasons: 1) this isn't a warning since it's actually impetus for the error and 2) I just didn't want to deal with lifting the string concatenation out of the base class into a utility header.

Now that I need to generate more utility functions though, it's probably worth it to just put the string concatenation in there too. But I'm still resistant to labeling these as warnings.

To illustrate my point, this is a sample output that I think is pretty effective:
```
### ERROR: Bad singularity-eos output
  Var: frac_ie
  Value: 9.656546e-309
### ERROR: Bad singularity-eos output
  Var: temp
  Value: 3.474671e-313
### Bad Value Output state:
  ~~Bulk state~~
    Pressure      :    3.084860764090054e+12
    Temperature   :    3.47467079861110e-313
    Specific IE   :    2.789552895739220e+09
    Bulk Modulus  :    1.127154852753680e+13
    Heat Capacity :    2.937634386418220e+08
    dPdE          :        0.611123943858706
  ~~Material States~~
    mat:            Mass fraction              Material IE          Material Volume
      1:     4.90001520366884e-05   9.656546035209074e-309      0.00053771637633576
      2:        0.999566645882337    7.229168124727389e+08     0.000283501725784361
      4:     0.000384353965626437   -9.152107585263315e+04     6.39064862153302e-06
  ~~PTE iteration *DID NOT* converge~~
### ERROR
  Message:     Bad value returned from singularity-eos interface. See output for details
  File:        /usr/projects/eap/users/jhp/XCAP/singularity-eos/github/singularity-eos/singularity-eos/eos/get_sg_eos_functors.hpp
  Line number: 413
```

I think if the `ERROR` messages said `WARNING`, it might falsely convey that the values aren't the _cause_ of the eventual error.

singularity-eos/eos/get_sg_eos_functors.hpp
jhp-lanl: My primary concern was clarity and I thought that using an integer to count the bad values, but I think the way you've written this with booleans it's also pretty clear.

singularity-eos/eos/get_sg_eos_functors.hpp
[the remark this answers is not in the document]
jhp-lanl: Added as an optional argument at the fortran interface level. Defaults to 1e-12

singularity-eos/eos/get_sg_eos_functors.hpp
jhp-lanl: Added static assert an pass by value

singularity-eos/eos/get_sg_eos_functors.hpp
jhp-lanl: Added a utility header for error utilities. I also followed @mauneyc-LANL 's suggestion and broke out the condition from the message.

I had originally wanted the value checks to print out different things if they were in the input versus output, but I realize that I'm covering that in my error message so the checks themselves can be generic.

singularity-eos/eos/get_sg_eos_functors.hpp
jhp-lanl: I think I split the difference here. I made index variables auto since their type is easily deduced from the Kokkos views or the iteration extent.

I want to keep the type of the bad values counter (now a `bool` and `any_bad_vals`) since I feel that it communicates a lot about the behavior of that variable in the function.

singularity-eos/eos/get_sg_eos_functors.hpp
jhp-lanl: This now lives at the namespace level of the error utilities

singularity-eos/eos/get_sg_eos_functors.hpp
jhp-lanl: I generalized this into the error utilities header

singularity-eos/eos/get_sg_eos_functors.hpp
jhp-lanl: Changed to `bool` for clarity of communication

singularity-eos/eos/get_sg_eos_functors.hpp
jhp-lanl: Changed

singularity-eos/eos/get_sg_eos_functors.hpp
jhp-lanl: As I was writing this, I realize that this kind of takes the role of an error information function. If what I wrote is useful, it can be moved to `ports_of_call` later.

jhp-lanl:
@jonahm-LANL or @mauneyc-LANL any ideas on what I'm doing wrong here?
```
/home/runner/work/singularity-eos/singularity-eos/singularity-eos/eos/singularity_eos.f90:410: undefined reference to `get_sg_eos'
```

I added the `mass_frac_cutoff` argument to the `get_sg_eos()` interface function both on the C++ side and the fortran side, and I'm adding an optional argument to the fortran interface. I can't for the life of me see where my mistake is and why the linker wouldn't be able to see `get_sg_eos()`. Any help is appreciated!

jhp-lanl:
> @jonahm-LANL or @mauneyc-LANL any ideas on what I'm doing wrong here?
> 
> ```
> /home/runner/work/singularity-eos/singularity-eos/singularity-eos/eos/singularity_eos.f90:410: undefined reference to `get_sg_eos'
> ```
> 
> I added the `mass_frac_cutoff` argument to the `get_sg_eos()` interface function both on the C++ side and the fortran side, and I'm adding an optional argument to the fortran interface. I can't for the life of me see where my mistake is and why the linker wouldn't be able to see `get_sg_eos()`. Any help is appreciated!

@dholladay00 with the fix. I forgot that the `get_sg_eos()` prototype exists in the `singularity_eos.hpp` header even though the function itself is defined the `get_sg_eos.cpp` file.

utils/scripts/format.sh
```diff
@@ -39,6 +39,6 @@ for f in $(git grep --untracked -ail res -- :/*.hpp :/*.cpp); do
     if [ ${VERBOSE} -ge 1 ]; then
        echo ${f}
     fi
-    ${CFM} -i ${REPO}/${f}
+    ${CFM} -style=file -i ${REPO}/${f}
```
jhp-lanl, line 42: FYI - this change just brings the format script in line with what the CI is doing. I don't think its omission was _really_ messing anything up, but I just made the change as part of trying to figure out why my clang format wasn't working.

jhp-lanl:
@mauneyc-LANL and @Yurlungur let me know if the most recent changes have addressed your comments. I think the new `error_utils.hpp` file needs a fresh review

singularity-eos/base/error_utils.hpp
```diff
@@ -0,0 +1,88 @@
...
+constexpr double _NORMAL_FACTOR = 1.0e10;
+
+struct is_normal_or_zero {
+  template <typename valT>
+  bool PORTABLE_FORCEINLINE_FUNCTION operator()(valT value) const {
+    static_assert(std::is_floating_point_v<valT>);
+    return (value == valT{0}) ||
+           (std::isnormal(_NORMAL_FACTOR * value) && std::isnormal(value));
+  }
+};
```
Yurlungur, line 37: These as structs is fine. But why structs instead of just functions?

Yurlungur (approved):
I'm happy now. Just one question below, but it is non-blocking.

singularity-eos/base/error_utils.hpp
jhp-lanl: Stylistically I like functors better than functions because I find it easier to reason about the type. I could have used lambdas here, but I find structs with templates more readable even if they're a bit more verbose.

I also thought it would be more extensible to be explicit that zero be the same type as the template parameter rather than letting the compiler compare to an `int` literal or maybe interpreting zero as a different type.

jhp-lanl: The type of zero thing isn't really that important, but you might imagine a future at some point where the types of the values are unit aware somehow and this explicitness might be favorable.

If you'd like, I can wrap these in an `impl` namespace to signal that they're implementation details.

jhp-lanl: Actually, these might be useful on their own. I'll leave them as-is if that's okay.

jhp-lanl:
I'm having trouble with one machine on our gitlab CI (network cloning issues... possibly proxy problems). Can one of you run the CI for me there? I think this is good to go otherwise.

singularity-eos/base/error_utils.hpp
Yurlungur: Yeah it's fine to leave them.

Yurlungur:
> I'm having trouble with one machine on our gitlab CI (network cloning issues... possibly proxy problems). Can one of you run the CI for me there? I think this is good to go otherwise.

Looks like it's passing on gitlab. Going to pull the trigger.

PR #383, opened 2024-06-27
Fix Bug in PTE Solver and Several Code Quality Improvements

jhp-lanl:
## PR Summary

This PR is dependent on https://github.com/lanl/singularity-eos/pull/382

This PR makes a few main changes:
- ~~The previous `ScaleDx` algorithm was written assuming that only one guard (small volume fraction, small temperature, minimum density, etc.) would activate at once. I believe I'm running into a test case where multiple guards are activating so they end up overwriting each other.~~
  Actually, since the newly-shrunk `scale` is used in each successive `if` block, they have the effect of only decreasing scale. I'll add a comment to this effect.
- Materials that didn't participate in the PTE solver had their energies set to zero with the volumes untouched. This PR changes this so that non-participating materials are assumed to be at the cell density, which determines their densities and energies. This also slightly changes the volume available for the PTE materials, reflected in a different volume fraction sum.
- I also added some comments to the line search algorithm and simplified the logic slightly by removing what appeared to be a superfluous `else` block.
- Fixed two bugs in the the rho-T PTE solver `ScaleDx` function. First, there were missing parentheses. Second, the logic was incorrect when the volume fraction of a material has already exceeded the imposed maximum. This maximum occurs at the minimum pressure on the EOS, effectively guarding against the EOS entering unphysical (thermodynamically unstable) states. However, it's possible for this maximum to change as the temperature guess changes such that an old volume fraction is above this maximum. In that case, the logic of the function was _increasing_ the step to get back into the physical regime, but possibly at the cost of violating other checks.
- Adds a test for a difficult PTE state. It was failing to converge but now succeeds.
- Improves descriptiveness of assert statements to identify the function they are in. This isn't strictly necessary since the line number is given, but it makes debugging a bit faster.
- The `TestUpdate()` function has been changed so that it's a bit more clear when it's also updating the state versus when it's supposed to cache the previous state.

## PR Checklist

- N/A Update the version in cmake.
- N/A Move the changes in the CHANGELOG.md file under a new header for the new release, and reset the categories.


jhp-lanl:
I'm still getting negative volume fractions despite the fact that I believe the code should be catching this. I'll add a unit test here shortly that should demonstrate this and then work on whatever fix is needed to prevent this.

singularity-eos/closure/mixed_cell_models.hpp
```diff
@@ -684,8 +691,15 @@ class PTESolverRhoT : public mix_impl::PTESolverBase<EOSIndexer, RealIndexer> {
...
+      if (alpha_max < vfrac[m]) {
+        // Despite our best efforts, we're already in the unstable regime (i.e.
+        // dPdV_T > 0) so we would actually want to *increase* the step instead
+        // of decreasing it. As a result, this code doesn't work as intended and
+        // should be skipped.
+        continue;
+      }
       if (scale * dx[m] > 0.5 * (alpha_max - vfrac[m])) {
-        scale = robust::ratio(0.5 * alpha_max - vfrac[m], dx[m]);
+        scale = robust::ratio(0.5 * (alpha_max - vfrac[m]), dx[m]);
```
jhp-lanl, line 702: There were two bugs here:
1. If the volume fraction is already greater than this maximum and the current step isn't large enough to bring it back, we need a _larger_ step, not a smaller step. This goes against the logic of this function in limiting the step and could undo other limits. As a result, I just give up on this limit if we're already beyond it.
2. There was a missing set of parentheses in the eventual scale update.

Yurlungur: Nice catch!

jhp-lanl:
@Yurlungur and @dholladay00 this is ready for review now.

One comment: my test is aborting due to a `double free or corruption (out)` and I'm not quite sure why. I tried removing the `PORTABLE_FREE` statements, but that didn't change anything.

The test assertion passes, but this abort causes the test to fail. Any guidance on how to debug this?

test/test_closure_pte.cpp
```diff
@@ -0,0 +1,134 @@
...
+            v_temperatures, v_pressures, lambdas, scratch);
+
+        // Solve the PTE system and ensure it converged
+        bool pte_converged = PTESolver(method);
+        CHECK(pte_converged);
+        PORTABLE_FREE(lambdas);
+      }
+    }
+    // Free EOS memory
+    PORTABLE_FREE(v_EOS);
```
Yurlungur, line 128: It looks like you're not calling `Finalize` for the EOS models? Or `PORTABLE_FREE` for `scratch?

Yurlungur (approved):
Looks good! Thankis for chasing these down. I'm somewhat concerned that the PTE solver needs to run whether or not something goes wrong, so I worry about about these `PORTABLE_FAIL` checks. I'd prefer warnings and returning a failure status. 

Regarding debugging the double free---it might be some interaction between ctest and the test code. Maybe pull the test code out into a stand alone executable? You could also try running it through valgrind on CPU.

test/test_closure_pte.cpp
jhp-lanl: `PORTABLE_FREE` for scratch caused an error that prevents the test from completing for some reason. Good catch on calling `Finalize()` though. Maybe that's part of it...

jhp-lanl:
>I'm somewhat concerned that the PTE solver needs to run whether or not something goes wrong, so I worry about about these PORTABLE_FAIL checks. I'd prefer warnings and returning a failure status.

Are you saying you'd prefer the PTE solver to fail at the point where an issue is reached rather than running and then returning a nonsense result?

Normally I'd agree with you, but in my experience of debugging the PTE solver, I found the indirection and indexing to be difficult to navigate in the debugger, hence creating the print statements for when the PTE solver fails. I found it much easier to extract the bad state, reproduce it in a unit test, and then debug from there rather than debugging inside the host code.

That said, maybe there's a middle ground? I could add a compile time option instead of just relying on debug mode. That way I could compile in release mode with the option to find the bad state and then compile in debug to actually find the point of failure in the PTE solve. What are your thoughts?

Yurlungur:
> > I'm somewhat concerned that the PTE solver needs to run whether or not something goes wrong, so I worry about about these PORTABLE_FAIL checks. I'd prefer warnings and returning a failure status.
> 
> Are you saying you'd prefer the PTE solver to fail at the point where an issue is reached rather than running and then returning a nonsense result?

Actually the opposite. ;) I was hoping for the PTE solver to continue running even in debug mode. In case the failure state is something the host code needs to be robust for.

> Normally I'd agree with you, but in my experience of debugging the PTE solver, I found the indirection and indexing to be difficult to navigate in the debugger, hence creating the print statements for when the PTE solver fails. I found it much easier to extract the bad state, reproduce it in a unit test, and then debug from there rather than debugging inside the host code.

That makes complete sense. I have to admit I'm concerned by how difficult this was to navigate.

> That said, maybe there's a middle ground? I could add a compile time option instead of just relying on debug mode. That way I could compile in release mode with the option to find the bad state and then compile in debug to actually find the point of failure in the PTE solve. What are your thoughts?

That makes sense to me!

jhp-lanl:
> > > I'm somewhat concerned that the PTE solver needs to run whether or not something goes wrong, so I worry about about these PORTABLE_FAIL checks. I'd prefer warnings and returning a failure status.
> > 
> > 
> > Are you saying you'd prefer the PTE solver to fail at the point where an issue is reached rather than running and then returning a nonsense result?
> 
> Actually the opposite. ;) I was hoping for the PTE solver to continue running even in debug mode. In case the failure state is something the host code needs to be robust for.

Ah okay I think we're on the same page then. I'll limit the assert failures to malformed inputs and rely on the checks in `get_sg*`. If we want, we can probably adapt those checks in the future to be an extra layer on the outputs of the PTE solver itself. I'll make an issue for that.

jhp-lanl:
I added a `Finalize()` for each EOS as well as freeing all the memory but I get this failure:
```
double free or corruption (!prev)

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
closure_unit_tests is a Catch2 v3.0.1 host application.
Run with -? for options

-------------------------------------------------------------------------------
Scenario: Density-Temperature PTE Solver
      Given: Equations of state
      Given: A difficult state for the density-temperature PTE solver
       Then: The PTE solver should converge
-------------------------------------------------------------------------------
/usr/projects/eap/users/jhp/XCAP/singularity-eos/github/singularity-eos/test/test_closure_pte.cpp:112
...............................................................................

/usr/projects/eap/users/jhp/XCAP/singularity-eos/github/singularity-eos/test/test_closure_pte.cpp:112: FAILED:
  {Unknown expression after the reported line}
due to a fatal error condition:
  SIGABRT - Abort (abnormal termination) signal

===============================================================================
test cases: 1 | 1 failed
assertions: 2 | 1 passed | 1 failed

Aborted
```
This is true whether the test is run in isolation or as part of the suite or as a standalone executable.

jhp-lanl:
I think I figured out the issue. I forgot that `PORTABLE_MALLOC` has no concept of the type being allocated and needs the size of the type.

test/test_closure_pte.cpp
jhp-lanl: Added the appropriate `free` statements and realized my mistake: I wasn't allocating enough memory.

jhp-lanl (commented):
Please don't merge until the gitlab tests pass

test/CMakeLists.txt
```diff
@@ -62,18 +68,29 @@ endif()
...
 if(SINGULARITY_TEST_SESAME)
   target_compile_definitions(eos_tabulated_unit_tests PRIVATE SINGULARITY_TEST_SESAME)
+  target_compile_definitions(closure_unit_tests PRIVATE SINGULARITY_TEST_SESAME)
 endif()
 if(SINGULARITY_TEST_STELLAR_COLLAPSE)
   target_compile_definitions(eos_tabulated_unit_tests
                              PRIVATE SINGULARITY_TEST_STELLAR_COLLAPSE)
 endif()
+if(SINGULARITY_USE_SPINER)
+  target_compile_definitions(closure_unit_tests
```
dholladay00, line 78: Are these preprocessor defines specific to this test or are they defined elsewhere as well @jhp-lanl ?

test/test_closure_pte.cpp
```diff
@@ -0,0 +1,143 @@
...
+    constexpr Real n = 2.66182;
+    constexpr Real vc = 0.75419;              // cm^3 / g
+    constexpr Real pc = 3.2 * GPa;            // microbar
+    constexpr Real Cv = 0.001072 * MJ_per_kg; // erg / g / K
+    constexpr Real Q = 4.115 * MJ_per_kg;
+    eos_arr[2] = DavisProducts(a, b, k, n, vc, pc, Cv).Modify<ShiftedEOS>(-Q);
+
+    // Move EOS array from host to device
+    for (auto eos : eos_arr) {
+      eos.GetOnDevice();
```
dholladay00, line 68: Recall that get on device is *not* a void function. It returns a new eos that needs to be assigned @jhp-lanl

test/test_closure_pte.cpp
```diff
@@ -0,0 +1,143 @@
...
+      Real *v_pressures = (Real *)PORTABLE_MALLOC(bytes);
+      portableCopyToDevice(v_pressures, pressures.data(), bytes);
+
+      THEN("The PTE solver should converge") {
+        // Allocate scratch space for the PTE solver
+        const int pte_solver_scratch_size = PTESolverRhoTRequiredScratch(num_eos);
+        const size_t scratch_bytes = pte_solver_scratch_size * sizeof(Real);
+        double *scratch = (double *)PORTABLE_MALLOC(scratch_bytes);
+        constexpr size_t lambda_bytes = MAX_NUM_LAMBDAS * num_eos * sizeof(Real);
+        double **lambdas = (double **)PORTABLE_MALLOC(lambda_bytes);
```
dholladay00, line 118: doesn't quite look right here. This is roughly creating an array of pointers of length MAX_NUM_LAMBDAS * num_eos, though not quite b/c the underlying type should be Real*. Creating the equivalent of a matrix with **'s is a somewhat complex procedure. First one mallocs an array of pointers (the ** quantity), then for each pointer, one does a malloc of the actual underlying type. And the frees need to handled in the same way.

jhp-lanl: Ah fair enough. I was trying to do it in one quick allocation, but I also need to populate the pointers. I'll change this.

test/test_closure_pte.cpp
jhp-lanl: I thought maybe the copy I'm doing later would be equivalent, but I suppose I need to return the new EOS on host and then copy that to device rather than copying the old EOS object to device.

jhp-lanl: I skipped the `portableCopyToDevice` and instead just put the output of `GetOnDevice()` in the device array. Is this what you meant?

test/test_closure_pte.cpp
jhp-lanl: Let me know if I'm doing this right now

jhp-lanl:
@dholladay00 I modified the tests to address your comments. I also broke out some code into functions so I can start a collection of PTE states that fail to converge. I have my next state in there already, but I'll back burner it for now and create an issue so we can get this merged in.

I'm worried that I might go down a rabbit hole of PTE issues if I keep running until I don't get any issues. I'll probably scale back the ambitions of my EAP MR and provide simpler tests of the functionality.

test/CMakeLists.txt
jhp-lanl: I went ahead and switched to `SINGULARITY_USE_SPINER_WITH_HDF5`. Let me know if you're satisfied here.

test/test_closure_pte.cpp
```diff
@@ -0,0 +1,214 @@
...
+  // Set up the PTE solver
+  PTESolverRhoT<decltype(v_EOS), Real *, Real **> method(
+      num_eos, v_EOS, vfrac_sum, sie_bulk, v_densities, v_vol_frac, v_sies,
+      v_temperatures, v_pressures, lambdas, scratch);
+
+  // Solve the PTE system
+  bool pte_converged = PTESolver(method);
+
+  // Free temp memory
+  PORTABLE_FREE(lambdas);
```
dholladay00, line 119: each lambda needs to be freed.

dholladay00:
@jhp-lanl we have gone back and forth on the portability of the memory, but now that I look more closely this test contains no device kernels, it is all host. Either we need to run the solve in a kernel (could do a length 1 portable for) _or_ make the memory host only.

test/test_closure_pte.cpp
jhp-lanl: Okay so keep the output of `GetOnDevice()` in the host array and then copy that array to device? I'll change it

test/test_closure_pte.cpp
jhp-lanl: Good catch!

test/test_closure_pte.cpp
jhp-lanl: Alright this should be fixed now

jhp-lanl:
The PTE test should be running on device now. I'll check the gitlab CI here in a bit to make sure it's running correctly (might need somebody to run the Roci runners).

spack-repo/packages/singularity-eos/package.py
```diff
@@ -112,6 +112,7 @@ class SingularityEos(CMakePackage, CudaPackage):
 
     depends_on("ports-of-call@1.4.2,1.5.2:", when="@:1.7.0")
     depends_on("ports-of-call@1.5.2:", when="@1.7.1:")
+    depends_on("ports-of-call@1.6.0:", when="@1.8.1:")
```
jhp-lanl, line 115: @rbberger do you have a preference on a different way to do this? I.e. how do I identify a new dependency in a version that hasn't been released yet?

Or should we handle these at the release and just set them to `when="@main"` for now?

test/test_closure_pte.cpp
jhp-lanl: I redid this again after a discussion with @dholladay00 about needing to call `Finalize()` on both the original object and the one that gets created when `GetOnDevice()` is called.

test/CMakeLists.txt
jhp-lanl: Going to resolve this

dholladay00: did you remove this compile definition then as it seems to still be showing up in the cmake diff.

jhp-lanl: Did I remove it? No. I didn't think it needed to be removed from your previous comment. This test shouldn't be run when `SINGULARITY_USE_SPINER_WITH_HDF5` is off.

I did forget to switch it to `SINGULARITY_USE_SPINER_WITH_HDF5` though. That has been corrected.

spack-repo/packages/singularity-eos/package.py
jhp-lanl: @cmauney @Yurlungur @dholladay00 do you have preferences here?

jhp-lanl:
I think I got all the bugs worked out with running the test on device.

The test should be a good platform to add tests that used to break the PTE solver (or still do while we're figuring out a solution). And it can be a bit of a template for using the new `ports-of-call` features for allocating on device without having to mess with lots of Kokkos views and `#IFDEF` statements.

test/CMakeLists.txt
dholladay00: The following macro should already be defined https://github.com/lanl/singularity-eos/blob/main/cmake/singularity-eos/hdf5.cmake#L92C21-L92C67 `SINGULARITY_USE_HDF5`

test/CMakeLists.txt
```diff
@@ -62,18 +68,29 @@ endif()
...
   target_compile_definitions(eos_tabulated_unit_tests
                              PRIVATE SINGULARITY_TEST_STELLAR_COLLAPSE)
 endif()
+if(SINGULARITY_USE_SPINER_WITH_HDF5)
+  target_compile_definitions(closure_unit_tests
+                             PRIVATE SINGULARITY_USE_SPINER_WITH_HDF5)
+endif()
+if(SINGULARITY_BUILD_CLOSURE)
+  target_compile_definitions(closure_unit_tests
+                             PRIVATE SINGULARITY_BUILD_CLOSURE)
```
dholladay00, line 83: this should already be defined https://github.com/lanl/singularity-eos/blob/main/CMakeLists.txt#L266

jhp-lanl: I'm confused here. Isn't this statement also true for all of the code above this that isn't part of this MR? It seems by this logic we don't need any of the `target_compile_definitions` here at all. If that's the case, I'd prefer this be a separate issue since it goes beyond the code I'm adding here.

dholladay00: I'm not sure that they are defined elsewhere so I think some of those definitions are necessary additions. Even if they were duplicated (which from what I can tell they _aren't_), I still think this code should be done the right way and I don't see how the correctness or incorrectness of the code above would result in these requested changes to _your_ change-set outside of the scope of the MR.

test/CMakeLists.txt
jhp-lanl: I prefer using the `SINGULARITY_USE_SPINER_WITH_HDF5` flag since I'm specifically testing for spiner and HDF5 not just the latter.

dholladay00:
great work @jhp-lanl! This not only fixes a long-standing bug in the solver, but it uses the latest best practices for portability and adds improvements to the configure/build system. In addition, this provides a tool to explore problematic PTE inputs separate from the context of a host code! 

Assuming this passes CI and pending review by those that understand the physics better than I, this is ready! 👍

spack-repo/packages/singularity-eos/package.py
dholladay00: @rbberger any thoughts?

test/CMakeLists.txt
jhp-lanl: Ah I understand now and have adopted the following convention:
1. any `SINGULARITY_TEST_*` variables are defined in the CMakeLists.txt file for the tests.
2. any other `SINGULARITY_USE_*` or `SINGULARITY_BUILD_*` variables should be defined in the main project CMakeLists.txt and the only one missing was `SINGULARITY_USE_SPINER_WITH_HDF5`. I've added a compile definition there and I think everything works now.

jhp-lanl:
> Assuming this passes CI and pending review by those that understand the physics better than I, this is ready! 👍

I think I'm still having some issues with the CI on gitlab unfortunately. I _thought_ I had fixed these issues on my build when executing on device, but I'll keep working to reproduce

singularity-eos/closure/mixed_cell_models.hpp
```diff
@@ -1417,25 +1457,26 @@ PORTABLE_INLINE_FUNCTION bool PTESolver(System &s) {
 
     // possibly scale the update to stay within reasonable bounds
     Real scale = s.ScaleDx();
-    // const Real scale_save = scale;
+    PORTABLE_REQUIRE(scale <= 1.0, "PTE Solver is attempting to increase the step size");
```
Yurlungur, line 1460: when can this be reached?

Yurlungur (approved):
Thanks for this, @jhp-lanl having unit tests here is a big improvement. As is the nasty bugs you caught. I still want @pdmullen and @jdolence 's eyes on this. But I'm happy with this. I also ran this branch through riot and ran some strenuous test problems, which still look good.

dholladay00:
> Thanks for this, @jhp-lanl having unit tests here is a big improvement. As is the nasty bugs you caught. I still want @pdmullen and @jdolence 's eyes on this. But I'm happy with this. I also ran this branch through riot and ran some strenuous test problems, which still look good.

@Yurlungur: Jeff (@jhp-lanl) and I discussed the timeline for this PR. Given the timeline for a release into EAP, we are planning on merging this no later than EOD Monday and beginning the release process immediately following. This gives some, but not much time for review but we really can't wait any longer than that. If fixes are needed beyond what is caught on Monday, they will need to go in after the release.

Yurlungur:
> > Thanks for this, @jhp-lanl having unit tests here is a big improvement. As is the nasty bugs you caught. I still want @pdmullen and @jdolence 's eyes on this. But I'm happy with this. I also ran this branch through riot and ran some strenuous test problems, which still look good.
> 
> @Yurlungur: Jeff (@jhp-lanl) and I discussed the timeline for this PR. Given the timeline for a release into EAP, we are planning on merging this no later than EOD Monday and beginning the release process immediately following. This gives some, but not much time for review but we really can't wait any longer than that. If fixes are needed beyond what is caught on Monday, they will need to go in after the release.

Yes understood. I believe @pdmullen at least plans to review Monday. We shouldn't wait for @jdolence .

dholladay00:
> > > Thanks for this, @jhp-lanl having unit tests here is a big improvement. As is the nasty bugs you caught. I still want @pdmullen and @jdolence 's eyes on this. But I'm happy with this. I also ran this branch through riot and ran some strenuous test problems, which still look good.
> > 
> > 
> > @Yurlungur: Jeff (@jhp-lanl) and I discussed the timeline for this PR. Given the timeline for a release into EAP, we are planning on merging this no later than EOD Monday and beginning the release process immediately following. This gives some, but not much time for review but we really can't wait any longer than that. If fixes are needed beyond what is caught on Monday, they will need to go in after the release.
> 
> Yes understood. I believe @pdmullen at least plans to review Monday. We shouldn't wait for @jdolence .

@Yurlungur I'd really like @jdolence's eyes on it as some point for a sanity check if at all possible, even if it's after the merge. Sometimes the original author has knowledge that is hard to obtain by reading alone.

singularity-eos/closure/mixed_cell_models.hpp
```diff
@@ -219,8 +218,9 @@ class PTESolverBase {
     // material m averaged over the full PTE volume
     rho_total = 0.0;
     for (int m = 0; m < nmat; ++m) {
-      PORTABLE_REQUIRE(vfrac[m] > 0., "Non-positive volume fraction");
-      PORTABLE_REQUIRE(rho[m] > 0., "Non-positive density");
+      PORTABLE_REQUIRE(vfrac[m] > 0.,
+                       "Non-positive volume fraction provided to PTE solver");
+      PORTABLE_REQUIRE(rho[m] > 0., "Non-positive density provided to PTE solver");
```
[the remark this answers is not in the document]
jhp-lanl: I believe I specifically chose these because we divide by either these or related quantities at some point. In other words, I don't think the PTE solver is capable of handling zero mass/volume materials. I'll look here in a bit to confirm.

I think that's also fine since it's probably appropriate for an extra layer above this to handle non-participating materials. I'd also argue that our `Accessor` wrappers further motivate only sending a subset of materials into the PTE solver and dealing with zeros ahead of time.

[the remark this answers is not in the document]
jhp-lanl:
```c++
rho[m] = robust::ratio(rhobar[m], vfrac[m]);
```
```c++
sie[m] = uscale * robust::ratio(u[m], rhobar[m]);
```
`rhobar[m]` would be zero if `rho[m]` is zero.

Yes we have robust::ratio to handle these, but I'd worry that there could be unintended consequences for dividing by something close to zero. It just seems more appropriate to ask the host code (or an appropriate wrapper) to eliminate materials with zero density or volume fraction.

singularity-eos/closure/mixed_cell_models.hpp
```diff
@@ -353,12 +353,12 @@ class PTESolverBase {
       Asum += vfrac[m] * robust::ratio(press[m], temp[m]);
       rhoBsum += rho[m] * vfrac[m] * robust::ratio(sie[m], temp[m]);
     }
-    PORTABLE_REQUIRE(Tnorm > 0., "Non-positive temperature guess");
+    PORTABLE_REQUIRE(Tnorm > 0., "Non-positive Ideal PTE temperature guess");
     Asum *= uscale / Tnorm;
     rhoBsum /= Tnorm;
-    PORTABLE_REQUIRE(rhoBsum > 0., "Non-positive energy density");
+    PORTABLE_REQUIRE(rhoBsum > 0., "Non-positive Ideal PTE energy density guess");
```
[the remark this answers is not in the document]
jhp-lanl: The next line is
```c++
Tideal = uscale / rhoBsum / Tnorm;
```

This would imply a negative temperature guess, so no, this code cannot admit a negative energy density. I think this is probably part of why it's not useful outside of gas mixtures.

singularity-eos/closure/mixed_cell_models.hpp
jhp-lanl: This is an attempt to catch future bugs like the one I found. We had _thought_ that the code as written would only ever produce a smaller step, but when you were on the other side of the minimum pressure density, it would actually attempt to _increase_ the step in order to move the PTE solver back to a higher density.

I read somewhere that the appropriate role of asserts is to ensure your code is acting as you expect it to act (i.e. in release it's likely a waste to actually check these things). I think this is a perfect instance where an assert is needed; we believed the code acted a certain way and were proven wrong.

There's a plausible argument that the asserts I have elsewhere should actually be hard errors, but I hesitate to impose that performance cost at the library level and instead I think the host codes can make that decision themselves before passing data to the PTE solver.

jhp-lanl:
> > Thanks for this, @jhp-lanl having unit tests here is a big improvement. As is the nasty bugs you caught. I still want @pdmullen and @jdolence 's eyes on this. But I'm happy with this. I also ran this branch through riot and ran some strenuous test problems, which still look good.
> 
> @Yurlungur: Jeff (@jhp-lanl) and I discussed the timeline for this PR. Given the timeline for a release into EAP, we are planning on merging this no later than EOD Monday and beginning the release process immediately following. This gives some, but not much time for review but we really can't wait any longer than that. If fixes are needed beyond what is caught on Monday, they will need to go in after the release.

That said, I still need to fix the CI. I'll work on that this afternoon.

singularity-eos/closure/mixed_cell_models.hpp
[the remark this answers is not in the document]
Yurlungur: Ah ok. I just want to make sure this is only used for ideal gas, as tabulated EOSs definitely support negative energy.

singularity-eos/closure/mixed_cell_models.hpp
Yurlungur: Right if we believe this is impossible to reach, I think this should be an assert. My concern was that this might now happen naturally now that you've modified the line search to sometimes grow the line step size.

Yurlungur:
> @Yurlungur I'd really like @jdolence's eyes on it as some point for a sanity check if at all possible, even if it's after the merge. Sometimes the original author has knowledge that is hard to obtain by reading alone.

I agree. My concern is I don't know when I will be able to get him to look. We were visiting him in Ann Arbor this week and I asked him to look, though. So there's hope.

singularity-eos/closure/mixed_cell_models.hpp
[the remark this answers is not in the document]
Yurlungur: yeah ok fair enough.

jhp-lanl:
> > @Yurlungur I'd really like @jdolence's eyes on it as some point for a sanity check if at all possible, even if it's after the merge. Sometimes the original author has knowledge that is hard to obtain by reading alone.
> 
> I agree. My concern is I don't know when I will be able to get him to look. We were visiting him in Ann Arbor this week and I asked him to look, though. So there's hope.

I think his input will be valuable whenever we receive it.

singularity-eos/closure/mixed_cell_models.hpp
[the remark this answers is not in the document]
jhp-lanl: yep! I think this is a good reason _not_ to use this guess for tabular EOS. You could also imagine a reacting system where the reaction products is represented by a shifted ideal gas with a negative energy. I think this guess isn't really used, but I figured it might be good to put guardrails on it.

singularity-eos/closure/mixed_cell_models.hpp
jhp-lanl: This code is actually before the line search:
```c++
    // possibly scale the update to stay within reasonable bounds
    Real scale = s.ScaleDx();
    PORTABLE_REQUIRE(scale <= 1.0, "PTE Solver is attempting to increase the step size");

    // Line search
    Real gradfdx = -2.0 * scale * err;
    scale = 1.0; // New scale for line search
    Real err_old = err;
```

We reset the scale to 1.0 for the line search, and my read of the code is that the line search _should_ just limit the step.

My changes to the line search should be completely cosmetic and I don't believe the line search can increase the step beyond 1. I did add this comment and it might be less than clear. This code itself is not mine though:
```c++
      scale = 0.5;
      Real err_mid = s.TestUpdate(scale);
      if (err_mid < err && err_mid < err_old) {
        // try a larger step since a half step reduces the error
        scale = 0.75 + 0.5 * robust::ratio(err_mid - err, err - 2.0 * err_mid + err_old);
      }
```
I admit I didn't fully understand this code before, but after staring at it a bit more, I think I get it. I _think_ this code is basically assuming a roughly parabolic error and trying to guess where the minimum lies. This minimum is then used for the first iteration of the line search and _should_ be a pretty good guess.

But more specifically, I believe we can guarantee that `scale` lies somewhere between 0.25 and 0.75 when this code is executed. If `err_mid < err_old` and `err_mid < err` then the numerator is negative and `scale` can be at most 0.75 when `err_old` is very large compared to the difference between `err` and `err_mid`. On the other hand, if `err` is much larger than `err_old`, then the ratio goes to one and the scale will be at least 0.25.

If however the mid step is worse than the full step or the old error, then we just start at a full step (limited by `s.ScaleDx()`) and back up until we're decreasing the error by some appropriate amount.

I'll amend my comment to make this more clear as I assume this is why you thought my changes could increase the step somehow.

jhp-lanl:
> That said, I still need to fix the CI. I'll work on that this afternoon.

CI Fixed! Array was out of bounds, and I just bit the bullet and enabled proxy variables for all my terminal sessions rather than limiting them to interactive ones.

singularity-eos/closure/mixed_cell_models.hpp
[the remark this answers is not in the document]
pdmullen: This function is called in the `Init` for `PTESolverRhoU` which I believe is supposed to handle non-ideal EOS.

Notice that `TryIdealPTE` is commented out from `PTESolverRhoT` with the following comment:

```c++
    // Leave this in for now, but comment out because I'm not sure it's a good idea
    // TryIdealPTE(this);
    // Set the current guess for the equilibrium temperature.  Note that this is already
    // scaled.
```

singularity-eos/closure/mixed_cell_models.hpp
pdmullen: nice find

pdmullen (approved):

singularity-eos/closure/mixed_cell_models.hpp
```diff
@@ -170,9 +170,9 @@ class PTESolverBase {
   PORTABLE_INLINE_FUNCTION
   virtual void Fixup() const {
     Real vsum = 0;
-    for (int m = 0; m < nmat; ++m)
+    for (int m = 0; m < nmat; ++m) {
       vsum += vfrac[m];
-    PORTABLE_REQUIRE(vsum > 0., "Volume fraction sum is non-positive");
```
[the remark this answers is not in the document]
jhp-lanl: Basically I found that the check could be sufficiently removed from the source of the issue that all it was telling me was that _something_ went wrong in the PTE solver. Instead of enforcing a check here, I decided to rely on the checks I have in the `get_sg_eos()` interface. See this comment (https://github.com/lanl/singularity-eos/pull/383#issuecomment-2223535810) for more discussion.

Granted, this isn't ideal since the `get_sg_eos()` interface is only used by one host code, but that's really just because of the use case I was looking at. I created an issue to move this code into a more generic form later that can be invoked by any host code: https://github.com/lanl/singularity-eos/issues/388

sesame2spiner/examples/unit_tests/copper.dat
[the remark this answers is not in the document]
jhp-lanl: Updated

singularity-eos/closure/mixed_cell_models.hpp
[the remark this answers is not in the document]
pdmullen: I unresolved this conversation because I was concerned that `PTESolverRhoU ` now no longer works for non-ideal EOS wherein u can be less than 0.  Any thoughts here?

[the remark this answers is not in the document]
Yurlungur: I suggest that we floor the energy/temperature inside the solver, as it's only used for guesses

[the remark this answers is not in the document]
jhp-lanl: > I unresolved this conversation because I was concerned that `PTESolverRhoU ` now no longer works for non-ideal EOS wherein u can be less than 0. Any thoughts here?

If `PTESolverRhoU` relies on a negative temperature guess, I'd argue that it never worked for non-ideal EOS that can have negative energies and that the guess infrastructure should be re-worked for that solver.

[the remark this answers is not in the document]
jhp-lanl: Ah I misread the above. I forgot that `TryIdealPTE()` is separate from `GetIdealPTE()`. So these error checks being located in `GetIdealPTE()` is a mistake since `TryIdealPTE()` will check the results of `GetIdealPTE()` and see if the results are sane.

Even though these changes are highlighted just because I changed the message and not because they were added here, I'll move them to the `TryIdealPTE()` function.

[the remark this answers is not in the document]
jhp-lanl: Actually, I ended up getting rid of all the `GetIdealPTE` asserts and instead I added a check in `TryIdealPTE()` to make sure the temperature guess is actually positive.

singularity-eos/closure/mixed_cell_models.hpp
jhp-lanl: @Yurlungur I'm going to resolve this. If there's something that needs to be resolved, let me know. But I think I'm fairly confident that this assert in question shouldn't be triggered

Yurlungur: 👍 I'm happy with your explanation.

jhp-lanl:
I think this is ready to go now

Yurlungur:
:+1: Merge when ready

jhp-lanl:
Merging... @dholladay00 I think we're ready for a release

PR #386, opened 2024-07-10
install eos_mgusup.hpp

pbrady:
`eos/eos_mgusup.hpp` is missing from install


jhp-lanl (approved):
Good catch. If this is ready, we can merge now.

pbrady:
Thanks.  It's ready!

PR #389, opened 2024-07-12
Add missing eos_powermg header to install

pbrady:
Add `eos/eos_powermg.hpp` to installed headers


Yurlungur (approved):
Thanks for the fix!

PR #399, opened 2024-07-25
treat plugins as code dependencies

rbberger:
## PR Summary

This simplifies the plugin spack situation by just treating plugins like any other code dependencies.

Each plugin just has to install its source code into the spack prefix and the singularity-eos plugin needs to be registered using its name, spack package name, and the relative path to the plugins source code.

## PR Checklist

### DIFF
--- a/spack-repo/packages/singularity-eos/package.py
+++ b/spack-repo/packages/singularity-eos/package.py
@@ -11,9 +11,9 @@ from spack.directives import directive
 
 
 @directive("singularity_eos_plugins")
-def singularity_eos_plugin(name, path):
+def singularity_eos_plugin(name, spackage, relpath):
     def _execute_register(pkg):
-        pkg.plugins[name] = path
+        pkg.plugins[name] = (spackage, relpath)
 
     return _execute_register
 
@@ -82,7 +82,7 @@ class SingularityEos(CMakePackage, CudaPackage):
 
     plugins = {}
 
-    singularity_eos_plugin("dust", "example/plugin")
+    singularity_eos_plugin("dust", "self", "example/plugin")
 
     variant(
         "plugins",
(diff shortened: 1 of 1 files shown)


jhp-lanl:
@Yurlungur what are you thoughts? I think this will simplify things, but I admit I'll need to see a full release process of a plugin to really understand the details. But based off of our other conversations, I think this is a good route to go in.

Yurlungur:
I think I'm supportive of this design. @rbberger when resolving a spack spec, how does this work? Does the ordering get swapped so the plugin is a dependency of singularity?

rbberger:
@Yurlungur found one more issue due to the `variant`, which was a directory into the source tree. Since it now no longer needs to be in one folder, it'll convert `plugin/some/path` to `{plugin.prefix}/some/path`

Yurlungur:
> Right. Plugin spackages need to be made build dependencies and registered as plugins.

Does that mean we need to specify all available plugins in the spackage? Or can this happen "on the fly"? I.e., does this make the plugin infrastructure only useful for us? Or can others leverage it still?

rbberger:
It means that plugins can be added by manipulating the singularity-eos package since they need to become a build dependencies. Either directly or via inheritance.

rbberger:
So if there are other plugins out there, in order to use it'll have to be registered in our spackage.

Yurlungur:
Ok I think this is fine. But it is a downside since it means using a plugin with spack requires more "manual" intervention. But maybe that's unavoidable.

spack-repo/packages/singularity-eos/package.py
```diff
@@ -218,11 +218,23 @@ def cmake_args(self):
...
         if self.spec.satisfies("@1.9.0:"):
             if "none" not in self.spec.variants["plugins"].value:
-                pdirs = [join_path(self.stage.source_path, self.plugins[p]) for p in self.spec.variants["plugins"].value]
+                pdirs = []
+                for p in self.spec.variants["plugins"].value:
+                  spackage, path = self.plugins[p]
+                  if spackage == "self":
+                      pdirs.append(join_path(self.stage.source_path, path))
+                  else:
+                      pdirs.append(join_path(self.spec[spackage].prefix, path))
```
Yurlungur, line 227: Is this to deal with, e.g., the dust plugin in the examples?

rbberger: Yes. They are part of the singularity-eos source tree.

spack-repo/packages/singularity-eos/package.py
```diff
@@ -218,11 +218,23 @@ def cmake_args(self):
...
-            if self.spec.variants["variant"].value != "default":
-                args.append(self.define_from_variant("SINGULARITY_VARIANT", "variant"))
+            variant_path = self.spec.variants["variant"].value
+            if variant_path != "default":
+                parts = os.path.normpath('variant_path').split(os.sep)
+                if parts[0] in self.plugins.keys():
+                    spackage, path = self.plugins[parts[0]]
+                    parts[0] = self.spec[spackage].prefix
+                    variant_path = join_path(*parts)
+                args.append(self.define("SINGULARITY_VARIANT", variant_path))
```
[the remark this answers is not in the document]
rbberger: Similarly, there is the default variant from singularity-eos, and if it is provided by a plugin spackage we need to translate the relative path to an absolute one. We assume the first directory is the name of the plugin (e.g. `extra`), and that gets replaced by its install prefix.

PR #401, opened 2024-08-06
Fix for internal energy scaling in PTE closure

pdmullen:
## PR Summary
The _total_ volumetric internal energy 

$$ u_\mathrm{total} = \sum_m \bar{\rho}_m e_m $$  

can be a negative quantity.  I believe that the internal energy scaling routines in `singularity/main` `mixed_cell_models.hpp` do not respect this.  For example, here: https://github.com/lanl/singularity-eos/blob/main/singularity-eos/closure/mixed_cell_models.hpp#L313, an `abs` is applied to the (potentially negative) `sie_total` to compute `uscale`.  In the case of a negative `sie_total`, the sum of the resultant, scaled `u[m]` can be `-1`.  Inside residual calculations (e.g., https://github.com/lanl/singularity-eos/blob/main/singularity-eos/closure/mixed_cell_models.hpp#L594), a positive 1 is assumed for the sum over scaled material internal energies.  

I believe the fix here is to just remove the `abs` (as here: https://github.com/lanl/singularity-eos/commit/c843e16ab51ee27f7d870ee7e2ae1a7a8b5d52c6).  However, after some discussion with folks, this means that the sign of `uscale` can depend on the sign of `sie_tot`... an alternative choice is to introduce `uscale_total = (sie_tot > 0.0) ? 1.0 : -1.0`

## PR Checklist

### DIFF
--- a/singularity-eos/closure/mixed_cell_models.hpp
+++ b/singularity-eos/closure/mixed_cell_models.hpp
@@ -310,7 +310,7 @@ class PTESolverBase {
 
     // intialize rhobar array and final density
     InitRhoBarandRho();
-    uscale = rho_total * abs(sie_total);
+    uscale = rho_total * sie_total;
 
     // guess some non-zero temperature to start
     const Real Tguess = GetTguess();
@@ -328,11 +328,8 @@ class PTESolverBase {
     }
 
     // note the scaling of the material internal energy densities
-    uscale_total = 0.0;
-    for (int m = 0; m < nmat; ++m) {
+    for (int m = 0; m < nmat; ++m)
       u[m] = sie[m] * robust::ratio(rhobar[m], uscale);
-      uscale_total += u[m];
-    }
   }
 
   PORTABLE_INLINE_FUNCTION
@@ -455,7 +452,7 @@ class PTESolverBase {
   const RealIndexer &press;
   Real *jacobian, *dx, *sol_scratch, *residual, *u, *rhobar;
   CacheAccessor Cache;
-  Real rho_total, uscale, uscale_total, Tnorm;
+  Real rho_total, uscale, Tnorm;
 };
 
 } // namespace mix_impl
@@ -542,7 +539,6 @@ class PTESolverRhoT : public mix_impl::PTESolverBase<EOSIndexer, RealIndexer> {
   using mix_impl::PTESolverBase<EOSIndexer, RealIndexer>::press;
   using mix_impl::PTESolverBase<EOSIndexer, RealIndexer>::rho_total;
   using mix_impl::PTESolverBase<EOSIndexer, RealIndexer>::uscale;
-  using mix_impl::PTESolverBase<EOSIndexer, RealIndexer>::uscale_total;
   using mix_impl::PTESolverBase<EOSIndexer, RealIndexer>::MatIndex;
   using mix_impl::PTESolverBase<EOSIndexer, RealIndexer>::TryIdealPTE;
   using mix_impl::PTESolverBase<EOSIndexer, RealIndexer>::jacobian;
@@ -594,7 +590,8 @@ class PTESolverRhoT : public mix_impl::PTESolverBase<EOSIndexer, RealIndexer> {
       esum += u[m];
     }
     residual[0] = vfrac_total - vsum;
-    residual[1] = uscale_total - esum;
+    // the 1 here is the scaled total internal energy density
+    residual[1] = 1.0 - esum;
     for (int m = 0; m < nmat - 1; ++m) {
       residual[2 + m] = press[m + 1] - press[m];
     }
@@ -769,7 +766,6 @@ class PTESolverFixedT : public mix_impl::PTESolverBase<EOSIndexer, RealIndexer> 
   using mix_impl::PTESolverBase<EOSIndexer, RealIndexer>::press;
   using mix_impl::PTESolverBase<EOSIndexer, RealIndexer>::rho_total;
   using mix_impl::PTESolverBase<EOSIndexer, RealIndexer>::uscale;
-  using mix_impl::PTESolverBase<EOSIndexer, RealIndexer>::uscale_total;
   using mix_impl::PTESolverBase<EOSIndexer, RealIndexer>::MatIndex;
   using mix_impl::PTESolverBase<EOSIndexer, RealIndexer>::jacobian;
   using mix_impl::PTESolverBase<EOSIndexer, RealIndexer>::residual;
@@ -821,11 +817,9 @@ class PTESolverFixedT : public mix_impl::PTESolverBase<EOSIndexer, RealIndexer> 
       // note the scaling of pressure
       press[m] = eos[m].PressureFromDensityTemperature(rho[m], Tequil, Cache[m]);
     }
-    uscale_total = 0.0;
     for (int m = 0; m < nmat; ++m) {
       press[m] = robust::ratio(press[m], uscale);
       u[m] = sie[m] * robust::ratio(rhobar[m], uscale);
-      uscale_total += u[m];
     }
     Residual();
     return ResidualNorm();
@@ -837,6 +831,7 @@ class PTESolverFixedT : public mix_impl::PTESolverBase<EOSIndexer, RealIndexer> 
     for (int m = 0; m < nmat; ++m) {
       vsum += vfrac[m];
     }
+    // the 1 here is the scaled volume fraction
     residual[0] = vfrac_total - vsum;
     for (int m = 0; m < nmat - 1; ++m) {
       residual[1 + m] = press[m] - press[m + 1];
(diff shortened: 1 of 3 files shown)


Yurlungur (approved):
Thanks for the fix! This was a subtle one.

dholladay00:
This does bring about a larger discussion of `uscale` in general. Generically, it is defined by sie_total at the moment, though the `Fixed[P,T]` solvers define it internal to the solver. My understanding is it is a quantity meant to stabilize the solver by capturing the rough order of magnitude of the energies and scaling them brings quantities close to O(1). One could think of it as a very simple preconditioner (like diagonal scaling but different, though @jdolence could provide better context). When the signs are allowed to differ, a sum may not represent such a quantity. It seems like uscale (this may be beyond the scope of this PR btw) should be defined by some norm or measure of the internal energy vector or some other scheme (shifting all energies by a constant such that all have the same sign and taking an average).

dholladay00:
at the very least for this PR, perhaps changing the `uscale` sum to be a sum of absolute values of the energies should also help those solvers.

Yurlungur:
> at the very least for this PR, perhaps changing the `uscale` sum to be a sum of absolute values of the energies should also help those solvers.

I think this was what it was originally and it caused problems. @pdmullen can comment though.

pdmullen:
> > at the very least for this PR, perhaps changing the `uscale` sum to be a sum of absolute values of the energies should also help those solvers.
> 
> I think this was what it was originally and it caused problems. @pdmullen can comment though.

Maybe a few thoughts here... 

First, I tend to agree that there are multiple ways of handling the scaling of internal energies.  As @dholladay00 argues, if we were careful, we could simply apply a shift so that the nuance of negative numbers don't bite us.  @jhp-lanl on mattermost also thought about the scenario wherein the scaled internal energies summed to exactly zero, and how to best handle that situation.  Lots of opportunity here.

Second, just to say it aloud... this PR is fixing a bug.  The bug was that the scaled internal energies could sum to `-1` when `sie_tot` < 0, despite `mixed_cell_models.hpp` assuming a sum to `1` elsewhere.  Maybe I don't quite understand your comment regarding re-introducing absolute values, as my first take is that this would just reintroduce the original issue, but I suspect I am misreading your comment.

dholladay00:
My comment wasn't necessarily related to the new introduced parameter as it seems to do the right thing. But seeing this PR with the fact that sie's can have opposite sign at the forefront of my mind, I think we should look again at `uscale` at some point. It is unlikely though not impossible that `uscale` could be 0 or near 0, which I don't think is the intent of the usage of `uscale`. My understanding of its usage is that it is supposed to bring sie-ish quantities close to O(1) by having uscale provide "a rough order of magnitude" of an internal energy value. Adding up values with opposing signs will not accomplish this task.

pdmullen:
Sorry folks... I have reverted to my original design for this fix.  

In contrast to the fixed P solver implementation for example (see https://github.com/lanl/singularity-eos/blob/main/singularity-eos/closure/mixed_cell_models.hpp#L1020), when `sie_tot` is provided, I feel the desired scaling is set by the bulk internal energy constraint, not the initial guess.

I have gone the route of simply eliminating the `abs` again, which means that `uscale` can be either positive or negative,  depending on the sign of `sie_tot`.  If you feel too strongly against this, I think the only other reasonable strategy in the short-term is introducing a `uscale_total = (sie_tot > 0.0) ? 1.0 : -1.0`.

jhp-lanl:
> Sorry folks... I have reverted to my original design for this fix.
> 
> In contrast to the fixed P solver implementation for example (see https://github.com/lanl/singularity-eos/blob/main/singularity-eos/closure/mixed_cell_models.hpp#L1020), when `sie_tot` is provided, I feel the desired scaling is set by the bulk internal energy constraint, not the initial guess.
> 
> I have gone the route of simply eliminating the `abs` again, which means that `uscale` can be either positive or negative, depending on the sign of `sie_tot`. If you feel too strongly against this, I think the only other reasonable strategy in the short-term is introducing a `uscale_total = (sie_tot > 0.0) ? 1.0 : -1.0`.
> 
> EDIT: Just for completeness, in my previous implementation (if anyone looks there), I missed a re-normalization.

I think what you had originally (in this MR at least) is the right approach and I don't think you've provided enough justification for _why_ that approach doesn't work yet. For this MR to go in, we need strong justification for _why_ the `uscale_total` approach doesn't work and why this is "the only reasonable strategy".

I am strongly advocating for the `uscale_total` approach because the discussion has already brought up several edge cases. Your MR shouldn't address these edge cases, but if you just remove the `abs()` then you're also not _helping_ other developers address the edge cases in the future. These fixes could be setting a floor on `uscale`, summing the absolute values, or even just using the maximum absolute value. Regardless of what the fix is, I think it makes sense to set `uscale_total` the way you were setting it and not hard-code `1` as the sum convergence criterion.

jhp-lanl (changes requested):
Can you please add the case that causes the PTE solver to fail in the unit tests?
https://github.com/lanl/singularity-eos/blob/main/test/test_closure_pte.cpp

pdmullen:
> I think what you had originally (in this MR at least) is the right approach and I don't think you've provided enough justification for _why_ that approach doesn't work yet. For this MR to go in, we need strong justification for _why_ the `uscale_total` approach doesn't work and why this is "the only reasonable strategy".

> I am strongly advocating for the `uscale_total` approach because the discussion has already brought up several edge cases. Your MR shouldn't address these edge cases, but if you just remove the `abs()` then you're also not _helping_ other developers address the edge cases in the future. These fixes could be setting a floor on `uscale`, summing the absolute values, or even just using the maximum absolute value. Regardless of what the fix is, I think it makes sense to set `uscale_total` the way you were setting it and not hard-code `1` as the sum convergence criterion.

The implementation in https://github.com/lanl/singularity-eos/pull/401/commits/9c35a24d05810d5374a586da6b595eae807f5bf5 contained ill-motivated, erroneous scaling _and_ residual estimations.  This was my fault... apologies; and I could have better described the flaw.  The original, _actual_ bug fix (i.e., simply eliminating `abs`) is the correct patch, hence my revert.  However, as you allude to, we can still consider different designs for applying the exact same patch.  In particular, we can accomplish it through the introduction of a `utotal_scale`, which might be a more extendible design when considering different edges in the future (c.f., comments above from @dholladay00).

I suspect that earlier we weren't on the same page about what was going on in with my failed attempt in https://github.com/lanl/singularity-eos/pull/401/commits/9c35a24d05810d5374a586da6b595eae807f5bf5.  If okay with all, let's forget about that implementation and reconsider afresh the most recent commit.  For this reason, I am re-triggering review requests.  

> Can you please add the case that causes the PTE solver to fail in the unit tests?
https://github.com/lanl/singularity-eos/blob/main/test/test_closure_pte.cpp

Will try my luck at this tomorrow.

jhp-lanl:
> I suspect that earlier we weren't on the same page about what was going on in with my failed attempt in https://github.com/lanl/singularity-eos/commit/9c35a24d05810d5374a586da6b595eae807f5bf5. If okay with all, let's forget about that implementation and reconsider afresh the most recent commit. For this reason, I am re-triggering review requests.

Yes that was my fault. I failed to appreciate what the sum over `u[m]` actually meant.

singularity-eos/closure/mixed_cell_models.hpp
```diff
@@ -310,7 +310,10 @@ class PTESolverBase {
 
     // intialize rhobar array and final density
     InitRhoBarandRho();
-    uscale = rho_total * abs(sie_total);
+    const Real utotal = rho_total * sie_total;
+    uscale = std::abs(utotal);
+    // TODO(): Consider edge case when utotal \simeq 0
+    utotal_scale = robust::ratio(utotal, uscale);
```
jhp-lanl, line 316: 👍

test/test_closure_pte.cpp
```diff
@@ -209,9 +229,18 @@ SCENARIO("Density-Temperature PTE Solver", "[PTE]") {
...
+                                                      sie_bulk, mass_frac, u_bulk_out);
+        // // Check that PTE converged
         // CHECK(pte_converged);
+        // // Check that PTE obeys the bulk internal energy constraint
+        // // NOTE(@pdmullen): The following fails prior to PR401
+        // const Real u_bulk = ratio(sie_bulk, spvol_bulk);
+        // const Real u_scale = std::abs(u_bulk);
+        // const Real u_bulk_scale = ratio(u_bulk, u_scale);
+        // const Real residual = std::abs(u_bulk_scale - ratio(u_bulk_out, u_scale));
+        // CHECK(residual < pte_rel_tolerance_e);
```
jhp-lanl, line 243: This looks good (assuming you'll uncomment it). Can you put this in an `AND_THEN` clause since it's separate from convergence?

pdmullen: I was planning on keeping this one commented... Notice
```c++
// Check(pte_converged)
```
itself is commented out for this test... with the accompanying comment: `// TODO: make this test converge`.  In other words, it seems that `GIVEN("A difficult two-material state")` is WIP...

The new, _active_ test is then only being applied here: https://github.com/lanl/singularity-eos/pull/401/files#diff-c129bbafa2a818eacbd83e80aa4797afe3a8d4ef822276feaa250a77eada78fcR208-R214 ; i.e., for the ` GIVEN("A difficult three-material state")` test.

I'll try to add the `AND_THEN` clause.

jhp-lanl: Yep! That is correct. I am intending this test as a repository of states that fail even if we don't _yet_ have fixes. That way people can document bad behavior without the need to necessarily fix it yet.

I missed your test above. Sorry about that. Yes please just put that in an `AND_THEN` clause. You can simply move your comment into the label string for the `AND_THEN` clause.

From a testing perspective, it just helps separate out the checks to ensure that we understand exactly _what_ is being tested.

pdmullen: Added the `AND_THEN` clause in [86232de](https://github.com/lanl/singularity-eos/pull/401/commits/86232de2c1153a7d8f4e7e99b192524826468a87)

jhp-lanl: Looks good

jhp-lanl:
Merging. LANL gitlab tests pass

Issue #409, opened 2024-08-24
EOSPAC backend update in CI


Yurlungur:
We need at least 6.5.7 for use with shared memory. This must also be threaded through the spackage so that `SINGULARITY_EOSPAC_ENABLE_SHMEM` is set to true when that version of eospac is used.

Issue #408, opened 2024-08-24
EOSPAC backend with serialization/deserialization


Yurlungur:
I am able to get the EOSPAC backend to work with serialization/deserialization and shared memory with a single instance. However in the case where eospac is constructed for multiple materials, There will be problems, because the library backend is global while an EOS object is per material. Worse, the de-serialization operation is global. There is no way to deserialize for only one material.

I am not sure what the solution to this is. We may wish to extend the EOSPAC object to load multiple materials at once. Alternatively perhaps there's a way to serialize and de-serialize a *set* of EOS objects, all at once. That could get around this problem.

jhp-lanl:
Let's try destroying just the tables that the class owns and see what happens. We'll discuss with EOSPAC team.

PR #373, opened 2024-05-10
Robustness additions to `get_sg_eos*()` functions and PTE solver

jhp-lanl:
## PR Summary

1. **Adds initialization of the scratch space for the PTE solvers and EOS in the `get_sg_eos()` functions.
    This is needed because these can sometimes initialize to `-nan` and trigger an FPE in the spiner EOS when it tries to check whether the values are garbage or not. Comparing to a NaN is an FPE in the GNU compiler.
2. **Adds a debug check to make sure the PTE solver is outputting a positive temperature**
    This was more of a sanity check than anything, but since it only exists in debug builds, I don't think it's much of an issue.

## PR Checklist

- [ ] Adds a test for any bugs fixed. Adds tests for new features.
- [x] Format your changes by using the `make format` command after configuring with `cmake`.
- N/A Document any new features, update documentation for changes made.
- [x] Make sure the copyright notice on any files you modified is up to date.
- [x] After creating a pull request, note it in the CHANGELOG.md file.
- [ ] LANL employees: make sure tests pass both on the github CI and on the Darwin CI

@dholladay00 can you review?

PR #410, opened 2024-08-24
Enable serialization and shared memory

Yurlungur:
## PR Summary

This PR enables both serialization and shared memory for all equations of state. This is for use with MPI Windows shared memory on, e.g., Rocinante and other limited environments. Note there is still an outstanding issue with the EOSPAC backend, see #408 , which I will need to resolve in a future MR. I'm still thinking about how best to resolve that particular issue, and insight is welcome.

Another outstanding issue is that the CI does not currently test against EOSPAC with shared memory enabled. When possible, we should update the CI, and thread this through the spackage. @rbberger maybe you can handle that in a separate PR at some point? #409. 

When shared memory is not enabled in EOSPAC, the machinery all still works. An EOSPAC EOS will report that it cannot utilize any shared memory, but the serialization and de-serialization routines will still work. The CI on Darwin will test this and I've also tested it by hand.

In the meantime, I plan to test the shared memory case by hand after I get a more up-to-date eospac environment set up through spack.

## PR Checklist


CMakeLists.txt
```diff
@@ -47,6 +47,9 @@ cmake_dependent_option(
 option(SINGULARITY_USE_FORTRAN "Enable fortran bindings" ON)
 option(SINGULARITY_USE_KOKKOS "Use Kokkos for portability" OFF)
 option(SINGULARITY_USE_EOSPAC "Enable eospac backend" OFF)
+option(SINGULARITY_EOSPAC_ENABLE_SHMEM
```
Yurlungur, line 50: I know `SHMEM` is something else but hopefully next to `EOSPAC` it's clear what this means

Yurlungur (commented):
I also simplified and cleaned up the `GetOnDevice` and `Finalize` methods for `spiner` based equations of state and cleaned up the code for modifiers and the `UnModify` machinery.

This MR is large, but in most files the changes are tirivial and most of the new code is tests.

AstroBarker (approved):
Very cool! Not much stood out to me, besides a couple of typos, but another look is probably good

singularity-eos/base/serialization_utils.hpp
```diff
@@ -0,0 +1,105 @@
...
+    });
+  }
+  std::size_t Serialize(char *dst) {
+    std::size_t vsize = eos_objects.size();
+    memcpy(dst, &vsize, sizeof(std::size_t));
+    std::size_t offst = sizeof(std::size_t);
+    for (auto &eos : eos_objects) {
+      offst += eos.Serialize(dst + offst);
+    }
+    PORTABLE_REQUIRE(offst == SerializedSizeInBytes(), "Serialization successful");
```
jhp-lanl, line 63: Does this need to be debug only? Also I thought the message is for when this _fails_ not when it succeeds.

singularity-eos/base/serialization_utils.hpp
```diff
@@ -0,0 +1,105 @@
...
+    return offst;
+  }
+  auto Serialize() {
+    std::size_t size = SerializedSizeInBytes();
+    char *dst = (char *)malloc(size);
+    std::size_t new_size = Serialize(dst);
+    PORTABLE_REQUIRE(size == new_size, "Serialization successful");
+    return std::make_pair(size, dst);
```
jhp-lanl, line 71: Just to make sure I understand correctly, is `size` here the same as `vsize`, i.e. the first `size_t` bytes of `dst`?

singularity-eos/base/serialization_utils.hpp
```diff
@@ -0,0 +1,105 @@
...
+  }
+  std::size_t DeSerialize(char *src,
+                          const SharedMemSettings &stngs = DEFAULT_SHMEM_STNGS) {
+    std::size_t vsize;
+    memcpy(&vsize, src, sizeof(std::size_t));
+    eos_objects.resize(vsize);
+    SharedMemSettings stngs_loc = stngs; // non-const
+    std::size_t offst = sizeof(std::size_t);
```
jhp-lanl, line 79: Is this initialized to `sizeof(std::size_t)` because of the `vsize` data?

singularity-eos/base/serialization_utils.hpp
```diff
@@ -0,0 +1,105 @@
...
+    std::size_t shared_offst = 0;
+    for (auto &eos : eos_objects) {
+      std::size_t shared_bytes = eos.SharedMemorySizeInBytes();
+      offst += eos.DeSerialize(src + offst, stngs_loc);
+      stngs_loc.data += shared_bytes;
+      shared_offst += shared_bytes;
+    }
+    PORTABLE_REQUIRE(offst = SerializedSizeInBytes(), "De-Serialization successful");
+    PORTABLE_REQUIRE(shared_offst = SharedMemorySizeInBytes(),
+                     "De-Serialization into shared memory successful");
```
jhp-lanl, line 89: Again, should these _just_ be debug checks? I'd argue this is important enough to be a `PORTABLE_ALWAYS_REQUIRE`.

And again, I'm pretty sure the message is for failure, not success.

singularity-eos/eos/eos_base.hpp
```diff
@@ -76,16 +77,28 @@ char *StrCat(char *destination, const char *source) {
...
+  std::size_t DynamicMemorySizeInBytes() const { return t_.DynamicMemorySizeInBytes(); } \
+  std::size_t SharedMemorySizeInBytes() const { return t_.SharedMemorySizeInBytes(); }   \
+  std::size_t DumpDynamicMemory(char *dst) { return t_.DumpDynamicMemory(dst); }         \
+  std::size_t SetDynamicMemory(char *src,                                                \
+                               const SharedMemSettings &stngs = DEFAULT_SHMEM_STNGS) {   \
+    return t_.SetDynamicMemory(src, stngs);                                              \
+  }                                                                                      \
+  constexpr bool AllDynamicMemoryIsShareable() const {                                   \
+    return t_.AllDynamicMemoryIsShareable();                                             \
+  }
```
jhp-lanl, line 101: I think this is a good enough solution for the time being. Good work!

singularity-eos/eos/eos_base.hpp
```diff
@@ -657,13 +670,76 @@ class EosBase {
...
 
   inline constexpr decltype(auto) UnmodifyOnce() { return *static_cast<CRTP *>(this); }
 
   inline constexpr decltype(auto) GetUnmodifiedObject() {
-    return *static_cast<CRTP *>(this);
+    if constexpr (CRTP::IsModified()) {
+      return ((static_cast<CRTP *>(this))->UnmodifyOnce()).GetUnmodifiedObject();
+    } else {
+      return *static_cast<CRTP *>(this);
+    }
```
jhp-lanl, line 742: Nice!

singularity-eos/eos/eos_davis.hpp
```diff
@@ -193,6 +200,10 @@ class DavisProducts : public EosBase<DavisProducts> {
   DavisProducts(const Real a, const Real b, const Real k, const Real n, const Real vc,
                 const Real pc, const Real Cv)
       : _a(a), _b(b), _k(k), _n(n), _vc(vc), _pc(pc), _Cv(Cv) {}
+  PORTABLE_INLINE_FUNCTION
+  void CheckParams() const {
+    // TODO(JMM): Stub.
```
jhp-lanl, line 205: Can you create an issue to just go through these and add more conditions? No need to assign yourself.

singularity-eos/eos/eos_eospac.hpp
```diff
@@ -140,7 +140,18 @@ class EOSPAC : public EosBase<EOSPAC> {
                 bool apply_smoothing = false,
                 eospacSplit apply_splitting = eospacSplit::none,
                 bool linear_interp = false);
+  PORTABLE_INLINE_FUNCTION void CheckParams() const {
+    // TODO(JMM): I really don't know what to put here...
```
jhp-lanl, line 144: I _think_ all `matid` are positive, but this may not _always_ be the case.

singularity-eos/eos/eos_eospac.hpp
```diff
@@ -1208,6 +1235,44 @@ inline EOSPAC::EOSPAC(const int matid, bool invert_at_setup, Real insert_data,
...
+  EOS_INTEGER error_code = EOS_OK;
+  eos_GetPackedTables(NTABLES, tablehandle, (EOS_CHAR *)dst, &error_code);
+  eosCheckError(error_code, "eos_GetPackedTables", Verbosity::Debug);
+  return packed_size_; // JMM: Note this is NOT shared memory size
+}
+inline std::size_t EOSPAC::SetDynamicMemory(char *src, const SharedMemSettings &stngs) {
+  static_assert(sizeof(char) == sizeof(EOS_CHAR), "EOS_CHAR is one byte");
+  EOS_INTEGER NTABLES[] = {NT};
+  EOS_INTEGER error_code = EOS_OK;
+  eos_DestroyAll(&error_code);
```
jhp-lanl, line 1251: I think we can just destroy the table handles owned by the class. I _think_ the `Finalize()` member function should do this, right?

singularity-eos/eos/eos_gruneisen.hpp
```diff
@@ -63,6 +63,11 @@ class Gruneisen : public EosBase<Gruneisen> {
         _Cv(Cv), _rho_max(RHOMAX_SAFETY * ComputeRhoMax(s1, s2, s3, rho0)) {}
   static PORTABLE_INLINE_FUNCTION Real ComputeRhoMax(const Real s1, const Real s2,
                                                      const Real s3, const Real rho0);
+  PORTABLE_INLINE_FUNCTION
+  void CheckParams() const {
+    PORTABLE_ALWAYS_REQUIRE(_T0 >= 0, "Non-negative reference temperature");
+    PORTABLE_ALWAYS_REQUIRE(_rho0 >= 0, "Non-negative reference density");
```
jhp-lanl, line 69: Just because I've spent so much time with this EOS...
```suggestion
    PORTABLE_ALWAYS_REQUIRE(_T0 >= 0, "Non-negative reference temperature required");
    PORTABLE_ALWAYS_REQUIRE(_rho0 >= 0, "Non-negative reference density required");
    PORTABLE_ALWAYS_REQUIRE(_C0 >= 0, "Non-negative Hugoniot intercept required");
    PORTABLE_ALWAYS_REQUIRE(_Cv >= 0, "Non-negative heat capacity required");
    PORTABLE_ALWAYS_REQUIRE(_rho_max > _rho0, "Maximum density must be greater than reference");
```

singularity-eos/eos/eos_helmholtz.hpp
```diff
@@ -203,16 +205,39 @@ inline void SetTablesFromFile(std::ifstream &file, int n1, Real r1min, Real r1ma
...
+    return SpinerTricks::DynamicMemorySizeInBytes(this);
+  }
+  std::size_t DumpDynamicMemory(char *dst) {
+    return SpinerTricks::DumpDynamicMemory(dst, this);
+  }
+  std::size_t SetDynamicMemory(char *src,
+                               const SharedMemSettings &stngs = DEFAULT_SHMEM_STNGS) {
+    return SpinerTricks::SetDynamicMemory((stngs.data == nullptr) ? src : stngs.data,
+                                          this);
+  }
```
jhp-lanl, line 240: This is _very_ compact! Nice!

singularity-eos/eos/eos_helmholtz.hpp
```diff
@@ -880,63 +924,15 @@ inline void HelmElectrons::InitDataFile_(const std::string &filename) {
...
-  other.ef_ = Spiner::getOnDeviceDataBox(ef_);
-  other.efd_ = Spiner::getOnDeviceDataBox(efd_);
-  other.eft_ = Spiner::getOnDeviceDataBox(eft_);
-  other.efdt_ = Spiner::getOnDeviceDataBox(efdt_);
-  other.xf_ = Spiner::getOnDeviceDataBox(xf_);
-  other.xfd_ = Spiner::getOnDeviceDataBox(xfd_);
-  other.xft_ = Spiner::getOnDeviceDataBox(xft_);
-  other.xfdt_ = Spiner::getOnDeviceDataBox(xfdt_);
-  return other;
+  return SpinerTricks::GetOnDevice(this);
```
jhp-lanl, line 932: 👏

test/test_eos_tabulated.cpp
```diff
@@ -257,6 +262,98 @@ SCENARIO("SpinerEOS depends on rho and sie", "[SpinerEOS],[DependsRhoSie]") {
     steelEOS.Finalize();      // cleans up device memory
   }
 }
+
+SCENARIO("SpinerEOS and EOSPAC Serialization",
+         "[SpinerEOS][DependsRhoT][DependsRhoSie][EOSPAC][Serialization]") {
+  GIVEN("Eoses initialized with matid") {
+    SpinerEOSDependsRhoT rhoT_orig = SpinerEOSDependsRhoT(eosName, steelID);
+    SpinerEOSDependsRhoSie rhoSie_orig = SpinerEOSDependsRhoSie(eosName, steelID);
+    EOS eospac_orig = EOSPAC(steelID);
```
jhp-lanl, line 271: I think we talked about this, but can you add another EOSPAC class and see if it can serialize/de-serialize properly?

jhp-lanl (approved):
Some small changes. I'll approve for now soas not to hold anything up.

jhp-lanl:
I'm guessing you're also missing an `ifdef` in the serialization tests

doc/sphinx/src/using-eos.rst
```diff
@@ -39,6 +40,247 @@ A more generic version of the vector calls exists in the ``Evaluate``
...
+    EOS eos_device = eos.GetOnDevice();
+    // ...
+  }
+
+
+.. warning::
+
+  Since EOSPAC is a library, DeSerialization is destructive for EOSPAC
+  and may have side-effects. DeSerializing an EOSPAC object will
+  *completely* reset the EOSPAC backend.
```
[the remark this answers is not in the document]
Yurlungur: Agreed. I will walk back on this warning and just say there may be side effects... to be tested.

singularity-eos/base/serialization_utils.hpp
```diff
@@ -0,0 +1,105 @@
...
+namespace singularity {
+
+template <typename EOS>
+struct BulkSerializer {
+  BulkSerializer() = default;
+  explicit BulkSerializer(const std::initializer_list<EOS> &eos_objects_)
+      : eos_objects(eos_objects_) {}
+  explicit BulkSerializer(const std::vector<EOS> &eos_objects_)
+      : eos_objects(eos_objects_) {}
+  explicit BulkSerializer(const int N) : eos_objects(N) {}
```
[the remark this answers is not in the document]
Yurlungur: You can't---it was just a convenience function for setting the size of the vector. I'll remove it as I agree it's misleading.

singularity-eos/base/serialization_utils.hpp
```diff
@@ -0,0 +1,105 @@
...
+  const EOS &Get(std::size_t i) const { return eos_objects[i]; }
+  EOS &Get(std::size_t i) { return eos_objects[i]; }
```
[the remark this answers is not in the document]
Yurlungur: Nope---the `std::vector` accessor has `const` and `non-const` versions, and the `const` one returns a `const` reference whereas the non-const one returns a non-const reference. This has more to do with how stl containers are decorated than anything else.

singularity-eos/base/serialization_utils.hpp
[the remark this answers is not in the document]
jhp-lanl: Yeah, if you need that convenience I'd argue for using a static member function instead.

singularity-eos/base/serialization_utils.hpp
```diff
@@ -0,0 +1,105 @@
...
+
+namespace singularity {
+
+template <typename EOS>
+struct BulkSerializer {
+  BulkSerializer() = default;
+  explicit BulkSerializer(const std::initializer_list<EOS> &eos_objects_)
+      : eos_objects(eos_objects_) {}
+  explicit BulkSerializer(const std::vector<EOS> &eos_objects_)
+      : eos_objects(eos_objects_) {}
```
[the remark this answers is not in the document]
Yurlungur: Mainly laziness. We can template.

singularity-eos/base/serialization_utils.hpp
Yurlungur: Will change it to always require and make the error message informative for failure

singularity-eos/base/serialization_utils.hpp
Yurlungur: `size` here is the total number of bytes in `dst`. It's the same as the `SerializedSizeInBytes` reported by the object. This includes the "header" and metadata info as well as the dyamic and static memory of each eos.

Yurlungur: this means it's `sizeof(std::size_t) + vsize*(EOS::SerializedSizeInBytes())` if all EOS's are the same same size. But it might be different if each EOS is a different type or has different table sizes.

singularity-eos/base/serialization_utils.hpp
Yurlungur: correct. For the `BulkSerializer`, the first `sizeof(size_t)` bytes are the `vsize` then after that it's all the EOS's contained in the object, one after the other, in contiguous memory.

singularity-eos/base/serialization_utils.hpp
Yurlungur: :+1: Will fix.

singularity-eos/eos/eos_davis.hpp
Yurlungur: :+1: #411

singularity-eos/eos/eos_eospac.hpp
Yurlungur: Yeah I was worried that a user might specify a negative matid for their own custom EOS just to avoid conflicts with the sesame database for example. I'll at least check the min density and temperature are non-negative.

singularity-eos/eos/eos_eospac.hpp
Yurlungur: Yes I think you're right. I will remove this `eos_DestroyAll` and just rely on the user calling `Finalize` appropriately as suggested by the documentation.

test/test_eos_tabulated.cpp
Yurlungur: Yeah we talked about this. Let me get that in in a subsequent MR after understanding the EOSPAC properties in a self-contained test.

Yurlungur:
Thanks for the feedback @jhp-lanl ! I think it should all be addressed now... modulo (a) CI issues and (b) the EOSPAC question we discussed. I will resolve the latter in a separate MR after doing some testing in a self-contained reproducer.

> I'm guessing you're also missing an `ifdef` in the serialization tests which is causing the minimal tests to fail.

Nope, it was actually unsanitized input in `test_sg_eos` after MR #373 . Not sure why it only showed up now. Should have shown up in earlier MRs.

singularity-eos/base/serialization_utils.hpp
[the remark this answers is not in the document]
Yurlungur: It's irrelevant now. To support general templating on container type I just got rid of these. `eos_objects` is public so no reason to even have a `Get` method anyway. ;)

jhp-lanl:
> Nope, it was actually unsanitized input in `test_sg_eos` after MR #373 . Not sure why it only showed up now. Should have shown up in earlier MRs.

wow! Thanks for the find and fix!

doc/sphinx/src/using-eos.rst
```diff
@@ -39,6 +40,271 @@ A more generic version of the vector calls exists in the ``Evaluate``
...
+       Initialization code goes here
+     */
+    serializer = singularity::VectorSerializer<EOS>(eos_vec);
+    packed_size = serializer.SerializedSizeInBytes();
+    shared_size = serializer.SharedMemorySizeInBytes();
+  }
+
+  // Send sizes
+  MPI_Bcast(&packed_size, 1, MPI_INT, 0, MPI_COMM_WORLD);
+  MPI_Bcast(&packed_size, 1, MPI_INT, 0, MPI_COMM_WORLD);
```
AstroBarker, line 266: Should be `MPI_Bcast(&shared_size, ...)`

Yurlungur: :+1: good catch

doc/sphinx/src/using-eos.rst
```diff
@@ -39,6 +40,271 @@ A more generic version of the vector calls exists in the ``Evaluate``
...
+  ``eos.GetOnDevice()``. Attempting to serialize device-initialized
+  objects is undefined behavior, but will likely result in a
+  segmentation fault.
+
+The pair is the pointer and its size. The function
+
+.. code-block:: cpp
+
+  std::size_t EOS::DeSerialize(char *src,
+                const SharedMemSettings &stngs = DEFAULT_SHMEM_SETTINGS)
```
AstroBarker, line 101: Should be `DEFAULT_SHMEM_STNGS` here and a few places below.

doc/sphinx/src/using-eos.rst
```diff
@@ -39,6 +40,271 @@ A more generic version of the vector calls exists in the ``Evaluate``
...
+    /*
+       Initialization code goes here
+     */
+    serializer = singularity::VectorSerializer<EOS>(eos_vec);
+    packed_size = serializer.SerializedSizeInBytes();
+    shared_size = serializer.SharedMemorySizeInBytes();
+  }
+
+  // Send sizes
+  MPI_Bcast(&packed_size, 1, MPI_INT, 0, MPI_COMM_WORLD);
```
dholladay00, line 265: Should these be `MPI_LONG`'s?

doc/sphinx/src/using-eos.rst
```diff
@@ -39,6 +40,271 @@ A more generic version of the vector calls exists in the ``Evaluate``
...
+  }
+  singularity::VectorSerializer<EOS> deserializer;
+  deserializer.DeSerialize(packed_data, settings);
+  if (use_mpi_shared_memory) {
+    // same MPI code as above
+  }
+  // extract each individual EOS and do something with it
+  std::vector<EOS> eos_host_vec = deserializer.eos_objects;
+  // get on device if you want
+  for (auto EOS : eos_host_vec) {
```
dholladay00, line 291: This could be a can of worms at some point when we want to have 1 copy per GPU rather than 1 copy per node. Some host codes may want to hand us GPU memory so we may need to have a facility to handle that case at some point.

doc/sphinx/src/using-eos.rst
```diff
@@ -445,6 +711,23 @@ unmodified EOS model, call
 
 The return value here will be either the type of the ``EOS`` variant
 type or the unmodified model (for example ``IdealGas``) or, depending
+
```
dholladay00, line 714: What going on here @Yurlungur ?

singularity-eos/base/constants.hpp
```diff
@@ -32,11 +32,21 @@ constexpr unsigned long all_values = (1 << 7) - 1;
 } // namespace thermalqs
 
 constexpr size_t MAX_NUM_LAMBDAS = 3;
-enum class DataStatus { Deallocated = 0, OnDevice = 1, OnHost = 2 };
+enum class DataStatus { Deallocated = 0, OnDevice = 1, OnHost = 2, UnManaged = 3 };
```
dholladay00, line 35: Do we have sufficient coverage in our testing to make sure we are catching the potential issues? I recall this took up many of @Yurlungur's cycles to get this to work correctly back in the day.

singularity-eos/eos/eos_base.hpp
```diff
@@ -657,13 +670,76 @@ class EosBase {
...
+    const CRTP *pcrtp = static_cast<const CRTP *>(this);
+    return pcrtp->DynamicMemorySizeInBytes();
+  }
+  constexpr bool AllDynamicMemoryIsShareable() const { return true; }
+  // JMM: These are generic and never need to be special-cased.
+  std::size_t SerializedSizeInBytes() const {
+    // sizeof(*this) returns the size of JUST the base class.
+    const CRTP *pcrtp = static_cast<const CRTP *>(this);
+    std::size_t dyn_size = pcrtp->DynamicMemorySizeInBytes();
+    return dyn_size + sizeof(CRTP);
```
dholladay00, line 692: hmmm... won't this give the size of the underlying class rather than the size of the variant? Without thinking too hard, I'd think we want the size of the variant, maybe that's the subtlety in the 2 size functions ?

singularity-eos/eos/eos_helmholtz.hpp
```diff
@@ -203,16 +205,39 @@ inline void SetTablesFromFile(std::ifstream &file, int n1, Real r1min, Real r1ma
 } // namespace HelmUtils
 
 class HelmElectrons {
+  friend class table_utils::SpinerTricks<HelmElectrons>;
+  using SpinerTricks = table_utils::SpinerTricks<HelmElectrons>;
```
dholladay00, line 209: ahhh, very good!

test/test_eos_tabulated.cpp
```diff
@@ -257,6 +262,98 @@ SCENARIO("SpinerEOS depends on rho and sie", "[SpinerEOS],[DependsRhoSie]") {
...
+        free(rhoSie_shared_data);
+        free(eospac_shared_data);
+      }
+      free(rhoT_data);
+      free(rhoSie_data);
+      free(eospac_data);
+    }
+
+    rhoT_orig.Finalize();
+    rhoSie_orig.Finalize();
```
dholladay00, line 353: Maybe I knew this once and forgot, but I'm realizing that this will likely require substantial modifications in the fortran interface to account for the new ways of initialization and finalization.

doc/sphinx/src/using-eos.rst
Yurlungur: They sure should!

Yurlungur: Updated but leaving unresolved in case this is the cause of @AstroBarker 's bugs in riot.

doc/sphinx/src/using-eos.rst
Yurlungur: What this will *currently* produce is 1 copy per MPI rank on device and 1 copy per node on host, because `GetOnDevice` will do its own allocation. The case where we want device shared memory is one I'm aware of but haven't thought deeply about. My suggestion is we punt on this but agree there's an engineering problem to solve here.

doc/sphinx/src/using-eos.rst
Yurlungur: great question. Looks like I accidentally introduced a ton of whitespace. Gonna blame my cat and just remove it lol.

singularity-eos/base/constants.hpp
Yurlungur: I *believe* so. I think our tests should catch this if there's an issue. But I can't promise that. The `UnManaged` tag just makes `Finalize` a no-op.

singularity-eos/base/spiner_table_utils.hpp
```diff
@@ -279,8 +282,74 @@ class Bounds {
...
+    int idb = 0;
+    for (auto *pdb_d : pdbs_d) {
+      auto *pdb_h = pdbs_h[idb++];
+      *pdb_d = pdb_h->getOnDevice();
+    }
+    // set memory status
+    eos_d.memoryStatus_ = DataStatus::OnDevice;
+    return eos_d;
+  }
+  static void Finalize(EOS *peos) {
```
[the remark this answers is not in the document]
Yurlungur: :+1: comment added.

test/test_eos_tabulated.cpp
Yurlungur: We will need to expose a serialization/deserialization interface on the fortran side, yes. That will need to be on the tasking for next FY.

singularity-eos/eos/eos_base.hpp
dholladay00: I'm still confused, shouldn't `SharedMemorySizeInBytes()` just return `sizeof(*this)` in the variant implementation rather than calling the held type's version?

Yurlungur: There might be some confusion as to the model here. When I call `Serialize` from either the variant or a bare class, I should get a `char` pointer pointing to serialized memory and a size. That size should be the information required to rebuild the tables and set the object, whether it be a variant or a bare class instance. The main difference here is that the variant `sizeof(*this)` and the bare class instance `sizeof(*this)` are different because the variant has a bigger memory footprint and different layout. So you need separate methods per class for dealing with the tables and then the variant and the base class have different serialization/deserialization routines for the trivially-copyable memory that account for the fact that the type is different.

`SharedMemory` is different. This is the amount of *non*-trivially copyable memory that can be written to/read from the shared allocation between MPI ranks. It's the part that's not thread local. *objects* are thread local so their metadata must be too. The only thing that goes in shared memory is tables. `SharedMemorySizeInBytes` is the same as `DynamicMemorySizeInBytes` for every object except EOSPAC. This is mainly because EOSPAC is a library. So it's serialized contents include not only tables but also metadata that are not contained in `sizeof(EOSPAC)`.

Yurlungur: Discussed with @dholladay00 offline. Resolving.

Yurlungur:
Tests pass on re-git (thanks @rbberger for fixing the CI!) and integration tested in riot downstream (thanks @AstroBarker !) So I'm calling this good and pulling the trigger.

test/test_eos_tabulated.cpp
jhp-lanl: Good point! I'll start thinking about this when the work packages are more finalized.

PR #417, opened 2024-09-11
eospac serialization fix

Yurlungur:
## PR Summary

In the previous serialization MR #410 I punted one issue with EOSPAC. This MR now fixes the inconsistencies. I test EOSPAC with shared memory in the unit tests and I update the default cmake option to assume shared memory is available.

I also had to add some link libraries (available in the standard library) that HDF5 depends on to FindHDF5 as cmake was not properly adding them. 

## PR Checklist

### DIFF
--- a/singularity-eos/eos/eos_eospac.hpp
+++ b/singularity-eos/eos/eos_eospac.hpp
@@ -1251,14 +1251,14 @@ inline std::size_t EOSPAC::SetDynamicMemory(char *src, const SharedMemSettings &
   static_assert(sizeof(char) == sizeof(EOS_CHAR), "EOS_CHAR is one byte");
   EOS_INTEGER NTABLES[] = {NT};
   EOS_INTEGER error_code = EOS_OK;
-  eosCheckError(error_code, "eos_DestroyAll", Verbosity::Debug);
 #ifdef SINGULARITY_EOSPAC_ENABLE_SHARED_MEMORY
   PORTABLE_ALWAYS_REQUIRE(
       stngs.data != nullptr,
       "EOSPAC with shared memory active requires a shared memory pointer");
-  // JMM: EOS_BOOLEAN is an enum with EOS_FALSE=0 and EOS_TRUE=1.
+  // EOS_BOOLEAN is not a type alias
+  EOS_BOOLEAN root = stngs.is_domain_root ? EOS_TRUE : EOS_FALSE;
   eos_SetSharedPackedTables(NTABLES, &packed_size_, (EOS_CHAR *)src,
-                            (EOS_CHAR *)stngs.data, stngs.is_node_root, tablehandle,
+                            (EOS_CHAR *)stngs.data, root, tablehandle,
                             &error_code);
 #else
   eos_SetPackedTables(NTABLES, &packed_size_, (EOS_CHAR *)src, tablehandle, &error_code);
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -47,12 +47,14 @@ add_executable(
   test_eos_stellar_collapse.cpp
   )
 
-add_executable(
-  closure_unit_tests
-  catch2_define.cpp
-  test_closure_pte.cpp
-  test_kpt_models.cpp
-  )
+if (SINGULARITY_BUILD_CLOSURE)
+  add_executable(
+    closure_unit_tests
+    catch2_define.cpp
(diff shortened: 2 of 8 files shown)


Yurlungur:
I also needed to more cleanly separate the closure tests as they should be disabled when closure is off.

aematts (approved):
Looks fine to my untrained eye.

test/CMakeLists.txt
```diff
@@ -47,12 +47,14 @@ add_executable(
...
-  test_kpt_models.cpp
-  )
+if (SINGULARITY_BUILD_CLOSURE)
+  add_executable(
+    closure_unit_tests
+    catch2_define.cpp
+    test_closure_pte.cpp
+    test_kpt_models.cpp
+    )
+endif()
```
jhp-lanl, line 57: When I added the closure unit tests, I was copying the approach for the tabulated unit tests and trying to simplify the CMake so that each option gets applied to its relevant executables one-by-one. Is it better to complicate the CMake logic to avoid compiling the test rather than rely on `#ifdef` statements to make sure its basically an empty test?

Put another way, in general should we be trying to push more logic into the CMake and less into the source via preprocessor flags?

singularity-eos/eos/eos_eospac.hpp
```diff
@@ -1251,15 +1251,14 @@ inline std::size_t EOSPAC::SetDynamicMemory(char *src, const SharedMemSettings &
...
   EOS_INTEGER NTABLES[] = {NT};
   EOS_INTEGER error_code = EOS_OK;
-  eosCheckError(error_code, "eos_DestroyAll", Verbosity::Debug);
 #ifdef SINGULARITY_EOSPAC_ENABLE_SHARED_MEMORY
   PORTABLE_ALWAYS_REQUIRE(
       stngs.data != nullptr,
       "EOSPAC with shared memory active requires a shared memory pointer");
-  // JMM: EOS_BOOLEAN is an enum with EOS_FALSE=0 and EOS_TRUE=1.
+  // EOS_BOOLEAN is not a type alias
+  EOS_BOOLEAN root = stngs.is_domain_root ? EOS_TRUE : EOS_FALSE;
```
jhp-lanl, line 1259: The need for this change is a bit surprising to me, but I trust you've found that the old method failed

test/test_eos_tabulated.cpp
```diff
@@ -269,12 +269,16 @@ SCENARIO("SpinerEOS and EOSPAC Serialization",
     SpinerEOSDependsRhoT rhoT_orig = SpinerEOSDependsRhoT(eosName, steelID);
     SpinerEOSDependsRhoSie rhoSie_orig = SpinerEOSDependsRhoSie(eosName, steelID);
     EOS eospac_orig = EOSPAC(steelID);
+    // not actually used but we want to stress test that we can serialize
+    // and deserialize multiple EOSPAC objects
+    EOS eospac_air = EOSPAC(airID);
```
jhp-lanl, line 274: This object appears to be used. I'm confused by this comment

jhp-lanl (approved):
Looks good! Minor questions

test/CMakeLists.txt
Yurlungur: Oh I don't know. Either is fine. I went overboard here, the tests weren't wrapped in preprocessor protection in this case or protected in the cmake so I just did both to be safe.

singularity-eos/eos/eos_eospac.hpp
Yurlungur: I might be missing something---for example a static cast might work. It appears to be an enum under the hood, so probably `EOS_TRUE` evaluates to 1 and `EOS_FALSE` to 0 but C++ booleans and C booleans aren't the same so I just decided to play it safe.

test/test_eos_tabulated.cpp
Yurlungur: Fair enough. Comment changed.

test/CMakeLists.txt
jhp-lanl: I'm pretty ignorant to the finer points of CMake so I'm happy to go either way. But if there's a superior way, let me know. @mauneyc-LANL do you have opinions?

Yurlungur: 👍 Yeah happy to revise this structure if there's one that makes sense.

PR #422, opened 2024-10-09
Draft: copy/past boiler-plate + some changes, does not compile at the moment

dholladay00:
## PR Summary
early beginnings of the pt space pte solver. It's mostly a copy-paste for now. Hopefully I can get things more cleaned up soon. Feel free to take a look @jhp-lanl 

## PR Checklist


singularity-eos/closure/mixed_cell_models.hpp
```diff
@@ -741,6 +741,242 @@ class PTESolverRhoT : public mix_impl::PTESolverBase<EOSIndexer, RealIndexer> {
...
+    // calculate an initil equilibrium pressure
+    Real psum = 0.0;
+    Real vsum = 0.0;
+    for (int m = 0; m < nmat; ++m) {
+      psum += vfrac[m]*press[m];
+      vsum += vfrac[m];
+    }
+    // all normalized temps set to 1 so no averaging necessary here.
+    Tequil = temp[0];
+    Pequil = psum / vsum;
```
jhp-lanl, line 814: This machinery has always confused me. We pass in a `Tguess` that then gets saved as `Tnorm` (unless it isn't passed in) but it seems like we always use `temp[0]` as the initial starting point for the solver, yes?

We could take the same approach with `press[0]`, but your approach does provide a reasonable guess if the pressure is evolved by the host code in some way. We will just need to note that `press` needs to contain sane values of some sort.

While I'm at it, I'll create an issue to rename `Tguess` and note that `temp` requires at least _one_ sane value.

singularity-eos/closure/mixed_cell_models.hpp
```diff
@@ -741,6 +741,242 @@ class PTESolverRhoT : public mix_impl::PTESolverBase<EOSIndexer, RealIndexer> {
...
+    Real psum = 0.0;
+    Real vsum = 0.0;
+    // is volume averaging the right thing here? 
+    for (int m = 0; m < nmat; ++m) {
+      tsum += vfrac[m]*temp[m];
+      psum += vfrac[m]*press[m];
+      vsum += vfrac[m];
+    }
+    residual[0] = Tequil - tsum/vsum;
+    residual[1] = Pequil - psum/vsum;
```
jhp-lanl, line 834: I think it's worth discussing possible convergence criteria, but your suggestion may end up being the best approach.

The residual equations are formulated based off of the density and energy sums, which is why they're included in the residuals for the other solvers. Of course in the `PTESolverRhoT` solver, we also populate the rest of the `neq - 2` entries with pressure differences, but that's because we require additional constraints in addition to the primary residual equations. Not including the energy/volume sums in the residual would mean that we're determining convergence separate from the actual residual equations themselves. This isn't _necessarily_ bad, but we need to be intentional about it.

If we imagine a situation with a small volume of metal surrounded by gas, the pressure is highly sensitive to changes in the volume fraction of metal in the mixture. The $\left(\frac{\partial P}{\partial e}\right)_V = \rho \Gamma$ derivative for the metal is likely near 10-ish, so this implies that the energy variability will be maybe 10% of the pressure variability due to volume fraction changes. If we assume the volume fraction is maybe order 1e-05, then the mass fraction might be order 1e-02, and since the energy sums with the mass fraction, I would estimate the variability to be around 1e-03 for a doubling of the pressure. The total volume fraction variability would be near zero since the bulk modulus is large.

With your residuals, the temperature residual variability would be similar to the energy residual variability above, but the pressure residual variability would be on the order of the volume fraction, or 1e-05. So this formulation might be _less_ sensitive to small volume fractions of stiff materials. The plus side of course is less iterations, but the downside might be a solid density inconsistent with the actual PTE pressure. I'm not sure that's very bad since small volume fractions of solid/metal are always annoying, and being robust to them is a good thing.

Maybe this is an opportunity though. Do you think you could add a switching variable to the class that switches between a traditional residual (energy and volume fraction sums) and this one you're proposing? I.e. keep what you have here but use something more like the first two entries of `PTESolverRhoT` when a switch is on/off?

singularity-eos/closure/mixed_cell_models.hpp
```diff
@@ -741,6 +741,242 @@ class PTESolverRhoT : public mix_impl::PTESolverBase<EOSIndexer, RealIndexer> {
...
+      error_p += residual[m + 1] * residual[m + 1];
+    }
+    error_p = std::sqrt(error_p);
+    Real error_u = std::abs(residual[1]);
+    // Check for convergence
+    bool converged_p = (error_p < pte_rel_tolerance_p * std::abs(mean_p) ||
+                        error_p < pte_abs_tolerance_p);
+    bool converged_u = (error_u < pte_rel_tolerance_e || error_u < pte_abs_tolerance_e);
+    return converged_p && converged_u;
+  }
```
jhp-lanl, line 853: It looks like this code needs to be updated to reflect the structure of the residual.

My suggestion above to include two residual formulations would complicate this function (perhaps requiring two implementations of the function actually).

singularity-eos/closure/mixed_cell_models.hpp
```diff
@@ -741,6 +741,242 @@ class PTESolverRhoT : public mix_impl::PTESolverBase<EOSIndexer, RealIndexer> {
...
+      Real dpdT = robust::ratio((p_pert - press[m]), dT);
+      rtor2_dr_dT_p_sum += rho_tot / (rho[m]*rho[m]) * dpdT / dpdv;
+      dedT_sum += robust::ratio(rhobar[m] * robust::ratio(e_pert, uscale) - u[m], dT);
+    }
+
+    // Fill in the Jacobian
+    for (int i = 0; i < neq * neq; ++i)
+      jacobian[i] = 0.0;
+    jacobian[0] = rtor2_dr_dT_p_sum;
+    jacobian[1] = rtor2_dr_dP_T_sum;
```
jhp-lanl, line 903: Make sure the sign is correct here, but let me know if there are any issues with my math too.

jhp-lanl (commented):
Great start! Let me know when you want further review

singularity-eos/closure/mixed_cell_models.hpp
jhp-lanl: See https://github.com/lanl/singularity-eos/issues/423

singularity-eos/closure/mixed_cell_models.hpp
```diff
@@ -741,6 +741,242 @@ class PTESolverRhoT : public mix_impl::PTESolverBase<EOSIndexer, RealIndexer> {
   Real Tequil, Ttemp;
 };
 
+// PT space solver
+inline int PTESolverPTRequiredScratch(const int nmat) {
+  int neq = 2;
+  return neq * neq                 // jacobian
+         + 4 * neq                 // dx, residual, and sol_scratch
+         + 3 * nmat                // all the nmat sized arrays
```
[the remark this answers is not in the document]
dholladay00: I think there is a 2nd copy of vfrac's around.

singularity-eos/closure/mixed_cell_models.hpp
dholladay00: Recall that `temp` is in arbitrary normalized temperature units and so it will always be multiplied by Tnorm.

jhp-lanl: Don't we also normalize the pressures too?

dholladay00: Yep, I think they are normalized by an internal energy.

singularity-eos/closure/mixed_cell_models.hpp
dholladay00: Should some of the checks you're talking about in the alternate scheme just be placed in the checkpte function? The residual function as I understand it is really about the unknowns of the system and the checkpte function exists for checking additional constraints AFAIK.

singularity-eos/closure/mixed_cell_models.hpp
dholladay00: I'll check again, but IIRC there is a double negative that cancels.

singularity-eos/closure/mixed_cell_models.hpp
jhp-lanl: In that case, it might make sense to add the `Tequil - tsum/vsum` and `Pequil - psum/vsum` checks to the `checkpte` function since they're not the actual residual equations being solved. The information is valuable, but I'm concerned that these aren't the actual residuals of the residual equations we're solving.

singularity-eos/closure/mixed_cell_models.hpp
jhp-lanl: I'm fine with either duplicating the temperature approach (i.e. `Pequil = press[0]`) or staying with this. Whatever we do though, it needs to be noted in the documentation what values need to be sane.

singularity-eos/closure/mixed_cell_models.hpp
dholladay00: from my reading of the document (equation 12), these exrpessions form the residual of [T* - T, P*-P].

singularity-eos/closure/mixed_cell_models.hpp
dholladay00: The pressures are initialized from a lookup so they are potentially different whereas there is 1 temperature guess calculated and it is folded into the temperature normalization and the temperature vector is initialized to 1 for every element.

singularity-eos/closure/mixed_cell_models.hpp
jhp-lanl: I'd consider equation 13 to be the residual. Equation 12 arises from the linearization of the nonlinear system, and is really just an update vector for the Newton solve.

In my experience, the convergence criterion is primarily a function of some norm of the residual itself, but it can also sometimes be related to the norm of the update when satisfying the residual equations isn't as important as finding a fixed point. In this case though, I'm concerned that volume/energy conservation is more important than all the materials being at the same temperature/pressure.

singularity-eos/closure/mixed_cell_models.hpp
jhp-lanl: > The pressures are initialized from a lookup so they are potentially different

Can you point to this code?

dholladay00: https://github.com/lanl/singularity-eos/blob/dholladay00/pt_solver/singularity-eos/closure/mixed_cell_models.hpp#L328

singularity-eos/closure/mixed_cell_models.hpp
dholladay00: yep, I'm all sorts of confused here. You are correct. I'll fix this when I get a chance.

singularity-eos/closure/mixed_cell_models.hpp
jhp-lanl: Ah okay, I'm with you now. We should definitely document this.

To summarize my understanding:
- The initial temperature guess for the _PTE solver_ is `Tequil = temp[0]`
- The provided pressure can be nonsense since it is populated by `InitBase()`
- `InitBase()` calculates the initial pressure from `Tscale` and the input density, **_NOT_** `Tequil`
- You're volume fraction averaging those pressures to get `Pequil`, which is the initial PTE state

Yurlungur:
Bits have rotted, so I had an easier time copying code piece by piece into #453, which should supersede this MR. Closing this one.

PR #433, opened 2024-11-17
nqt o2

Yurlungur:
## PR Summary

This threads in the second order NQT method developed by Peter Hammond on the AthenaK team. More details soon as I develop them. Also there will be an archive note eventually.

## PR Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -96,16 +96,17 @@ option(SINGULARITY_BETTER_DEBUG_FLAGS "Better debug flags for singularity" ON)
 option(SINGULARITY_HIDE_MORE_WARNINGS "hide more warnings" OFF)
 
 # toggle code options
-option(SINGULARITY_USE_SINGLE_LOGS
-       "Use single precision logs. Can harm accuracy." OFF)
 option(SINGULARITY_USE_TRUE_LOG_GRIDDING
-       "Use grids that conform to log spacing." OFF)
-# TODO(JMM): Should this automatically be activated when true log gridding is
-# off?
-cmake_dependent_option(
-  SINGULARITY_USE_HIGH_RISK_MATH
-  "Use integer aliased logs, may not be portable" OFF
-  "NOT SINGULARITY_USE_TRUE_LOG_GRIDDING" OFF)
+  "Use grids that conform to log spacing." OFF)
+cmake_dependent_option(SINGULARITY_USE_SINGLE_LOGS
+  "Use single precision logs. Only available for true log gridding. Can harm accuracy."
+  OFF "SINGULARITY_USE_TRUE_LOG_GRIDDING" OFF)
+option(SINGULARITY_NQT_ORDER_1
+  "In NQT logs, use first order. Faster but less accurate."
+  OFF)
+option(SINGULARITY_NQT_PORTABLE
+  "In NQT logs, use portable, rather than bithacked implementation. Slower, but more likely to function on exotic architectures."
+  OFF)
 
 # misc options
 option(SINGULARITY_FORCE_SUBMODULE_MODE "Submodule mode" OFF)
@@ -125,6 +126,26 @@ option(SINGULARITY_PATCH_MPARK_VARIANT
 set(SINGULARITY_PLUGINS "" CACHE STRING "List of paths to plugin directories")
 set(SINGULARITY_VARIANT "singularity-eos/eos/default_variant.hpp" CACHE STRING
   "The include path for the file containing the definition of singularity::EOS.")
+
+# Detect ARM architecture
+set(SINGULARITY_ON_ARM OFF CACHE BOOL "We are running on an ARM system")
+if(CMAKE_SYSTEM_PROCESSOR MATCHES "^(arm|aarch64)")
+    if (NOT SINGULARITY_USE_CUDA)
+      message(STATUS
+        "ARM architecture detected: ${CMAKE_SYSTEM_PROCESSOR}")
+      set(SINGULARITY_ON_ARM ON CACHE BOOL
+        "We are running on an ARM system")
+    endif()
+endif()
+
+if (SINGULAIRTY_ON_ARM)
+  if (NOT SINGULARITY_USE_TRUE_LOG_GRIDDING)
+    message(WARNING
+      "Fast logs not necessarily better on ARM CPU systems. "
+      "You may wish to build with "
+      "-DSINGULARITY_USE_TRUE_LOG_GRIDDING=ON.")
+  endif()
+endif()
 
 # ------------------------------------------------------------------------------#
 # singularity-eos Library
@@ -161,6 +182,15 @@ include(CMakeDetermineCCompiler)
 if(SINGULARITY_USE_FORTRAN)
   enable_language(Fortran)
   include(CMakeDetermineFortranCompiler)
+endif()
+
+# Big endianness
+include(TestBigEndian)
+TEST_BIG_ENDIAN(IS_BIG_ENDIAN)
+if (BIG_ENDIAN)
+  message(WARNING "Big endian detected! "
+    "Integer aliasing as currently implemented will not function. "
+    "Please set -DSINGULARITY_NQT_PORTABLE=ON.")
 endif()
 
 include(GNUInstallDirs)
@@ -255,16 +285,21 @@ if(SINGULARITY_BUILD_SESAME2SPINER)
 endif()
 
 # defines
-if(SINGULARITY_USE_SINGLE_LOGS)
-  target_compile_definitions(singularity-eos_Interface INTERFACE SINGULARITY_USE_SINGLE_LOGS)
-endif()
-if(SINGULARITY_USE_HIGH_RISK_MATH)
-  target_compile_definitions(singularity-eos_Interface
-                             INTERFACE SINGULARITY_USE_HIGH_RISK_MATH)
-endif()
 if (SINGULARITY_USE_TRUE_LOG_GRIDDING)
   target_compile_definitions(singularity-eos_Interface
     INTERFACE SINGULARITY_USE_TRUE_LOG_GRIDDING)
+endif()
+if(SINGULARITY_USE_SINGLE_LOGS)
+  target_compile_definitions(singularity-eos_Interface
+    INTERFACE SINGULARITY_USE_SINGLE_LOGS)
+endif()
+if(SINGULARITY_NQT_ORDER_1)
+  target_compile_definitions(singularity-eos_Interface
+    INTERFACE SINGULARITY_NQT_ORDER_1)
+endif()
+if(SINGULARITY_NQT_PORTABLE)
+  target_compile_definitions(singularity-eos_Interface
+    INTERFACE SINGULARITY_NQT_PORTABLE)
 endif()
 if(SINGULARITY_TEST_SESAME)
   target_compile_definitions(singularity-eos_Interface INTERFACE SINGULARITY_TEST_SESAME)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -284,6 +284,18 @@ if(SINGULARITY_BUILD_SESAME2SPINER)
   add_subdirectory(sesame2spiner)
 endif()
 
+# Define the full version as a string macro
+target_compile_definitions(singularity-eos_Interface INTERFACE
+    SINGULARITY_VERSION=\"${PROJECT_VERSION}\"
+)
+
+# Optionally, define major, minor, and patch versions separately
+target_compile_definitions(singularity-eos_Interface INTERFACE
+    SINGULARITY_VERSION_MAJOR=${PROJECT_VERSION_MAJOR}
+    SINGULARITY_VERSION_MINOR=${PROJECT_VERSION_MINOR}
+    SINGULARITY_VERSION_PATCH=${PROJECT_VERSION_PATCH}
+)
+
 # defines
 if (SINGULARITY_USE_TRUE_LOG_GRIDDING)
   target_compile_definitions(singularity-eos_Interface
--- a/doc/sphinx/src/contributing.rst
+++ b/doc/sphinx/src/contributing.rst
@@ -667,34 +667,42 @@ of the mantissa plus the exponent:
    \lg(x) = \lg(m) + e
 
 Therefore, if we can find a fast, invertible approximation to
-:math:`\lg(m)`, we will have achieved our goal. It turns out the
-expression
+:math:`\lg(m)`, we will have achieved our goal. The linear
+interpolation of :math:`\lg(m)` on the given interval is
 
 .. math::
 
    2 (x - 1)
 
-works pretty well, so we use that. (To convince yourself of this note
-that for :math:`x=1/2` this expression returns -1 and for :math:`x=1`,
-it returns 0, which are the correct values of :math:`\lg(x)` at the
-bounds of the interval.) Thus our approximate, invertible expression
-for :math:`\lg` is just
+and the quadratic is
 
 .. math::
 
-   2 (m - 1) + e
+  -\frac{4}{3} (m -2) (m - 1)
 
-for the mantissa and exponent extracted via ``frexp``. This differs
-from :math:`lg` by a maximum of about 0.1, which translates to at most
-a 25 percent difference. As discussed above, however, the function
-itself is an exact representation of itself and the difference from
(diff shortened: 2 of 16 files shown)


singularity-eos/base/fast-math/logs.hpp
```diff
@@ -78,76 +62,166 @@
...
+PORTABLE_FORCEINLINE_FUNCTION auto sgn(const T &x) {
+  return (T(0) < x) - (x < T(0));
+}
+
 PORTABLE_FORCEINLINE_FUNCTION
-auto as_long_int(double f) { return *reinterpret_cast<long long int *>(&f); }
+auto as_int(double f) { return *reinterpret_cast<std::int64_t *>(&f); }
 PORTABLE_FORCEINLINE_FUNCTION
-auto as_double(long long int i) { return *reinterpret_cast<double *>(&i); }
+auto as_double(std::int64_t i) { return *reinterpret_cast<double *>(&i); }
```
mauneyc-LANL, line 73: Can these (and subsequent) functions be `constexpr`?

mauneyc-LANL, line 73: Would it not be more easier to write

```c++
template<class From, class To, class = std::enable_if<some_type_restrictions<From,To>::value>::type>
To as(From f) { return *reinterpret_cast<To*>(&f); }
```

singularity-eos/base/fast-math/logs.hpp
```diff
@@ -78,76 +62,166 @@
...
+PORTABLE_FORCEINLINE_FUNCTION
+double lg_o1_aliased(const double x) {
+  PORTABLE_REQUIRE(x > 0, "log divergent for x <= 0");
   // Magic numbers constexpr because C++ doesn't constexpr reinterpret casts
   // these are floating point numbers as reinterpreted as integers.
-  // as_long_int(1.0)
-  constexpr long long int one_as_long_int = 4607182418800017408;
-  // 1./static_cast<double>(as_long_int(2.0) - as_long_int(1.0))
+  // as_int(1.0)
+  constexpr std::int64_t one_as_int = 4607182418800017408;
```
mauneyc-LANL, line 103: This is used a few times, can it be placed in a higher scope so that it's not redefined in several places? Same for a couple other constants used.

singularity-eos/base/fast-math/logs.hpp
```diff
@@ -78,76 +62,166 @@
...
+double pow2_o2_aliased(const double x) {
+  // as_int(1.0) == 2^62 - 2^52
+  constexpr std::int64_t one_as_int = 4607182418800017408;
+  // as_int(2.0) - as_int(1.0) == 2^52
   constexpr double scale_up = 4503599627370496;
-  return as_double(static_cast<long long int>(x * scale_up) + one_as_long_int);
+  constexpr std::int64_t mantissa_mask = 4503599627370495; // 2^52 - 1
+  constexpr std::int64_t a = 9007199254740992;             // 2 * 2^52
+  constexpr double b = 67108864;                           // 2^26
+  constexpr std::int64_t c = 18014398509481984;            // 4 * 2^52
```
mauneyc-LANL, line 177: Since these are all bit-shifted values, the use here would be less occluded with some explicit shifting

```c++
  // 1 << N == 2 * 1 << (N - 1)
  constexpr double scale_up = 1 << 53
  constexpr std::int64_t mantissa_mask = (1 << 53) - 1; // or scale_up - 1
  constexpr std::int64_t a = 1 << 54; // or scale_up << 1;
  constexpr double b =1 << 27; 
  constexpr std::int64_t c = 1 << 55; // or scale_up << 2;
```

singularity-eos/base/fast-math/logs.hpp
```diff
@@ -78,76 +62,166 @@
...
+#ifdef SINGULARITY_NQT_ORDER_1
+#ifdef SINGULARITY_NQT_PORTABLE
+  return lg_o1_portable(x);
 #else
-  int n;
-  assert(x > 0 && "log divergent for x <= 0");
-#ifndef SINGULARITY_USE_SINGLE_LOGS
-  // double precision is default
-  const double y = frexp(x, &n);
+  return lg_o1_aliased(x);
```
mauneyc-LANL, line 194: Since we're already here, can we move to `if constexpr`? Note that we'd probably want to translate preproccessor defines into some kind of typed boolean or enum values. You're call.

mauneyc-LANL (approved):
Comments for discussion rather than blocking requests.

singularity-eos/base/fast-math/logs.hpp
Yurlungur: :+1: good idea.

singularity-eos/base/fast-math/logs.hpp
Yurlungur: Agree that would be more readable. Let me play with it.

singularity-eos/base/fast-math/logs.hpp
Yurlungur: ah, you're thinking to enforce type width. Yes, that's a good idea.

mauneyc-LANL: Huh. Led me to `std::bit_cast` (https://en.cppreference.com/w/cpp/numeric/bit_cast). It's C++20 but does have the tempting "Possible Implementation"

```c++
template<class To, class From>
std::enable_if_t<
    sizeof(To) == sizeof(From) &&
    std::is_trivially_copyable_v<From> &&
    std::is_trivially_copyable_v<To>,
    To>
// constexpr support needs compiler magic
bit_cast(const From& src) noexcept
{
    static_assert(std::is_trivially_constructible_v<To>,
        "This implementation additionally requires "
        "destination type to be trivially constructible");
 
    To dst;
    std::memcpy(&dst, &src, sizeof(To));
    return dst;
}
```

I'm guessing ``// constexpr support needs compiler magic`` is the fun part here.

Yurlungur: lol that seems plausible. I think there's a gcc extension? At least that's what my googling led me to believe. I don't think we want to rely on that though.

mauneyc-LANL: Ah, the issue is `std::memcpy` accepts `void*` which can't be `constexpr` because the compiler isn't allowed to evaluate `void*` to an explicit type! same seems to apply to `reinterpret_cast`.

Found [this](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/p0202r0.html) (see III.B) linked on SO.

Yurlungur: Ah, which explains why it's a compiler extension but not directly in the standard. Good find!

mauneyc-LANL: I like there being "Possible Implementations" on [cppreference](https://en.cppreference.com) and don't see why it shouldn't be used in most cases, but I don't think I've come across one that was had a "not really" comment attached.

Nah, shouldn't use it. Left at the mercy of C compatibility, yet again!

Yurlungur: gonna leave this for now just for simplicity. But will add a TODO.

singularity-eos/base/fast-math/logs.hpp
Yurlungur: Done

singularity-eos/base/fast-math/logs.hpp
mauneyc-LANL: Well, `std::variant` _is_ just a fancy `union`

Yurlungur: fair point.

singularity-eos/base/fast-math/logs.hpp
Yurlungur: The proper way to do this would be to get `CMake` to emit booleans with generated code, but I got lazy and stuck `constexpr bool`s in a `Settings` namespace.

singularity-eos/base/fast-math/logs.hpp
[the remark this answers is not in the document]
Yurlungur: There are not equivalent `float` procedures at this time. Nor are there little endians. However all functions in this code are explicitly `double`s. 

You're worried about building for single-precision I assume. Yes, we should be worried about that. Let's extend that in a later PR, though. Getting this working was enough of a lift.

Yurlungur:
Timing results

# x86-volta

## NQT o1
```
                ...L2 difference = 0.0254583
                ...EOSPAC time/point (microseconds) = 0.493015
                ...spiner host time/point (microseconds) = 0.0335447
                ...spiner device time/point (microseconds) = 0.00160943
```
## NQT o2
```
                ...L2 difference = 0.00807353
                ...EOSPAC time/point (microseconds) = 0.497736
                ...spiner host time/point (microseconds) = 0.0283349
                ...spiner device time/point (microseconds) = 0.0426973
```
## NQT true
```
                ...L2 difference = 0.0171703
                ...EOSPAC time/point (microseconds) = 0.491845
                ...spiner host time/point (microseconds) = 0.0619208
                ...spiner device time/point (microseconds) = 0.00149041
```

# grace-hopper

## NQT o1
```
                ...L2 difference = 0.0254583
                ...EOSPAC time/point (microseconds) = 0.327759
                ...spiner host time/point (microseconds) = 0.0110745
                ...spiner device time/point (microseconds) = 0.000517656
```

## NQT o2
```
                ...L2 difference = 0.00807353
                ...EOSPAC time/point (microseconds) = 0.302807
                ...spiner host time/point (microseconds) = 0.0137009
                ...spiner device time/point (microseconds) = 0.000571444
```

## NQT true
```
                ...L2 difference = 0.0171703
                ...EOSPAC time/point (microseconds) = 0.326109
                ...spiner host time/point (microseconds) = 0.0203759
                ...spiner device time/point (microseconds) = 0.000564959

```

Yurlungur:
Pointwise error
copper_nqt_comp

Yurlungur:
@mauneyc-LANL I'm sorry but some compilers, notably the CI machine, don't seem to properly compute the magic numbers at constexpr time. So I'm moving them back to hardcoded magic numbers.

Yurlungur:
Timings for the stellar collapse reader
```
jonahm@sgrb:~/simulations/spiner$ tail -n 2 timings_SFHo_o1_x86_volta.txt 
host = 22.3432
device = 0.641151
jonahm@sgrb:~/simulations/spiner$ tail -n 2 timings_SFHo_o2_x86_volta.txt 
host = 31.665
device = 0.612225
jonahm@sgrb:~/simulations/spiner$ tail -n 2 timings_SFHo_true_x86_volta.txt 
host = 77.0338
device = 0.665203
jonahm@sgrb:~/simulations/spiner$ tail -n 2 timings_SFHo_o1_grace_hopper.txt 
host = 9.79919
device = 0.443848
jonahm@sgrb:~/simulations/spiner$ tail -n 2 timings_SFHo_o2_grace_hopper.txt 
host = 15.484
device = 0.453125
jonahm@sgrb:~/simulations/spiner$ tail -n 2 timings_SFHo_true_grace_hopper.txt 
host = 22.6683
device = 0.443237
```
so on GPUs, no difference. On grace, o2/true speedup is ~25% and on x86, its ~2x.

Yurlungur:
@jhp-lanl @dholladay00 @pdmullen @jdolence this is ready for review

example/eos_grid.cpp
Yurlungur, line 1: New example showing how to profile. Can be set to profile any EOS, even analytic.

CMakeLists.txt
```diff
@@ -96,16 +96,17 @@ option(SINGULARITY_BETTER_DEBUG_FLAGS "Better debug flags for singularity" ON)
...
-# off?
-cmake_dependent_option(
-  SINGULARITY_USE_HIGH_RISK_MATH
-  "Use integer aliased logs, may not be portable" OFF
-  "NOT SINGULARITY_USE_TRUE_LOG_GRIDDING" OFF)
+  "Use grids that conform to log spacing." OFF)
+cmake_dependent_option(SINGULARITY_USE_SINGLE_LOGS
+  "Use single precision logs. Only available for true log gridding. Can harm accuracy."
+  OFF "SINGULARITY_USE_TRUE_LOG_GRIDDING" OFF)
+option(SINGULARITY_NQT_ORDER_1
```
pdmullen, line 104: So this is backwards compatible with existing tables because of this, right? Should it be? Or do we want all users to now always use NQT_O2 (which admittedly would force table regeneration)?

test/test_bounds.cpp
```diff
@@ -41,7 +41,7 @@ constexpr Real T_split = 1e4;
 constexpr int N_per_decade_fine = 200;
 constexpr Real N_factor = 5;
 
-constexpr Real REAL_TOL = std::numeric_limits<Real>::epsilon() * 1e3;
+constexpr Real REAL_TOL = std::numeric_limits<Real>::epsilon() * 1e4;
```
pdmullen, line 44: Why the change here? If it was for NQT_O2, I thought that would have promoted a stricter tolerance?

test/test_math_utils.cpp
```diff
@@ -64,7 +76,7 @@ SCENARIO("Test that fast logs are invertible and run on device", "[FastMath]") {
       portableFor(
           "try out the fast math", 0, NX, PORTABLE_LAMBDA(const int i) {
             constexpr Real machine_eps = std::numeric_limits<Real>::epsilon();
-            constexpr Real acceptable_err = 100 * machine_eps;
+            constexpr Real acceptable_err = 1000 * machine_eps;
```
pdmullen, line 79: I guess my same question above applies here

singularity-eos/base/fast-math/logs.hpp
```diff
@@ -78,102 +62,216 @@
...
-  assert(x > 0 && "log divergent for x <= 0");
+  return as_double(x_as_int + a - frac_sqrt - frac_as_int + one_as_int);
+}
+// ----------------------------------------------------------------------
 
-#ifndef SINGULARITY_USE_TRUE_LOG_GRIDDING
-  // Default expression
-  return fastlg(x);
+PORTABLE_FORCEINLINE_FUNCTION
+double fastlg(const double x) {
```
pdmullen, line 213: So I was incorrectly expecting to see `sesame2spiner` to be touched by this MR... but I suspect it is not because all the logic for which log we use is handled by these `#ifdef`s.  Should `sesame2spiner` tables we marked with a header denoting which version of log they use (maybe they already are)?

pdmullen (approved):

doc/sphinx/src/contributing.rst
```diff
@@ -667,34 +667,42 @@ of the mantissa plus the exponent:
...
+provides second-order convergence in all norms, however.
+
+Finally, while ``frexp`` and ``ldexp`` are portable and performant,
+they are less performant than hand-implemented, low-level methods that
+leverage the bitwise structure of floating point numbers. These
+"bithacked" or "integer aliased" implementations are what are used in
+practice in the code.
+
+This approach is described in more detail in our `short note`_ on the
+topic.
```
AstroBarker, line 706: Nice changes to the docs. I noticed a typo, not introduced here, above `x = m2^e`: mantiss -> mantissa

CMakeLists.txt
Yurlungur: I wanted backwards compatibility for, e.g., testing. But the default is `NQT_O2`. So we should be regenerating tables.

singularity-eos/base/fast-math/logs.hpp
Yurlungur: they aren't but they should be. I will add.

test/test_bounds.cpp
Yurlungur: The sqrt in fast pow appears to introduce error at the level of roundoff, since sqrt is itself transcendental. Previously the fastpow was all algebraic primitives.

test/test_math_utils.cpp
Yurlungur: see above. It's the `sqrt` in the NQT_2 pow2 function.

doc/sphinx/src/contributing.rst
Yurlungur: Thanks for the catch!

Yurlungur:
Tests passing on re-git. Any objections to merging this? @AstroBarker lets discuss phoebus when you get the chance.

Issue #393, opened 2024-07-22
Issues with CRTP `Evaluate` on GPUs


pdmullen:
I am experiencing an issue in a downstream app wherein an `Evaluate` call (see https://github.com/lanl/singularity-eos/blob/dcd7a6efd5acbfa53903f08f02aca7abfa5c3f5f/singularity-eos/eos/eos_base.hpp#L135-L139) leads to data corruption on GPUs (`volta` with `nvhpc`).  I can seemingly only encounter the issue when working with `Spiner` EOS tables.

Printing data at addresses from explicit `ptr`s inside and outside the `Evaluate` call show differing values.  

@jonahm-LANL asked me to make an issue --- we will work on isolating the problem to see if this is a downstream app or `singularity` issue.  If the latter, I am sure a reproducer is in order...

jhp-lanl:
Which pointers? The functor? The class pointer? Or data pointers specific to spiner?

Yurlungur:
I think @pdmullen is refering to the pointers to data accessed. The call pattern that fails looks something like this:
```C++
// on device
Real *some_device_pointer = &my_device_view(0);
auto my_functor = [=](auto eos) {
  Kokkos::parallel_for(inner_range, [=](int it) {
    eos.SomeCall(pointer[i]);
  });
};
eos.Evaluate(my_functor);
```

jhp-lanl:
> I think @pdmullen is refering to the pointers to data accessed. The call pattern that fails looks something like this:
> 
> ```c++
> // on device
> Real *some_device_pointer = &my_device_view(0);
> auto my_functor = [=](auto eos) {
>   Kokkos::parallel_for(inner_range, [=](int it) {
>     eos.SomeCall(pointer[i]);
>   });
> };
> eos.Evaluate(my_functor);
> ```

I assume that should be a `KOKKOS_LAMBDA` rather than a regular lambda?

Yurlungur:
Nope because that whole call is inside a kokkos kernel already.

jhp-lanl:
> Nope because that whole call is inside a kokkos kernel already.

`Evaluate` isn't a `PORTABLE_FUNCTION`. Does it need to be if it's being called on device?

Yurlungur:
That's why `Evaluate` is marked `constexpr` that's supposed to allow us to evaluate it on device. But it's possible something is going wrong there.

jhp-lanl:
ahhhhhh gotcha

PR #434, opened 2024-11-22
Add switches for some functions to compile differently for host vs device

Yurlungur:
## PR Summary

This MR is motivated by two things:
- EOSPAC currently can't be built on HIP because EOSPAC functions are not device-capable, but we need `PORTABLE_INLINE_FUNCTION` markings.
- The `Evaluate` method in the base class needs to work on both host and device, but the `constexpr` method did not appear to fully function as desired and we were seeing segfualts with tables in riot.

Here we use the macro
```C++
#if defined(__CUDACC__) || defined(__HIPCC__)
/* My device code */
#else
/* My host code */
#endif // ON DEVICE
```
to create switches for separate implementations or prototypes depending on whether a function is called on either host or device. This unfortunately breaks the portability layer in `ports-of-call` but I don't see a way to hide `#if blah` inside an abstraction layer. So this is what we get.

@rbberger at your convenience, please try enabling EOSPAC for the RZadams CI for this MR and see if it kills it. There's a WIP PR on re-git you can use.

@pdmullen plase try to reproduce the segfault in riot that you experienced. Lets see if it still happens.

## PR Checklist


Yurlungur:
Correction, unfortunately this alternate implementation trick does not work for function prototypes. I had to split `Evaluate` into two functions, an `EvaluateHost` and an `EvaluateDevice`.

Yurlungur:
@rbberger @pdmullen this is now ready for review. `EvaluateDevice` now works and EOSPAC should how build for CUDA and HIP.

dholladay00:
> Correction, unfortunately this alternate implementation trick does not work for function prototypes. I had to split `Evaluate` into two functions, an `EvaluateHost` and an `EvaluateDevice`.

So does this mean codes that use evaluate are going to have a bunch of `#if` stuff to call the right thing? Something doesn't seem right.

singularity-eos/eos/eos_eospac.hpp
```diff
@@ -1274,43 +1274,52 @@ inline std::size_t EOSPAC::SetDynamicMemory(char *src, const SharedMemSettings &
 }
 inline std::size_t EOSPAC::SharedMemorySizeInBytes() const { return shared_size_; }
 
-SG_PIF_NOWARN
 template <typename Indexer_t>
 PORTABLE_INLINE_FUNCTION Real EOSPAC::TemperatureFromDensityInternalEnergy(
     const Real rho, const Real sie, Indexer_t &&lambda) const {
+#if defined(__CUDA_ARCH__) || __HIP_DEVICE_COMPILE__
```
dholladay00, line 1280: I'd prefer to put this logic in a single place and use it to set an ON_DEVICE macro. That way if the logic changes we have to change it in fewer places.

Yurlungur:
> > Correction, unfortunately this alternate implementation trick does not work for function prototypes. I had to split `Evaluate` into two functions, an `EvaluateHost` and an `EvaluateDevice`.
> 
> So does this mean codes that use evaluate are going to have a bunch of `#if` stuff to call the right thing? Something doesn't seem right.

I'm not sure I understand what you mean. `EvaluateHost` and `EvaluateDevice` both work regardless of the target architecture you compiled for. But the model I'm imagining now is that `EvaluateHost` is called when you want to call your own `portableFor` and `EvaluateDevice` is called from within a `portableFor`

singularity-eos/eos/eos_eospac.hpp
Yurlungur: I agree---do you know of a way to do that? can I define a "defined" macro as a preprocessor macro so I can do `#if PORTABLE_ON_DEVICE`?

singularity-eos/eos/eos_base.hpp
```diff
@@ -148,8 +148,13 @@ class EosBase {
 
   // Generic evaluator
   template <typename Functor_t>
-  constexpr void Evaluate(Functor_t &f) const {
-    CRTP copy = *(static_cast<CRTP const *>(this));
+  PORTABLE_INLINE_FUNCTION void EvaluateDevice(const Functor_t f) const {
```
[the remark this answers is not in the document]
Yurlungur: I can try that.

singularity-eos/eos/eos_eospac.hpp
dholladay00: If you do something like this (probably better ways but this works):
```c++
#if defined(__CUDA_ARCH__) || defined(__HIP_ARCH__)
#define ON_DEVICE 1
#else
#define ON_DEVICE 0
#endif
...
#if ON_DEVICE
// do stuff
#else
// do host stuff
#endif 
```

Yurlungur: Hmm. I can try that. Let me try it first in just the eospac file since that's only where we need it and we can generalize and clean up later. Seem reasonable?

Also, apparently there's no `__HIP_ARCH__` its `__HIP_DEVICE_COMPILE__`

singularity-eos/eos/eos_base.hpp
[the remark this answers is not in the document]
Yurlungur: Actually the current version works, so if you're happy with the split function signature (as discussed below) lets keep it as is. However, if you really dislike the split function signature I can take a look and try games with forwarding references, etc.

singularity-eos/eos/eos_eospac.hpp
Yurlungur: Done.

CMakeLists.txt
```diff
@@ -32,6 +32,7 @@ list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake")
 set(SINGULARITY_GOLDFILES_VERSION "goldfiles-1.8.0")
 set(SINGULARITY_GOLDFILE_HASH
     249772f3314c4b6b9386aa08895280698ae905ef188f4383b819f28c1484603b)
+set(DOWNLOAD_EXTRACT_TIMESTAMP ON)
```
jhp-lanl, line 35: I assume this is an unrelated improvement for downstream projects?

doc/sphinx/src/using-eos.rst
```diff
@@ -36,9 +36,10 @@ conditions. The type of parallelism used depends on how
 ``singularity-eos`` is compiled. If the ``Kokkos`` backend is used,
 any parallel dispatch supported by ``Kokkos`` is supported.
 
-A more generic version of the vector calls exists in the ``Evaluate``
-method, which allows the user to specify arbitrary parallel dispatch
-models by writing their own loops. See the relevant section below.
+A more generic version of the vector calls exists in the
+``EvaluateHost`` and ``EvaluateDevice`` methods, which allows the user
```
jhp-lanl, line 40:
```suggestion
``EvaluateHost`` and ``EvaluateDevice`` methods, which allow the user
```

singularity-eos/eos/eos_base.hpp
```diff
@@ -148,8 +148,13 @@ class EosBase {
...
+    const CRTP copy = *(static_cast<CRTP const *>(this));
+    f(copy);
+  }
+  template <typename Functor_t>
+  void EvaluateHost(Functor_t &f) const {
+    const CRTP copy = *(static_cast<CRTP const *>(this));
     f(copy);
```
jhp-lanl, line 158: We don't really need the explicit copy machinery for the host evaluation, do we? My memory is that this was meant to solve an issue where the object wasn't appropriately copied to device, but I could have misunderstood the issue.

That said, it's possible the compiler returns the same assembly on host regardless of how we write this.

singularity-eos/eos/eos_eospac.hpp
```diff
@@ -1274,43 +1270,52 @@ inline std::size_t EOSPAC::SetDynamicMemory(char *src, const SharedMemSettings &
 }
 inline std::size_t EOSPAC::SharedMemorySizeInBytes() const { return shared_size_; }
 
-SG_PIF_NOWARN
 template <typename Indexer_t>
 PORTABLE_INLINE_FUNCTION Real EOSPAC::TemperatureFromDensityInternalEnergy(
     const Real rho, const Real sie, Indexer_t &&lambda) const {
+#if SINGULARITY_ON_DEVICE
+  EOS_ERROR("EOSPAC calls not supported on device\n");
```
jhp-lanl, line 1277: I forget, but is there a reason to prefer `EOS_ERROR` or `PORTABLE_ERROR`?

test/test_eos_modifiers.cpp
```diff
@@ -62,7 +62,6 @@ using variadic_utils::transform_variadic_list;
 static constexpr const auto full_eos_list = tl<IdealGas>{};
 static constexpr const auto relativistic_eos_list = tl<IdealGas>{};
 static constexpr const auto unit_system_eos_list = tl<IdealGas>{};
-static constexpr const auto apply_to_all = al<ScaledEOS, ShiftedEOS>{};
```
jhp-lanl, line 65: I assume this is removed because we don't need it? I see we test shifted and scaled below

jhp-lanl (approved):

singularity-eos/eos/eos_base.hpp
jhp-lanl: I decided to see what the difference was in the canonical CRTP pattern (i.e. calling the base class function which calls the derived implementation) and this was the difference in assembly (GCC 12.3):

copy:
```
 mov    QWORD PTR [rbp-0x8],rdi
 mov    rax,QWORD PTR [rbp-0x8]
 ```
 without an explicit copy:
 ```
 mov    QWORD PTR [rbp-0x18],rdi
 lea    rax,[rbp-0x1]
 ```
 
 From what little I know of assembly, these seem to functionally be the same, right?

Feel free to modify for this exact case, but I'm satisfied that there's no real difference.
https://godbolt.org/z/3vox6on9x

Yurlungur: We do want the type to be resolved here though to short-circuit inheritance on per-device calls. Will that happen if we don't pull out the type?

CMakeLists.txt
Yurlungur: Nope it was to silence cmake, which told me to add it.

test/test_eos_modifiers.cpp
Yurlungur: Correct. Removed because of `unused variable` warning.

singularity-eos/eos/eos_base.hpp
Yurlungur: OOH I see what you mean. Just drop the `copy` var. Yes that's fine.

jhp-lanl: > OOH I see what you mean. Just drop the `copy` var. Yes that's fine.

I mean you tell me which assembly you like better. From my quick test they appear pretty similar to me (maybe `mov` is slightly better than `lea`?) so I'm fine leaving the code as you've written it.

singularity-eos/eos/eos_eospac.hpp
Yurlungur: Done.

singularity-eos/eos/eos_base.hpp
Yurlungur: Ah I see. I think the assembly's are equivalent. I prefer to leave the code as written just because it fits with the pattern of the rest of the file. But no big deal either way.

pdmullen (approved):
LGTM! My only comment is just to document here for future self that the associated symptom was not a segfault, but rather corrputed data.  This should close Issue #393 (which I think is already linked to this PR).

jhp-lanl:
> @jhp-lanl @dholladay00 any objections?

Looks good 👍

PR #437, opened 2024-11-29
Run singularity-eos through clang address sanitizer... and fix the HIP segfault!

Yurlungur:
## PR Summary

I ran singularity-eos through the clang address and undefined behavior sanitizer on my laptop. It found a few minor things like undefined variables, but it also found the source of the HIP segfault. The issue was that the constructors for the pte solvers use forwarding references, e.g.,
```C++
  template <typename EOS_t, typename Real_t, typename CReal_t, typename Lambda_t>
  PORTABLE_INLINE_FUNCTION
  PTESolverFixedT(const std::size_t nmat, EOS_t &&eos, const Real vfrac_tot,
                  const Real T_true, Real_t &&rho, Real_t &&vfrac, Real_t &&sie,
                  CReal_t &&temp, Real_t &&press, Lambda_t &&lambda, Real *scratch,
                  const MixParams &params = MixParams())
```
but the accessors in `get_sg_eos` all call the constructor by passing in accessors to 2d views, e.g., `&rho_pte(tid, 0)` which are located from the unique token logic used to prevent over-allocated scratch memory (for example). 

The problem is that the compiler is capturing these in the function call as temprorary variables, but the perfect forwarding references are then capturing them by reference, not value, meaning they go out of scope before the solver is actually used.

The issue is resolved by explicitly declaring variables for each of the accessors passed in. e.g., 
```C++
          Real *prho_pte = &rho_pte(tid, 0);
          Real *pvfrac_pte = &vfrac_pte(tid, 0);
          Real *psie_pte = &sie_pte(tid, 0);
          Real *ptemp_pte = &temp_pte(tid, 0);
          Real *ppress_pte = &press_pte(tid, 0);
          Real *pscratch = &solver_scratch(tid, 0);
          PTESolverFixedT<singularity::EOSAccessor_, Real *, Real **> method(
              npte, eos_inx, vfrac_sum, temp_pte(tid, 0), prho_pte,
              pvfrac_pte, psie_pte, ptemp_pte, ppress_pte,
              cache, pscratch);
```

We may wish to remove perfect forwarding down the line and instead insist these constructors capture by value. See #438 . 

Here's the build command I used.

```bash
cmake -DSINGULARITY_FORCE_SUBMODULE_MODE=ON -DSINGULARITY_USE_KOKKOS=ON -DSINGULARITY_BUILD_TESTS=ON -DCMAKE_CXX_COMPILER=clang++-15 -DCMAKE_CXX_FLAGS="-fsanitize=address -fsanitize=undefined -Wall -fno-omit-frame-pointer" -DCMAKE_EXPORT_COMPILE_COMMANDS=ON -DCMAKE_BUILD_TYPE=Debug ..
```

I added this to the github-workflows as a test. In the same test, I also add a strict check for warnings. If there are any warnings in the subset of code built, the test fails. Naturally, I also went through the code and eliminated said warnings. :)

## PR Checklist

### DIFF
--- a/.github/workflows/sanitizer.yml
+++ b/.github/workflows/sanitizer.yml
@@ -1,4 +1,4 @@
-name: Run clang sanitizer on minimal code subset
+name: Sanitizer
 
 on:
   push:
--- a/.github/workflows/sanitizer.yml
+++ b/.github/workflows/sanitizer.yml
@@ -1,0 +1,40 @@
+name: Run clang sanitizer on minimal code subset
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
+jobs:
+    sanitizer:
+      name: Run clang sanitizer on minimal code subset
+      runs-on: ubuntu-latest
+
+      steps:
+        - name: Checkout code
+          uses: actions/checkout@v3
+          with:
+            submodules: recursive
+        - name: Set system to non-interactive mode
+          run: export DEBIAN_FRONTEND=noninteractive
+        - name: install dependencies
+          run: |
+            sudo apt-get update -y -qq
+            sudo apt-get install -y --allow-downgrades --allow-remove-essential --allow-change-held-packages -qq build-essential clang llvm
+        - name: build and run tests
+          run: |
+            mkdir -p bin
+            cd bin
+            cmake -DCMAKE_CXX_COMPILER=clang++ \
+                  -DCMAKE_CXX_FLAGS="-fsanitize=address -fsanitize=undefined -fno-omit-frame-pointer" \
+                  -DCMAKE_BUILD_TYPE=Debug \
+                  -DSINGULARITY_STRICT_WARNINGS=ON \
+                  -DSINGULARITY_USE_FORTRAN=OFF \
+                  -DSINGULARITY_BUILD_FORTRAN_BACKEND=ON \
+                  -DSINGULARITY_BUILD_TESTS=ON \
+                  -DSINGULARITY_FORCE_SUBMODULE_MODE=ON \
+                  -DSINGULARITY_USE_KOKKOS=ON \
+                  ..
+            make -j4
+            ctest --output-on-failure
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -46,6 +46,14 @@ cmake_dependent_option(
   "SINGULARITY_USE_SPINER" OFF)
 
 option(SINGULARITY_USE_FORTRAN "Enable fortran bindings" ON)
+# Optionally build these if you want to build/test the fortran
+# infrastructure without invoking a fortran compiler If fortran is
+# off, you can set this. If fortran is on, it's forced to ON and is
+# not set-able.
+cmake_dependent_option(SINGULARITY_BUILD_FORTRAN_BACKEND
+  "Build the C++ code to which the fortran bindings bind"
+  OFF "NOT SINGULARITY_USE_FORTRAN" ON)
+
 option(SINGULARITY_USE_KOKKOS "Use Kokkos for portability" OFF)
 option(SINGULARITY_USE_EOSPAC "Enable eospac backend" OFF)
 option(SINGULARITY_EOSPAC_ENABLE_SHMEM
@@ -95,6 +103,7 @@ cmake_dependent_option(
 # modify flags options
 option(SINGULARITY_BETTER_DEBUG_FLAGS "Better debug flags for singularity" ON)
 option(SINGULARITY_HIDE_MORE_WARNINGS "hide more warnings" OFF)
+option(SINGULARITY_STRICT_WARNINGS "Make warnings strict" OFF)
 
 # toggle code options
 option(SINGULARITY_USE_TRUE_LOG_GRIDDING
@@ -552,6 +561,10 @@ target_compile_options(
          > # release
          > # cuda
 )
+if (SINGULARITY_STRICT_WARNINGS)
+  target_compile_options(singularity-eos_Interface INTERFACE
+    -Wall -Werror)
+endif()
 
 if(TARGET singularity-eos_Library)
   target_compile_options(singularity-eos_Library PRIVATE ${xlfix})
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -563,7 +563,7 @@ target_compile_options(
 )
 if (SINGULARITY_STRICT_WARNINGS)
   target_compile_options(singularity-eos_Interface INTERFACE
-    -Wall -Werror)
+    -Wall -Werror -Wno-unknown-pragmas)
 endif()
 
 if(TARGET singularity-eos_Library)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -603,6 +603,7 @@ endif()
 
 if(SINGULARITY_BUILD_TESTS)
   include(CTest)
+  enable_testing()
   add_subdirectory(test)
 endif()
 
--- a/singularity-eos/base/root-finding-1d/root_finding.hpp
+++ b/singularity-eos/base/root-finding-1d/root_finding.hpp
@@ -70,20 +70,20 @@ enum class Status { SUCCESS = 0, FAIL = 1 };
 */
 class RootCounts {
  private:
-  static constexpr int nbins_{15};
+  static constexpr std::size_t nbins_{15};
   mutable Real counts_[nbins_];
 
  public:
   PORTABLE_INLINE_FUNCTION
   RootCounts() {
-    for (int i{0}; i < nbins_; ++i)
+    for (std::size_t i{0}; i < nbins_; ++i)
       counts_[i] = 0;
   }
   PORTABLE_INLINE_FUNCTION void reset() {
-    for (int i{0}; i < nbins_; ++i)
+    for (std::size_t i{0}; i < nbins_; ++i)
       counts_[i] = 0;
   }
-  PORTABLE_INLINE_FUNCTION void increment(int i) const {
+  PORTABLE_INLINE_FUNCTION void increment(std::size_t i) const {
     assert(i < nbins_ && i >= 0);
 #ifdef PORTABILITY_STRATEGY_NONE
     counts_[i] += 1;
@@ -91,24 +91,24 @@ class RootCounts {
   }
   PORTABLE_INLINE_FUNCTION Real total() const {
     Real tot{1.e-20};
-    for (int i{0}; i < nbins_; ++i)
+    for (std::size_t i{0}; i < nbins_; ++i)
       tot += counts_[i];
     return tot;
   }
-  PORTABLE_INLINE_FUNCTION const Real &operator[](const int i) const {
+  PORTABLE_INLINE_FUNCTION const Real &operator[](const std::size_t i) const {
     assert(i < nbins_ && i >= 0);
     return counts_[i];
   }
-  PORTABLE_INLINE_FUNCTION Real &operator[](const int i) {
+  PORTABLE_INLINE_FUNCTION Real &operator[](const std::size_t i) {
     assert(i < nbins_ && i >= 0);
     return counts_[i];
   }
   PORTABLE_INLINE_FUNCTION void print_counts() const {
-    for (int i{0}; i < nbins_; ++i)
+    for (std::size_t i{0}; i < nbins_; ++i)
       printf("%e\n", counts_[i]);
   }
-  PORTABLE_INLINE_FUNCTION int nBins() const { return nbins_; }
-  PORTABLE_INLINE_FUNCTION int more() const { return nbins_ - 1; }
+  PORTABLE_INLINE_FUNCTION std::size_t nBins() const { return nbins_; }
+  PORTABLE_INLINE_FUNCTION std::size_t more() const { return nbins_ - 1; }
 };
 
 PORTABLE_INLINE_FUNCTION bool check_bracket(const Real ya, const Real yb) {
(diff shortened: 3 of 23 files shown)


rbberger:
@Yurlungur I think the issue here is that `&rho_pte(tid, 0)` is a pointer, but the memory location storing that pointer is a **temporary** if you pass it within the function call. That temporary is then passed as reference and stored as reference. However, when the call statement completes, that temporary goes out of scope and you therefore end up with a reference to a invalid memory location.

Yurlungur:
> @Yurlungur I think the issue here is that `&rho_pte(tid, 0)` is a pointer, but the memory location storing that pointer is a **temporary** if you pass it within the function call. That temporary is then passed as reference and stored as reference. However, when the call statement completes, that temporary goes out of scope and you therefore end up with a reference to an invalid memory location that was storing the pointer value.

Thanks @rbberger . Right this makes sense. If we were passing by value, this wouldn't be a problem right? Its because we use forwarding references in our accessors, which allows the compiler to bind them as references.

I wonder if this suggests that we should insist our accessors be passed by value. And we just assume reference semantics under the hood, which is how C-style arrays and views work anyway.

singularity-eos/eos/eos_gruneisen.hpp
```diff
@@ -359,14 +359,14 @@ PORTABLE_INLINE_FUNCTION Real Gruneisen::PressureFromDensityInternalEnergy(
 template <typename Indexer_t>
 PORTABLE_INLINE_FUNCTION Real
 Gruneisen::MinInternalEnergyFromDensity(const Real rho_in, Indexer_t &&lambda) const {
-  const Real rho = std::min(rho_in, _rho_max);
+  // const Real rho = std::min(rho_in, _rho_max);
```
Yurlungur, line 362: These are unused var warnings but I want to keep the vars here as a comment, for if/when we fill this in

singularity-eos/eos/eos_mgusup.hpp
```diff
@@ -370,7 +370,7 @@ MGUsup::FillEos(Real &rho, Real &temp, Real &sie, Real &press, Real &cv, Real &b
   }
   if (output & thermalqs::temperature)
     temp = TemperatureFromDensityInternalEnergy(rho, sie);
-  if (output & thermalqs::specific_internal_energy) sie = sie;
+  // if (output & thermalqs::specific_internal_energy) sie = sie;
```
Yurlungur, line 373: Silences warning from self-assignment. But I want to leave it in as a comment.

singularity-eos/eos/eos_sap_polynomial.hpp
```diff
@@ -40,8 +40,8 @@ class SAP_Polynomial : public EosBase<SAP_Polynomial> {
   SAP_Polynomial(const Real rho0, const Real a0, const Real a1, const Real a2c,
                  const Real a2e, const Real a3, const Real b0, const Real b1,
                  const Real b2c, const Real b2e, const Real b3)
-      : _rho0(rho0), _a0(a0), _a1(a1), _a2c(a2c), _a2e(a2e), _a3(a3), _b0(b0), _b1(b1),
-        _b2c(b2c), _b2e(b2e), _b3(b3) {
+      : _a0(a0), _a1(a1), _a2c(a2c), _a2e(a2e), _a3(a3), _b0(b0), _b1(b1), _b2c(b2c),
+        _b2e(b2e), _b3(b3), _rho0(rho0) {
```
Yurlungur, line 44: Silences warning about initializer list ordering.

singularity-eos/eos/get_sg_eos_rho_e.cpp
```diff
@@ -54,10 +54,18 @@ void get_sg_eos_rho_e(const char *name, int ncell, indirection_v &offsets_v,
...
+          Real *psie_pte = &sie_pte(tid, 0);
+          Real *ptemp_pte = &temp_pte(tid, 0);
+          Real *ppress_pte = &press_pte(tid, 0);
+          Real *pscratch = &solver_scratch(tid, 0);
           PTESolverRhoT<singularity::EOSAccessor_, Real *, Real **> method(
-              npte, eos_inx, vfrac_sum, sie_v(i), &rho_pte(tid, 0), &vfrac_pte(tid, 0),
-              &sie_pte(tid, 0), &temp_pte(tid, 0), &press_pte(tid, 0), cache,
-              &solver_scratch(tid, 0));
+              npte, eos_inx, vfrac_sum, sie_v(i), prho_pte, pvfrac_pte, psie_pte,
+              ptemp_pte, ppress_pte, cache, pscratch);
```
Yurlungur, line 68: This (and similar changes) silence the segfault. :) We may want to do more, though. See issue #438

CMakeLists.txt
```diff
@@ -46,6 +46,14 @@ cmake_dependent_option(
   "SINGULARITY_USE_SPINER" OFF)
 
 option(SINGULARITY_USE_FORTRAN "Enable fortran bindings" ON)
+# Optionally build these if you want to build/test the fortran
+# infrastructure without invoking a fortran compiler If fortran is
+# off, you can set this. If fortran is on, it's forced to ON and is
+# not set-able.
+cmake_dependent_option(SINGULARITY_BUILD_FORTRAN_BACKEND
+  "Build the C++ code to which the fortran bindings bind"
+  OFF "NOT SINGULARITY_USE_FORTRAN" ON)
```
Yurlungur, line 55: I added this mainly to support building with clang, because flang is still not super well supported, and I wanted to be able to run the C++ code through clang sanitizer.

CMakeLists.txt
```diff
@@ -552,6 +561,10 @@ target_compile_options(
          > # release
          > # cuda
 )
+if (SINGULARITY_STRICT_WARNINGS)
+  target_compile_options(singularity-eos_Interface INTERFACE
+    -Wall -Werror)
+endif()
```
Yurlungur, line 567: There is now a test where warnings cause a failure on the github side. :)

CMakeLists.txt
```diff
@@ -590,6 +603,7 @@ endif()
 
 if(SINGULARITY_BUILD_TESTS)
   include(CTest)
+  enable_testing()
```
Yurlungur, line 606: I don't know how tests ever ran without this. Did cmake get stricter at some point?

test/test_closure_pte.cpp
```diff
@@ -30,12 +30,12 @@
...
-constexpr Real GPa = 1.0e10;
-constexpr Real MJ_per_kg = 1.0e10;
-
 #ifdef SINGULARITY_TEST_SESAME
 #ifdef SINGULARITY_USE_SPINER_WITH_HDF5
 
+constexpr Real GPa = 1.0e10;
+constexpr Real MJ_per_kg = 1.0e10;
+
 using singularity::DavisProducts;
```
Yurlungur, line 39: Silences unused variable warning when this test is disabled.

test/test_eos_ideal.cpp
```diff
@@ -103,10 +103,10 @@ class CheckPofRE {
   int nwrong = 0;
 
  private:
-  int N_;
   Real *P_;
   Real *rho_;
   Real *sie_;
+  int N_;
```
Yurlungur, line 109: Silences initializer list warning

utils/kokkos
Yurlungur, line 1: I snuck in submodule updates to more recent versions of kokkos, ports-of-call, and kokkos-kernels. Only impacts submodule mode, but the versions we were pinned to were pretty old. This also silences some warnings from ports of call. @pdmullen this should fix the warnings you guys were seeing in artemis. :)

.github/workflows/sanitizer.yml
Yurlungur, line 1: Test will fail if the sanitizer is unhappy OR if there are any warnings in the set of code the code tests. I want to be stricter about this going forward, as I think it will help us catch problems like this in the future.

Yurlungur (commented):
Ok I went through and cleaned up more warnings. I also added the clang-sanitizer test as a github workflow. This required adding some new cmake build options to support that. I made this test also fail if it encounters any warnings, which I think will help us catch problematic code going forward. The clang sanitizer checks address/memory errors as well as looks for undefined behavior.

I also snuck in some submodule updates (which shouldn't impact the spack builds at all).

Yurlungur:
@dholladay00 @jhp-lanl @rbberger @mauneyc-LANL this is ready for review.

@jdolence this resolves the HIP segfault. I don't think it was going to be an issue for riot, but now we can be confident.

@pdmullen this should silence the warnings you were seeing in artemis.

Yurlungur:
All tests now passing on re-git including the HIP tests that were previously failing.

Yurlungur:
OK also added checks for compiler warnings from GCC, which caught a few more things, now cleaned up.

singularity-eos/base/root-finding-1d/root_finding.hpp
```diff
@@ -70,45 +70,45 @@ enum class Status { SUCCESS = 0, FAIL = 1 };
 */
 class RootCounts {
  private:
-  static constexpr int nbins_{15};
+  static constexpr std::size_t nbins_{15};
```
Yurlungur, line 73: Clang didn't care, but gcc did. Changed a bunch of `int`s to `size_t`s to avoid warnings from comparing signed vs unsigned.

singularity-eos/eos/modifiers/eos_unitsystem.hpp
```diff
@@ -35,9 +35,9 @@ using namespace eos_base;
 
 // tag dispatch for constructors for UnitSystem
 namespace eos_units_init {
-static struct ThermalUnitsInit {
+[[maybe_unused]] static struct ThermalUnitsInit {
 } thermal_units_init_tag;
-static struct LengthTimeUnitsInit {
+[[maybe_unused]] static struct LengthTimeUnitsInit {
 } length_time_units_init_tag;
```
Yurlungur, line 41: clang didn't care that these tags might be unused, maybe because it could tell they were for tag dispatch. But gcc did care. `[maybe_unused]` is a C++17 feature, so we can safely use it and assume it's compiler agnostic.

singularity-eos/base/root-finding-1d/root_finding.hpp
dholladay00: most of these seem like small-ish numbers, could we get away with `uint32_t`?

Yurlungur: We probably could. Is there a reason to do so though? This machinery is only turned on for debugging.

dholladay00: This was in regard to all of the size_t changes. Are the changes only for debug portions of code?

Yurlungur: I'd have to go through carefully. I think several are loop variables and could almost certainly be `uint32_t`. But will changing that have any real impact on memory footprint or performance?

singularity-eos/eos/eos_gruneisen.hpp
jhp-lanl: Good call!

CMakeLists.txt
jhp-lanl: I think this is a great idea! Thanks for adding it

CMakeLists.txt
```diff
@@ -552,6 +561,10 @@ target_compile_options(
          > # release
          > # cuda
 )
+if (SINGULARITY_STRICT_WARNINGS)
+  target_compile_options(singularity-eos_Interface INTERFACE
+    -Wall -Werror)
```
jhp-lanl: Are there any other warning flags we want at this point like`-Wextra` or `-Wpedantic` maybe? As I'm sure you know, `-Wall` is far from all the warnings you might want.

singularity-eos/eos/get_sg_eos_rho_e.cpp
```diff
@@ -54,10 +54,18 @@ void get_sg_eos_rho_e(const char *name, int ncell, indirection_v &offsets_v,
           // eos accessor
           singularity::EOSAccessor_ eos_inx(eos_v, &pte_idxs(tid, 0));
           // reset inputs
+          // JMM: Address sanitizer likes these named. My guess is
+          // that the forwarding references are not being resolved
+          // properly.
```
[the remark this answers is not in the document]
jhp-lanl: There are other comments like this elsewhere

jhp-lanl (approved):
Approving with some minor comments/requests.

CMakeLists.txt
Yurlungur: IMO `-Wextra` and `-Wpedantic` are too pedantic for us. `-Wextra` will catch a bunch of macros, where we've added `;` inside the macro for our own safety/sanity. `-Wpedantic` will catch variables in function signatures that are unused, just due to consistent API. 

That said, I'm open to having my mind changed.

jhp-lanl: Fair enough on those counts. I figure we can also just add more as we need them or want them.

Yurlungur: Sounds good.

singularity-eos/eos/get_sg_eos_rho_e.cpp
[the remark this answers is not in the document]
Yurlungur: Resolved. Though we may wish to change the cause of this comment down the line. (The forwarding references.)

singularity-eos/base/root-finding-1d/root_finding.hpp
Yurlungur: OK I think making sure `uint32_t` is safe everywhere we want to use it is a bit of a rat's nest. I agree it's worth doing though. I created an issue, but I'm not changing it for now. Better done in a later MR. #441

singularity-eos/eos/get_sg_eos_rho_e.cpp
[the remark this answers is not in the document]
Yurlungur: Comment is now 
```
          // JMM: The solver constructor is (deep under the hood)
          // capturing by reference. So to avoid out-of-scope access,
          // these must be "anchored" at caller scope.
```
As you say it's more than just making the sanitizer happy. This was a real bug.

Yurlungur:
Merging this now, but still open to feedback if anyone else reviews. I can submit an MR later.

Yurlungur:
Also thanks for the review @jhp-lanl !

PR #444, opened 2024-12-11
Analytic electron EOS and zsplit modifier

Yurlungur:
## PR Summary

In this MR I implement an analytic electron EOS---this is essentially an ideal gas but with a Gruneisen coefficient set by the the degrees of freedom of an electron and a specific heat that scales with ionization state divided by mean atomic mass. The idea here is that you pass in ion density and electron temperature. The ionization state is also passed in through the lambda.

I am also in the process of implementing the z split as a modifier, which splits an equation of state into a component from the ions and a component from the electrons. The z split will be two modifiers, one EOS for electrons and one for ions. Both applied to a single EOS split it into two. z split is exact for an ideal gas but not for anything else, but it is commonly used. I'm not done with this yet, so it might not be in the diff yet, depending on when you look at the MR.

Note that in the case of nuclear burning, EOS's will also need to be potentially modified by a mass scale, as per our `ScaledEOS` modifier.

I will cover adding electron or ion sub-tables of tabulated Sesame EOS's in a subsequent MR.

## PR Checklist


.github/workflows/formatting.yml
```diff
@@ -9,7 +9,7 @@ on:
 jobs:
     formatting:
       name: Check Formatting
-      runs-on: ubuntu-latest
+      runs-on: ubuntu-noble
```
Yurlungur, line 12: Latest is now 24.04 which does not support clang-format-12. For now, just pin to older Ubuntu. WIll resolve the dependency issue long term in a formal single MR in the new year. See #446

CMakeLists.txt
```diff
@@ -440,8 +440,8 @@ if(SINGULARITY_BUILD_TESTS)
     FetchContent_Declare(
       Catch2
       GIT_REPOSITORY https://github.com/catchorg/Catch2.git
-      # or later is fine too
-      GIT_TAG v3.0.1)
+      # jmm: updated Dec 17, 2024 to avoid build errors on modern gcc
+      GIT_TAG v3.7.1)
```
Yurlungur, line 444: Catch2 v3.0.1 does not compile on ubuntu 24.04. Update to 3.7.1 to rescue it.

CMakeLists.txt
```diff
@@ -560,6 +560,10 @@ target_compile_options(
          -use_fast_math
          > # release
          > # cuda
+
+         # Suppresses annoying ABI notes. See:
+         # https://stackoverflow.com/questions/52020305/what-exactly-does-gccs-wpsabi-option-do-what-are-the-implications-of-supressi
+         $<$<CXX_COMPILER_ID:GNU>:-Wno-psabi>
```
Yurlungur, line 566: GCC outputs these ABI notes. When compiling with -Werr those notes cause compilation to fail, but the notes explain gcc bugs, and are not singularity-eos problems. In the way we build singularity-eos (via spack or in-tree), ABI incompatibilities will never impact us.

singularity-eos/eos/eos_base.hpp
```diff
@@ -66,23 +66,24 @@ char *StrCat(char *destination, const char *source) {
...
-  using EosBase<EOSDERIVED>::BulkModulusFromDensityInternalEnergy;                       \
-  using EosBase<EOSDERIVED>::GruneisenParamFromDensityTemperature;                       \
-  using EosBase<EOSDERIVED>::GruneisenParamFromDensityInternalEnergy;                    \
-  using EosBase<EOSDERIVED>::FillEos;                                                    \
-  using EosBase<EOSDERIVED>::EntropyFromDensityTemperature;                              \
-  using EosBase<EOSDERIVED>::EntropyFromDensityInternalEnergy;                           \
-  using EosBase<EOSDERIVED>::GibbsFreeEnergyFromDensityTemperature;                      \
-  using EosBase<EOSDERIVED>::GibbsFreeEnergyFromDensityInternalEnergy;
+// JMM: Use VA_ARGS to capture more complex template types
+#define SG_ADD_BASE_CLASS_USINGS(...)                                                    \
```
Yurlungur, line 70: The problem here is that preprocessor macros interpret commas as separate arguments. This allows us to pass classes with multiple template arguments through the macro transparently.

singularity-eos/eos/eos_base.hpp
```diff
@@ -111,7 +112,9 @@ char *StrCat(char *destination, const char *source) {
   }                                                                                      \
   constexpr bool AllDynamicMemoryIsShareable() const {                                   \
     return t_.AllDynamicMemoryIsShareable();                                             \
-  }                                                                                      \
+  }
+
+#define SG_ADD_MODIFIER_MEAN_METHODS(t_)                                                 \
```
Yurlungur, line 117: split these out into separate modifier methods in preparation for isotopics.

singularity-eos/eos/modifiers/zsplit_eos.hpp
```diff
@@ -0,0 +1,274 @@
...
+  static constexpr const int NL = 1;
+  template <typename Indexer_t = Real *>
+  PORTABLE_FORCEINLINE_FUNCTION Real GetIonizationState_(Indexer_t &&lambda) const {
+    return std::max(0.0, lambda[t_.nlambda()]);
+  }
+  // TODO(JMM): Runtime?
+  template <typename Indexer_t = Real *>
+  PORTABLE_FORCEINLINE_FUNCTION Real GetScale_(Indexer_t &&lambda) const {
+    Real Z = GetIonizationState_(lambda);
+    if constexpr (ztype == ZSplitComponent::Electrons) {
```
Yurlungur, line 248: Notice the if constexpr here, which makes this a compile-time switch rather than runtime. We could make it runtime, but I chose not to. The trade-off here is that run-time would make the variant smaller. But compile time increases performance.

Yurlungur (commented):
@jhp-lanl @jdolence @pdmullen @chadmeyer this is now ready for review.

Yurlungur:
Github updated the latest ubuntu runner to 24.04. This breaks everything. Actions taken:
- For formatting, I pinned the runner to an older one. We'll need to resolve a formatting sweep later.
- For other tests, I massaged build system or resolved caught warnings to keep things building.
- For whatever reason gcc builds are much slower on 24.04 than previously.  Builds are timing out. I have taken the following actions:
  - github runners have 4 threads. Now all github tests use `make -j4`
  - I shrank the default variant by default, removing the relativistic eos modifier and some analytic equations of state
  - If that's not enough, I will disable Python builds from the github build.

Yurlungur:
OK in addition to the timeout, the full fortran interface was running out of memory in the linker phase. Additional actions taken:
- For github full test build, switch to `-DCMAKE_BUILD_TYPE=Release` which drops debug info
- Remove user memory limits in the docker image
- Switch to the `ld.gold` linker, which is faster and more memory efficient.

.github/workflows/docs.yml
```diff
@@ -40,7 +40,7 @@ jobs:
                   -DSINGULARITY_USE_SPINER_WITH_HDF5=OFF \
                   -DSINGULARITY_FORCE_SUBMODULE_MODE=ON \
                   -B build .
-            cmake --build build
+            cmake --build build --parallel 4
```
Yurlungur, line 43: builds the python bits faster. Speeds up doc build time by a factor of 4ish

.github/workflows/formatting.yml
```diff
@@ -9,7 +9,7 @@ on:
 jobs:
     formatting:
       name: Check Formatting
-      runs-on: ubuntu-latest
+      runs-on: ubuntu-22.04
```
Yurlungur, line 12: Needed until we update clang-format to clang-17

.github/workflows/tests.yml
Yurlungur, line 1: Changes in this file required to deal with out-of-memory failure when linking the full fortran interface.

.github/workflows/tests.yml
```diff
@@ -21,13 +21,15 @@ jobs:
         - name: install dependencies
           run: |
             sudo apt-get update -y -qq
-            sudo apt-get install -y --allow-downgrades --allow-remove-essential --allow-change-held-packages -qq build-essential gfortran libhdf5-serial-dev
+            sudo apt-get install -y --allow-downgrades --allow-remove-essential --allow-change-held-packages -qq build-essential gfortran libhdf5-serial-dev binutils-gold
```
Yurlungur, line 24: `binutils-gold` gets us use the [gold linker](https://en.wikipedia.org/wiki/Gold_(linker)) developed by google specifically for big C++ projects. It seems to be *much* faster and more memory efficient. Generically I don't know if we want to rely on it, but honestly I was very impressed.

.github/workflows/tests.yml
```diff
@@ -21,13 +21,15 @@ jobs:
...
             pip install numpy
             pip install h5py
         - name: build and run tests
           run: |
             mkdir -p bin
             cd bin
+            ulimit -m unlimited
+            ulimit -v unlimited
```
Yurlungur, line 32: Not sure if this does anything inside a docker container, but it doesn't hurt.

.github/workflows/tests.yml
```diff
@@ -38,11 +40,14 @@ jobs:
                   -DSINGULARITY_USE_HELMHOLTZ=ON \
                   -DSINGULARITY_TEST_HELMHOLTZ=ON \
                   -DSINGULARITY_FORCE_SUBMODULE_MODE=ON \
+                  -DSINGULARITY_USE_V_AND_V_EOS=OFF \
```
Yurlungur, line 43: Removes `SAP_Polynomial`, `NobleAbel` and `CarnahanStarling` equations of state from the default variant. Down the line, we may also wish to remove `StiffGas` and the relativistic modifier.

.github/workflows/tests.yml
```diff
@@ -38,11 +40,14 @@ jobs:
...
                   -DSINGULARITY_PLUGINS=$(pwd)/../example/plugin \
+                  -DCMAKE_LINKER=ld.gold \
```
Yurlungur, line 45: The gold linker in question

.github/workflows/tests.yml
```diff
@@ -38,11 +40,14 @@ jobs:
...
+                  -DCMAKE_BUILD_TYPE=Release \
```
Yurlungur, line 46: Compared to the default build type, which is `RelWithDebInfo`, this drops debug symbols, shrinking the memory footprint during link stage.

.github/workflows/tests.yml
```diff
@@ -38,11 +40,14 @@ jobs:
...
                   ..
                   #-DSINGULARITY_TEST_PYTHON=ON \
                   #-DSINGULARITY_TEST_STELLAR_COLLAPSE=ON \
                   #..
-            make
+            make -j4
```
Yurlungur, line 51: parallel build.

CMakeLists.txt
```diff
@@ -560,6 +570,13 @@ target_compile_options(
          -use_fast_math
          > # release
          > # cuda
+
+         # Suppresses annoying ABI notes. See:
+         # https://stackoverflow.com/questions/52020305/what-exactly-does-gccs-wpsabi-option-do-what-are-the-implications-of-supressi
+         $<$<AND:$<CXX_COMPILER_ID:GNU>,$<COMPILE_LANGUAGE:CXX>>:-Wno-psabi>
+         # `-Wclass-memaccess now default with -Wall but we explicitly
+         # manage this ourselves in our serialization routines.
+         $<$<AND:$<CXX_COMPILER_ID:GNU>,$<COMPILE_LANGUAGE:CXX>>:-Wno-class-memaccess>
```
Yurlungur, line 579: This is apparently a new warning. Since we sometimes build with `-Werr` its also an error. I am silencing it as `memcpy` is an intrinsic part of the serialization infrastructure and the heuristic that normally justifies warning doesn't apply in our case. We have introspection about the trivial and non-trivial copy properties of our data and we manage it.

spack-repo/packages/singularity-eos/package.py
Yurlungur, line 1: ping @rbberger just so you're aware

Yurlungur, line 1: Ping @rbberger so you're aware

Yurlungur:
Tests now passing on github and re-git.

.github/workflows/formatting.yml
Yurlungur: this is now ubuntu-22.04. Must use version number, not version name.

doc/sphinx/src/models.rst
```diff
@@ -164,6 +164,61 @@ functional forms for :math:`\Gamma` and the reference curves, the task of
...
+  Note that there is notational ambiguity between the average
+  ionization state and the average atomic number, as the symbol for
+  both is :math:`Z`. To disambiguate in ``singularity-eos``, we use
+  overbars to reference mean atomic properties such as mean atomic
+  mass :math:`\bar{A}` and mean atomic number :math:`\bar{Z}` while we
+  use :math:`\left\langle Z\right\rangle` to denote mean ionizaiton
+  state.
+
+Also by convention, the specific internal energy carried by electrons
+:math:`\varepsilon_e` is specific with respect to the *ion* mass. In
```
swjones, line 204: ion mass or total mass, which is dominated by ions?

.github/workflows/tests.yml
jhp-lanl: I'll defer to @rbberger @dholladay00 and @mauneyc-LANL on their opinions here

.github/workflows/tests.yml
jhp-lanl: I'm surprised that's such a huge memory hit to be honest. Honestly though, I find that if I _really_ need debug symbols, I'm compiling with full `Debug` anyway so I'm okay with this change.

.github/workflows/tests.yml
Yurlungur: Yeah i think for now we don't need to change it elsewhere. In this particular CI test, this, GNU ld is too memory inefficient and the test doesn't finish without gold.

.github/workflows/tests.yml
Yurlungur: Yeah. I agree. Is also only for this test. The default build is still "RelWithDebInfo."

singularity-eos/eos/modifiers/zsplit_eos.hpp
```diff
@@ -0,0 +1,274 @@
...
+    const Real iscale = GetInvScale_(lambda);
+    return scale * t_.SpecificHeatFromDensityInternalEnergy(rho, iscale * sie, lambda);
+  }
+
+  template <typename Indexer_t = Real *>
+  PORTABLE_INLINE_FUNCTION Real BulkModulusFromDensityTemperature(
+      const Real rho, const Real temperature,
+      Indexer_t &&lambda = static_cast<Real *>(nullptr)) const {
+    const Real scale = GetScale_(lambda);
+    return scale * t_.BulkModulusFromDensityTemperature(rho, temperature, lambda);
```
jhp-lanl, line 133: I _think_ this is a better approximation than for the heat capacity. I haven't looked at ionization models in great detail, but I'd initially imagine that the ionization state is a weak function of state along an isentrope.

In fact, we can estimate its importance by assuming that ionization is a pure function of temperature and ask how the temperature changes along an isentrope. Using some thermodynamic relations, the temperature changes along an isentrope as
```math
\left(\frac{\partial T}{\partial V}\right)_S = -\frac{T \alpha B_T}{C_V} = \Gamma \frac{T}{V}
```
For an ideal gas, this can be simplified to $P / C_V$, which is actually a stronger than I thought it would be. Without an explicit ionization model of course, this is impossible to take into account.

Regardless, it might be worth including a statement about the fact that the derivatives in these models do not take into account changing ionization state, which might be important in some applications.

jhp-lanl (changes requested):
Minor changes

doc/sphinx/src/models.rst
Yurlungur: Will add a comment. :+1:

.github/workflows/tests.yml
Yurlungur: I'll leave it. In worst case it does nothing.

singularity-eos/eos/modifiers/zsplit_eos.hpp
```diff
@@ -0,0 +1,274 @@
...
+  ZSplit() = default;
+  PORTABLE_INLINE_FUNCTION
+  ZSplit(T &&t) : t_(std::forward<T>(t)) {}
+
+  PORTABLE_INLINE_FUNCTION void CheckParams() const { t_.CheckParams(); }
+  auto GetOnDevice() { return ZSplit<ztype, T>(t_.GetOnDevice()); }
+  inline void Finalize() { t_.Finalize(); }
+
+  template <typename Indexer_t = Real *>
+  PORTABLE_FUNCTION Real TemperatureFromDensityInternalEnergy(
```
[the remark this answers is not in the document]
Yurlungur: Yeah so this is a fair point. There were a few reasons for this design. The first is performance (as discussed below). Fewer runtime switches is better. Another is that this is more hot-swappable. One could imagine, for example, using z-split for an ion EOS and doing something else for an electron EOS. Similarly, one might imagine composing different modifiers with different z split components.

The biggest reason, though, was from the host code perspective. This seems to be a reasonable model that is (sort of) implemented in both Flash and xrage, and it's also the model I implemented in riot.

singularity-eos/eos/modifiers/zsplit_eos.hpp
```diff
@@ -0,0 +1,274 @@
...
+    const Real scale = GetScale_(lambda);
+    const Real et = t_.InternalEnergyFromDensityTemperature(rho, temperature, lambda);
+    return scale * et;
+  }
+
+  template <typename Indexer_t = Real *>
+  PORTABLE_FUNCTION Real PressureFromDensityTemperature(
+      const Real rho, const Real temperature, Indexer_t &&lambda = nullptr) const {
+    const Real scale = GetScale_(lambda);
+    const Real Pt = t_.PressureFromDensityTemperature(rho, temperature, lambda);
```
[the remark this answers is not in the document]
Yurlungur: This model really isn't compatible with tension---it's derived from an ideal gas, after all.

I think we have to hope that in tension, the ionization state is approximately zero, so that ion and total pressure are both negative. If you have tension the zsplit will make the electron pressure negative too, but that shouldn't matter if we're not ionized.

Physically, that seems like a reasonable assumption... after all, if you're in tension, you're presumably cold enough that you're not ionized. But perhaps this means we should floor the electron pressure.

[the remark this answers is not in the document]
jhp-lanl: I've seen wonky things with ionization models (usually for gasses) where they claim to be ionized at low temperatures. A floor wouldn't be unreasonable

doc/sphinx/src/models.rst
```diff
@@ -469,6 +524,71 @@ the atomic mass and number as a final optional parameter, e.g.,
...
+
+  \gamma = \Gamma + 1 = \frac{5}{3}.
+
+The pressure is given by the number density of electrons times
+:math:`k_b T` for Boltzmann constant :math:`k_b` and temperature
+:math:`T`:
+
+.. math::
+
+  P = \frac{\rho}{m_p \bar{A}}\left\langle Z \right\rangle k_b T
```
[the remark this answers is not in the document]
Yurlungur: It's defined in line 187 above, when discussing the 3T model. I'll mention it here too though.

doc/sphinx/src/models.rst
```diff
@@ -164,6 +164,61 @@ functional forms for :math:`\Gamma` and the reference curves, the task of
 calculating a **thermodynamically consistent** temperature becomes more
 complicated.
 
+The 3T Model
+``````````````
```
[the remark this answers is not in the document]
Yurlungur: Your suggestion dropped the underline which makes it a section. But added.

doc/sphinx/src/modifiers.rst
```diff
@@ -187,6 +189,99 @@ internal energy, and temperature. On the other hand,
 specifies the unit system by specifying units for time, mass, length,
 and temperature.
 
+Z-Split EOS
+-------------
+
+For 3T physics (as described in the models section) it is often
```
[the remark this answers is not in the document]
Yurlungur: Good idea. Added.

doc/sphinx/src/modifiers.rst
```diff
@@ -187,6 +189,99 @@ internal energy, and temperature. On the other hand,
...
+
+.. code-block:: cpp
+
+  template<ZSplitComponent ztype, typename T>
+  class ZSplit
+
+where ``ZSPlitComponent`` may either be ``ZSplitComponent::Electrons``
+or ``ZSplitComponent::Ions``. As syntactic sugar, ``ZSplitE<T>`` and
+``ZSplitI<T>`` are available. The Z-split constructor does not require
+any additional parameters, so you may construct one as, e.g.,
```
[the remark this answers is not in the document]
Yurlungur: `ZSplit` uses the mean atomic properties of the parent EOS. I will make this explicit and also add the assume hydrogen comment.

singularity-eos/eos/modifiers/zsplit_eos.hpp
```diff
@@ -0,0 +1,274 @@
...
+    const Real scale = GetScale_(lambda);
+    const Real et = t_.MinInternalEnergyFromDensity(rho, lambda);
+    return scale * et;
+  }
+  template <typename Indexer_t = Real *>
+  PORTABLE_INLINE_FUNCTION Real
+  EntropyFromDensityTemperature(const Real rho, const Real temperature,
+                                Indexer_t &&lambda = static_cast<Real *>(nullptr)) const {
+    const Real scale = GetScale_(lambda);
+    return scale * t_.EntropyFromDensityTemperature(rho, temperature, lambda);
```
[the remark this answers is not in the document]
Yurlungur: Thanks for double checking! We may wish to revisit this at some point.

singularity-eos/eos/modifiers/zsplit_eos.hpp
```diff
@@ -0,0 +1,274 @@
...
+    const Real iscale = GetInvScale_(lambda);
+    return scale * t_.EntropyFromDensityInternalEnergy(rho, iscale * sie, lambda);
+  }
+
+  template <typename Indexer_t = Real *>
+  PORTABLE_INLINE_FUNCTION Real SpecificHeatFromDensityTemperature(
+      const Real rho, const Real temperature,
+      Indexer_t &&lambda = static_cast<Real *>(nullptr)) const {
+    const Real scale = GetScale_(lambda);
+    return scale * t_.SpecificHeatFromDensityTemperature(rho, temperature, lambda);
```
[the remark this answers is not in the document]
Yurlungur: Ah you're right---but we don't have access to $d scale / d T$ because it depends on the ionization model. I will add a warning in the documentation.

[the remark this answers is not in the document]
Yurlungur: I also added a TODO comment in the code describing the issue.

singularity-eos/eos/modifiers/zsplit_eos.hpp
Yurlungur: :+1: Clarified in the documentation with a warning and added a comment inline in the code.

After we start thinking more carefully about ionization models, I wonder if we could include this information inline in the sense that we could maybe make zsplit depend on an ionization model with a template parameter or something similar. Food for thought down the road.

doc/sphinx/src/models.rst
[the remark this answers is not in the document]
jhp-lanl: Whoops! Sorry about that

Yurlungur:
Thanks @jhp-lanl and @pdmullen for the reviews! I think all comments should be addressed (as best they currently can with available physics fidelity). Regarding thermodynamic derivatives... I wonder if we can do better down the line. Something to consider once we have access to an ionization model.

doc/sphinx/src/models.rst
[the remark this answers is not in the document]
Yurlungur: no problem. Just explaining why I implemented it by hand, rather than clicking "commit suggestion" :)

singularity-eos/eos/modifiers/zsplit_eos.hpp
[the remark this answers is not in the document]
jhp-lanl: Yeah it's just something to keep in mind for the future when we're implementing ionization models. I'm not quite sure how the ionization temperature dependence could be communicated, but we should have that in the back of our minds.

jhp-lanl (approved):

singularity-eos/eos/modifiers/zsplit_eos.hpp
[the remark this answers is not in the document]
pdmullen: Gross.

jhp-lanl:
> Regarding thermodynamic derivatives... I wonder if we can do better down the line. Something to consider once we have access to an ionization model.

I completely agree. Not something to tackle now, but definitely something to keep in mind

singularity-eos/eos/modifiers/zsplit_eos.hpp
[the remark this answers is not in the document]
pdmullen: _maybe_ the performance concern be avoided via

```c++
tion = eos.TemperatureFromDensityInternalEnergy<ZSplitComponent::Ions>(rho, sie);
```

and `constexpr`, but as aforementioned, I still think I prefer to have unique `eos` objects as you have done.  Thanks for talking me through your thought process.

Yurlungur:
Thanks all! Merged. 🎊

PR #449, opened 2024-12-24
Make sure that DensityEnergyFromPressureTemperature works for all equations of state

Yurlungur:
## PR Summary

The easiest way to get the Jacobian for the PT-space PTE solver we are building is to ensure that the `DensityEnergyFromPressureTemperature` call works consistently for all equations of state. This MR makes sure thats the case and threads this call through all the tests. 

To ease this process, I added a base class implementation. However, I think most classes will want their own implementations, and I added them when appropriate.

## PR Checklist

### DIFF
--- a/.github/workflows/tests_minimal.yml
+++ b/.github/workflows/tests_minimal.yml
@@ -36,5 +36,4 @@ jobs:
                   ..
             make -j4
             make install
-            make test
-            ctest --output-on-failure --rerun-failed
+            ctest --output-on-failure
--- a/.github/workflows/tests_minimal.yml
+++ b/.github/workflows/tests_minimal.yml
@@ -37,3 +37,4 @@ jobs:
             make -j4
             make install
             make test
+            ctest --output-on-failure --rerun-failed
--- a/doc/sphinx/src/using-eos.rst
+++ b/doc/sphinx/src/using-eos.rst
@@ -1172,3 +1172,35 @@ Similarly,
 
 returns the density at which pressure is minimized for a given
 temperature. This is again useful for root finds.
+
+The function
+
+.. code-block:: cpp
+
+  template <typename Indexer_t = Real*>
+  void DensityEnergyFromPressureTemperature(const Real press, const Real temp,
+                                            Indexer_t &&lambda, Real &rho, Real &sie) const;
+
+is designed for working in Pressure-Temperature space. Given a
+pressure ``press`` and temperature ``temp``, it sets a density ``rho``
+and specific internal energy ``sie``.
+
+.. note::
+
+  Note that ``lambda`` must be passed in, whether or not a given
+  equation of state requires one. You may pass in ``nullptr`` safely,
+  however.
+
+Typically this operation requires a root find. You may pass in an
+initial guess for the density ``rho`` in-place and most EOS models
+will use it.
+
+.. warning::
+
+  Pressure is not necessarily monotone in density and it may be double
+  valued. Thus you are not guaranteed to find the correct root and the
+  value of your initial guess may determine correctness. The fact that
+  ``rho`` may be used as an initial guess means you **must** pass in
+  an initialized variable, even if it is zero-initialized. Do not pass
+  uninitialized memory into this function!
+
--- a/doc/sphinx/src/using-eos.rst
+++ b/doc/sphinx/src/using-eos.rst
@@ -1171,7 +1171,18 @@ Similarly,
 .. cpp:function:: Real RhoPmin(const Real temp) const;
 
 returns the density at which pressure is minimized for a given
-temperature. This is again useful for root finds.
+temperature. The function
+
+.. cpp:function:: Real MinimumPressure() const;
+
+provides the minimum pressure an equation of state supports, which may
+be the most negative tension state. The function
+
+.. cpp:function:: Real MaximumPressureFromTemperature() const;
+
+provides a maximum possible pressure an equation of state
+supports. (Most models are unbounded in pressure.) This is again
+useful for root finds.
 
 The function
 
--- a/doc/sphinx/src/using-eos.rst
+++ b/doc/sphinx/src/using-eos.rst
@@ -1194,13 +1194,8 @@ The function
 
 is designed for working in Pressure-Temperature space. Given a
 pressure ``press`` and temperature ``temp``, it sets a density ``rho``
-and specific internal energy ``sie``.
-
-.. note::
-
-  Note that ``lambda`` must be passed in, whether or not a given
-  equation of state requires one. You may pass in ``nullptr`` safely,
-  however.
+and specific internal energy ``sie``. The ``lambda`` is optional and
+defaults to a ``nullptr``.
 
 Typically this operation requires a root find. You may pass in an
 initial guess for the density ``rho`` in-place and most EOS models
--- a/doc/sphinx/src/using-eos.rst
+++ b/doc/sphinx/src/using-eos.rst
@@ -1143,16 +1143,30 @@ quantities as outputs.
 Methods Used for Mixed Cell Closures
 --------------------------------------
 
-Several methods were developed in support of mixed cell closures. In particular:
+Several methods were developed in support of mixed cell closures. In particular the function
 
 .. cpp:function:: Real MinimumDensity() const;
 
-and 
+the function
 
 .. cpp:function:: Real MinimumTemperature() const;
 
+and the function
+
+.. cpp:function:: Real MaximumDensity() const;
+
 provide bounds for valid inputs into a table, which can be used by a
-root finder to meaningful bound the root search. Similarly,
+root finder to meaningful bound the root search.
+
+.. warning::
+
+  For unbounded equations of state, ``MinimumDensity`` and
+  ``MinimumTemperature`` will return zero, while ``MaximumDensity``
+  will return a very large finite number. Which number you get,
+  however, is not guaranteed. You may wish to apply more sensible
+  bounds in your own code.
+
+Similarly,
 
 .. cpp:function:: Real RhoPmin(const Real temp) const;
 
--- a/doc/sphinx/src/using-eos.rst
+++ b/doc/sphinx/src/using-eos.rst
@@ -1178,11 +1178,11 @@ temperature. The function
 provides the minimum pressure an equation of state supports, which may
 be the most negative tension state. The function
 
-.. cpp:function:: Real MaximumPressureFromTemperature() const;
+.. cpp:function:: Real MaximumPressureAtTemperature(const Real temp) const;
 
-provides a maximum possible pressure an equation of state
-supports. (Most models are unbounded in pressure.) This is again
-useful for root finds.
+provides a maximum possible pressure an equation of state supports at
+a given temperature. (Most models are unbounded in pressure.) This is
+again useful for root finds.
 
 The function
 
--- a/singularity-eos/eos/eos_base.hpp
+++ b/singularity-eos/eos/eos_base.hpp
@@ -869,8 +869,8 @@ class EosBase {
     auto PofRT = [&](const Real r) {
       return copy.PressureFromDensityTemperature(r, temp, lambda);
     };
-    Real rhoguess = rho; // use input density
-    if ((rhoguess < rhomin) || (rhoguess > rhomax)) {
+    Real rhoguess = rho;                                // use input density
+    if ((rhoguess <= rhomin) || (rhoguess >= rhomax)) { // avoid edge effects
       if ((rhomin < DEFAULT_RHO_GUESS) && (DEFAULT_RHO_GUESS < rhomax)) {
         rhoguess = DEFAULT_RHO_GUESS;
       } else {
--- a/singularity-eos/eos/eos_base.hpp
+++ b/singularity-eos/eos/eos_base.hpp
@@ -856,10 +856,10 @@ class EosBase {
     using RootFinding1D::findRoot; // more robust but slower. Better default.
     using RootFinding1D::Status;
 
-    // Pressure is not monotone in density at low densities, due to a
-    // phase transition, which can prevent convergence. We want to
-    // approach tension from above, not below. Choose close to, but
-    // above, normal density for a metal like copper.
+    // Pressure is not monotone in density at low densities, which can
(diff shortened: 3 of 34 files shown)


.github/workflows/tests_minimal.yml
```diff
@@ -36,4 +36,4 @@ jobs:
                   ..
             make -j4
             make install
-            make test
+            ctest --output-on-failure
```
Yurlungur, line 39: I think I'd like to leave this permanently. You still get a summary out at the end, but here we can see what went wrong with a test. Useful for debugging, especially when the CI machine and your laptop disagree...

singularity-eos/eos/eos_base.hpp
```diff
@@ -836,6 +838,44 @@ class EosBase {
     PORTABLE_ALWAYS_THROW_OR_ABORT(msg);
   }
 
+  // JMM: This method is often going to be overloaded for special cases.
+  template <typename Indexer_t = Real *>
+  PORTABLE_INLINE_FUNCTION void
+  DensityEnergyFromPressureTemperature(const Real press, const Real temp,
+                                       Indexer_t &&lambda, Real &rho, Real &sie) const {
```
Yurlungur, line 845: This default implementation is usually not optimal but does seem to work if a given EOS doesn't need special cases. In general, though, an EOS should overload and implement its own.

singularity-eos/eos/eos_base.hpp
```diff
@@ -836,6 +838,44 @@ class EosBase {
...
+      } else {
+        rhoguess = 0.5 * (rhomin + rhomax);
+      }
+    }
+    auto status = findRoot(PofRT, press, rhoguess, rhomin, rhomax, EPS, EPS, rho);
+    // JMM: This needs to not fail and instead return something sane
+    if (status != Status::SUCCESS) {
+      PORTABLE_WARN("DensityEnergyFromPressureTemperature failed to find root\n");
+      rho = rhoguess;
+    }
```
Yurlungur, line 874: This is a function that may be passed data and I see it as similar to a PTE solve. Even if it fails out, something reasonable must be done.

singularity-eos/eos/eos_helmholtz.hpp
```diff
@@ -676,16 +676,37 @@ class Helmholtz : public EosBase<Helmholtz> {
   ValuesAtReferenceState(Real &rho, Real &temp, Real &sie, Real &press, Real &cv,
                          Real &bmod, Real &dpde, Real &dvdt,
                          Indexer_t &&lambda = static_cast<Real *>(nullptr)) const {
-    // JMM: I'm not sure what to put here or if it matters. Some
-    // reference state, maybe stellar denity, would be appropriate.
-    PORTABLE_ALWAYS_ABORT("Stub");
+    // TODO(JMM): Is this right???
+    rho = 1e7;
+    temp = 1.5e9;
```
Yurlungur, line 681: Conditions for an oxygen burning shell in a stellar core.

singularity-eos/eos/eos_mgusup.hpp
```diff
@@ -125,6 +125,47 @@ class MGUsup : public EosBase<MGUsup> {
...
+  Real MinimumDensity() const { return 10 * robust::EPS(); }
+
+  PORTABLE_FORCEINLINE_FUNCTION
+  Real MaximumDensity() const {
+    if (_s > 1) {
+      return 0.99 * robust::ratio(_s * _rho0, _s - 1);
+    } else { // for s <= 1, no maximum, but we need to pick something.
+      return 1e3 * _rho0;
+    }
+  }
```
Yurlungur, line 139: These are the conditions where the reference isotherm is a positive temperature.

singularity-eos/eos/eos_mgusup.hpp
```diff
@@ -125,6 +125,47 @@ class MGUsup : public EosBase<MGUsup> {
   FillEos(Real &rho, Real &temp, Real &energy, Real &press, Real &cv, Real &bmod,
           const unsigned long output,
           Indexer_t &&lambda = static_cast<Real *>(nullptr)) const;
+
+  PORTABLE_FORCEINLINE_FUNCTION
+  Real MinimumDensity() const { return 10 * robust::EPS(); }
```
Yurlungur, line 130: Since the reference isotherm scales as _rho0/rho, the EOS diverges at zero density. This bounds it at least to some degree.

singularity-eos/eos/eos_stellar_collapse.hpp
```diff
@@ -653,7 +653,25 @@ PORTABLE_INLINE_FUNCTION Real StellarCollapse::GruneisenParamFromDensityInternal
...
+  Real lT = lT_(temp);
+  Real lP = P2lP_(press);
+  Real Ye = lambda[Lambda::Ye];
+
+  if ((lrguess < lRhoMin_) || (lrguess > lRhoMax_)) {
+    lrguess = lRho_(rhoNormal_);
+  }
+  auto lPofRT = [&](Real lR) { return lP_.interpToReal(Ye, lT, lR); };
+  auto status = regula_falsi(lPofRT, lP, lrguess, lRhoMin_, lRhoMax_, ROOT_THRESH,
+                             ROOT_THRESH, lrguess);
```
Yurlungur, line 669: Since the EOS is tabulated in logrho, use logrho for the root find.

singularity-eos/eos/eos_vinet.hpp
```diff
@@ -131,6 +131,10 @@ class Vinet : public EosBase<Vinet> {
   ValuesAtReferenceState(Real &rho, Real &temp, Real &sie, Real &press, Real &cv,
                          Real &bmod, Real &dpde, Real &dvdt,
                          Indexer_t &&lambda = static_cast<Real *>(nullptr)) const;
+
+  PORTABLE_FORCEINLINE_FUNCTION
+  Real MinimumDensity() const { return std::cbrt(10 * robust::EPS()) * _rho0; }
```
Yurlungur, line 136: EOS depends on $\left(\frac{\rho_0}{\rho}\right)^{1/3}$ so a reasonable bound before the EOS will become ill behaved is this.

singularity-eos/eos/eos_vinet.hpp
```diff
@@ -152,11 +156,6 @@ class Vinet : public EosBase<Vinet> {
     }
     printf("\n\n");
   }
-  // Density/Energy from P/T not unique, if used will give error
-  template <typename Indexer_t = Real *>
-  PORTABLE_INLINE_FUNCTION void
-  DensityEnergyFromPressureTemperature(const Real press, const Real temp,
-                                       Indexer_t &&lambda, Real &rho, Real &sie) const;
```
Yurlungur, line 159: use the default in the base class

test/eos_unit_test_helpers.hpp
```diff
@@ -143,6 +144,34 @@ inline void compare_two_eoss(const E1 &test_e, const E2 &ref_e) {
   return;
 }
 
+template <typename EOS, typename Indexer_t = Real *>
+PORTABLE_INLINE_FUNCTION bool
+CheckRhoSieFromPT(EOS eos, Real rho, Real T,
+                  Indexer_t &&lambda = static_cast<Real *>(nullptr)) {
```
Yurlungur, line 150: Convenience function to let me test the functionality more easily

Yurlungur:
Ok that was finicky but tests on github now seem to be passing. re-git tests appear to be having some network issues. Ping @rbberger for after the holidays. :)

singularity-eos/eos/eos_base.hpp
```diff
@@ -836,6 +838,44 @@ class EosBase {
...
+  // JMM: This method is often going to be overloaded for special cases.
+  template <typename Indexer_t = Real *>
+  PORTABLE_INLINE_FUNCTION void
+  DensityEnergyFromPressureTemperature(const Real press, const Real temp,
+                                       Indexer_t &&lambda, Real &rho, Real &sie) const {
+    // TODO(JMM): A lot hardcoded in here... Hopefully relevent EOS's
+    // overwrite.
+    constexpr Real MAXFAC = 1e8;
+    constexpr Real EPS = 10 * robust::EPS();
+    constexpr Real MINR_DEFAULT = 1e-4;
```
jhp-lanl, line 850: Is this a meaningful value? If not, I'd pick something more like 1e-08. That _should_ be low enough for most sesame tables.

singularity-eos/eos/eos_helmholtz.hpp
jhp-lanl: Can  you comment this in the code?

singularity-eos/eos/eos_mgusup.hpp
jhp-lanl: Honestly the $s < 0$ case implies that the fundamental shock derivative is negative and the EOS is no longer convex, so it's not a very useful EOS. I think what you picked is as good as any for that case.

singularity-eos/eos/eos_mgusup.hpp
```diff
@@ -125,6 +125,47 @@ class MGUsup : public EosBase<MGUsup> {
...
+    if (_s > 1) {
+      return 0.99 * robust::ratio(_s * _rho0, _s - 1);
+    } else { // for s <= 1, no maximum, but we need to pick something.
+      return 1e3 * _rho0;
+    }
+  }
+
+  template <typename Indexer_t = Real *>
+  PORTABLE_INLINE_FUNCTION void
+  DensityEnergyFromPressureTemperature(const Real press, const Real temp,
```
jhp-lanl, line 143: Why not just use the base class version?

singularity-eos/eos/eos_mgusup.hpp
```diff
@@ -200,6 +236,8 @@ PORTABLE_INLINE_FUNCTION Real MGUsup::HugTemperatureFromDensity(Real rho) const
   Real eta = 1.0 - robust::ratio(_rho0, rho);
   Real f1 = 1.0 - _s * eta;
   if (f1 <= 0.0) {
+    printf("f1, eta, rho, rho0, s = %.14e %.14e %.14e %.14e %.14e\n", f1, eta, rho, _rho0,
+           _s);
```
jhp-lanl, line 240: Good addition!

jhp-lanl (approved):

singularity-eos/eos/eos_helmholtz.hpp
Yurlungur: :+1: done

singularity-eos/eos/eos_mgusup.hpp
Yurlungur: :+1: I will add a comment to this effect.

singularity-eos/eos/eos_mgusup.hpp
Yurlungur: Because of the maximum density, which is a runtime parameter for eos_mgusup but not the others. In the future we might want introspection for max density though. I'll add a comment explaining and a TODO to remove this when max density is available.

Yurlungur: Actually, we should just provide a max density that can be overwritten. I will add that.

singularity-eos/eos/eos_base.hpp
Yurlungur: It's not meaningful. I think I can actually make these hardcoded constants go away by being a bit more clever.

Yurlungur:
OK... that took longer than I thought it would. All tests now passing once again and the machinery is more flexible and robust. @jhp-lanl please re-review when you get the chance. I believe I've addressed all your concerns. 

I also noticed that the `DensityEnergyFromPressureTemperature` function wasn't documented. It's documented now.

Yurlungur:
Ping @jhp-lanl

jhp-lanl:
> Ping @jhp-lanl

I'll try to re-review that by the end of today.

Yurlungur:
Snuck in introspection for pressure into this MR as well.

jhp-lanl (approved):
Approving because none of my comments are really blocking. That said, I do think the name change to `MaximumPressureAtTemperature()` would be preferable. I also think the documentation should be changed to indicate that there _could_ be a temperature dependence for the maximum pressure if an EOS supports it.

doc/sphinx/src/using-eos.rst
```diff
@@ -1143,18 +1143,75 @@ quantities as outputs.
...
+.. cpp:function:: Real MinimumPressure() const;
+
+provides the minimum pressure an equation of state supports, which may
+be the most negative tension state. The function
+
+.. cpp:function:: Real MaximumPressureFromTemperature() const;
+
+provides a maximum possible pressure an equation of state
+supports. (Most models are unbounded in pressure.) This is again
+useful for root finds.
```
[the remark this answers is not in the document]
Yurlungur: I completely agree--I change the name. For the tabulated equations of state, I actually do pull it out and record it now (thinking ahead), but I let this float as the table interpolation does extrapolate off the top of the table. It's not particularly trustworthy, but it's not a hard limit. Only Gruneisen currently reports a maximum pressure bound. On the other hand, I do have each table report the minimum pressure, which is important.

doc/sphinx/src/using-eos.rst
```diff
@@ -1143,18 +1143,75 @@ quantities as outputs.
...
+
+is designed for working in Pressure-Temperature space. Given a
+pressure ``press`` and temperature ``temp``, it sets a density ``rho``
+and specific internal energy ``sie``.
+
+.. note::
+
+  Note that ``lambda`` must be passed in, whether or not a given
+  equation of state requires one. You may pass in ``nullptr`` safely,
+  however.
```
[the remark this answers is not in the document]
Yurlungur: I was going to say it's because of the order of the arguments---you can't have a default argument in the middle. But you know what? There's no reason we can't just add an overload in the base class. Done.

singularity-eos/eos/eos_eospac.hpp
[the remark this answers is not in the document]
Yurlungur: Agree this is a concern. I made an issue, #454, but will punt for now.

Yurlungur:
Thanks for the review, @jhp-lanl ! Agree with your comments and have addressed them. Will merge once tests pass on re-git.

PR #461, opened 2025-02-05
Fixed EOSPAC.cmake Install Directory

tuxfan:
The existing install location is not added to the `CMAKE_MODULE_PATH`. Installing into `lib/cmake/singularity-eos` fixes this issue, as that location is already in the search path.


Yurlungur (approved):
Ah I see. This is just a bug introduced when we re organized the cmake folder. Thanks for the catch and fix!

PR #466, opened 2025-02-13
thread lambda through pte

Yurlungur:
## PR Summary

I discovered that we never fully threaded lambda arguments through the PTE solver. This MR properly threads them through. To test it, I also switch the test_pte_ideal test from using ideal gas to using IdealElectrons, which adds code coverage there as well.

This change is needed to support EOS's that use the lambda, notably partially ionized EOS's, with PTE.

## PR Checklist

### DIFF
--- a/test/test_pte_ideal.cpp
+++ b/test/test_pte_ideal.cpp
@@ -35,13 +35,20 @@
 #include <test/eos_unit_test_helpers.hpp>
 
 using singularity::ApproxTemperatureFromRhoMatU;
-using singularity::IdealGas;
+using singularity::IdealElectrons;
+using singularity::MeanAtomicProperties;
 using singularity::PTESolverPT;
 using singularity::PTESolverPTRequiredScratch;
 using singularity::PTESolverRhoT;
 using singularity::PTESolverRhoTRequiredScratch;
 using singularity::Variant;
-using EOS = Variant<IdealGas>;
+using EOS = Variant<IdealElectrons>;
+
+struct LambdaIndexer {
+  Real *operator[](const int i) { return &z; }
+  const Real *operator[](const int i) const { return &z; }
+  Real z = 0.9;
+};
 
 template <typename RealIndexer, typename EOSIndexer>
 PORTABLE_INLINE_FUNCTION Real set_state(Real rho_nom, Real sie_nom, RealIndexer &&rho,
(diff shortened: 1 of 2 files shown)


test/test_pte_ideal.cpp
```diff
@@ -35,13 +35,20 @@
 #include <test/eos_unit_test_helpers.hpp>
 
 using singularity::ApproxTemperatureFromRhoMatU;
-using singularity::IdealGas;
```
pdmullen, line 38: Should we instead just add a `test_pte_ideal_electrons.cpp`, thereby keeping the `IdealGas` test pathway?

pdmullen (approved):

test/test_pte_ideal.cpp
Yurlungur: I don't want to write an almost identical test... but I could potentially play some template games to run both tests in this file.

singularity-eos/eos/get_sg_eos_rho_p.cpp
```diff
@@ -76,7 +76,7 @@ void get_sg_eos_rho_p(const char *name, int ncell, indirection_v &offsets_v,
           // calculate sie from single eos
           auto p_from_t = [&](const Real &t_i) {
             return eos_v(pte_idxs(tid, 0))
-                .PressureFromDensityTemperature(rho_pte(tid, 0), t_i, cache[0]);
+                .PressureFromDensityTemperature(rho_pte(tid, 0), t_i, cache);
```
dholladay00, line 79: shouldn't this be cache[0] since this is the single material case and it is expecting the indexer of a material (the only material in this case)?

Yurlungur: oops yes that one should be. Good catch.

dholladay00: Actually I have been thinking about this a bit more. While this is correct in the vast majority of cases where `npte == nmat == 1`, there could be cases where `1 == npte < nmat`. In this case we need to use the correct offsetting index for the density and the cache. Maybe this should be its own issue and in a separate PR, but thought I should point it out.

Yurlungur: I'll create an issue. This code will also need to change in the same way I changed the C++ code to support non-trivial lambdas for, e.g., partial ionization at some point.

dholladay00: all of the pte solver files likely suffer from this bug for the single material case. Wonder how many times its caused issues?

These are my fault, sorry about that this.

Yurlungur: I think it's harmless right now, since the cache isn't filled with anything anyway for the EOS's used. But when threading through real lambdas, it will need to change. #467

Yurlungur: > all of the pte solver files likely suffer from this bug for the single material case. Wonder how many times its caused issues?
> 
> These are my fault, sorry about that this.

I will double check---I think it's not so bad right now because the cache is only used as an initial guess.

dholladay00: I think the rho index is also potentially incorrect though as well.

jhp-lanl (approved):

PR #468, opened 2025-02-18
move def_en and def_v to CPP file to fix minor linking issue with singularity-eos-extra

Yurlungur:
## PR Summary

When adding a fortran interface to singularity-eos-extra, the `singularity_eos_init_utils.hpp` header is needed. However, it defines `def_en` and `def_v` in global scope in a header file. This produces linking errors when using plugins. Here I resolve the linking error by simply moving the *definition* of these arrays into `singularity_eos.cpp` but still make them "available" by declaring them as `extern` in the original header.

@dholladay00 should I declare them constant?

## PR Checklist

### DIFF
--- a/singularity-eos/eos/singularity_eos.cpp
+++ b/singularity-eos/eos/singularity_eos.cpp
@@ -19,6 +19,11 @@
 #include <singularity-eos/eos/singularity_eos.hpp>
 #include <singularity-eos/eos/singularity_eos_init_utils.hpp>
 
+namespace singularity {
+int def_en[4] = {0, 0, 0, 0};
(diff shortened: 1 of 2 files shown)


dholladay00:
My recollection on constant was that there are runtime selectable paths in the init functions that modify the arrays thus for the sake of consistent types it's easiest to leave them as non-const.

Yurlungur:
> My recollection on constant was that there are runtime selectable paths in the init functions that modify the arrays thus for the sake of consistent types it's easiest to leave them as non-const.

Got it thanks. Can you approve if this change seems reasonable?

PR #470, opened 2025-02-21
Jmm/indexable types

Yurlungur:
## PR Summary

While integrating 3T into riot, I began to encounter scenarios where the `lambda` machinery of `singularity-eos` starts to become burdensome. In particular, the z-split modifier on top of (for example) a `SpinerEOS` means that both the base/underlying EOS and the modifier require `lambda` arguments. This is fine in principle; modifier lambda arguments get "appended" on to the back of the lambda and the code traverses through it like a stack.  However, given that type is erased, it can be difficult for a host code to keep track of what lambda argument goes where, or even if a lambda argument is needed.

This MR resolves this issue by the introduction of `IndexableTypes` which may be used in indexers for lambdas. Put simply, an `IndexableType` is an empty struct that can be used to name an index, like an enum, but more free-form. If you overload the `operator[]` function for a lambda indexer to accept an `IndexableType`, most EOS classes that expect `IndexableTypes` will be able to interact with it by name, rather than position in the lambda. For example, if you define:

```C++
class MyLambda_t {
 public:
  MyLambda_t() = default;
  PORTABLE_FORCEINLINE_FUNCTION
  Real &operator[](const std::size_t idx) {
    return data_[idx];
  }
  PORTABLE_FORCEINLINE_FUNCTION
  Real &operator[](const MeanIonizationState &zbar) {
    return data_[2];
  }
 private:
  std::array<Real, 3> data_;
};
```
then you would be able to interact with it as
```C++
MyLambda_t lambda;
lambda[MeanIonizationState()] = zbar;
Real P = eos.PressureFromDensityTemperature(rho, T, lambda);
```
integer and named indexes are interchange-able and backwards compatible, and I provide the ability to create new named types for custom EOS's as well as a convenience struct that automatically builds an indexer form a variadic list of named types. This machinery is all in the `singularity-eos/base/indexable_types.hpp` header.

I also introduce the function `NeedsLambda`, which is now exposed in the variant. This function takes an indexable type and returns true if that indexable type is needed by the class for a given lambda and false otherwise. 

## PR Checklist

### DIFF
--- a/singularity-eos/base/indexable_types.hpp
+++ b/singularity-eos/base/indexable_types.hpp
@@ -22,28 +22,13 @@
 #include <ports-of-call/portability.hpp>
 #include <singularity-eos/base/variadic_utils.hpp>
 
-#define SINGULARITY_DECLARE_INDEXABLE_TYPE(TYPE_NAME)                                    \
-  struct TYPE_NAME : public singularity::IndexableTypes::IndexableTypesBase<TYPE_NAME> {}
-
 namespace singularity {
-namespace IndexableTypes {
-// Base class for indexable types that uses CRTP
-template <typename CRTP>
-struct IndexableTypesBase {
-  template <typename, typename = void>
-  struct IsOverloadedIn : std::false_type {};
-  template <typename T>
-  struct IsOverloadedIn<T, std::void_t<decltype(std::declval<T>()[std::declval<CRTP>()])>>
-      : std::true_type {};
-};
-} // namespace IndexableTypes
-
 namespace IndexerUtils {
 // Convenience function for accessing an indexer by either type or
 // natural number index depending on what is available
 template <typename T, typename Indexer_t>
 PORTABLE_FORCEINLINE_FUNCTION auto &Get(Indexer_t &&lambda, std::size_t idx = 0) {
-  if constexpr (T::template IsOverloadedIn<Indexer_t>::value) {
+  if constexpr (variadic_utils::is_indexable_v<T, Indexer_t>) {
     return lambda[T()];
   } else {
     return lambda[idx];
@@ -52,39 +37,40 @@ PORTABLE_FORCEINLINE_FUNCTION auto &Get(Indexer_t &&lambda, std::size_t idx = 0)
 
 // This is a convenience struct to easily build a small indexer with
 // a set of indexable types.
-template <typename... Ts>
-class VariadicIndexer {
+template <typename Data_t, typename... Ts>
+class VariadicIndexerBase {
  public:
-  VariadicIndexer() = default;
+  VariadicIndexerBase() = default;
+  PORTABLE_FORCEINLINE_FUNCTION
+  VariadicIndexerBase(const Data_t &data) : data_(data) {}
   template <typename T,
             typename = std::enable_if_t<variadic_utils::contains<T, Ts...>::value>>
   PORTABLE_FORCEINLINE_FUNCTION Real &operator[](const T &t) {
-    // get the index of T in the variadic type list
-    std::size_t idx = 0;
-    std::size_t i = 0;
-    ((std::is_same_v<T, Ts> ? idx = i : ++i), ...);
-
+    constexpr std::size_t idx = variadic_utils::GetIndexInTL<T, Ts...>();
     return data_[idx];
   }
   PORTABLE_FORCEINLINE_FUNCTION
   Real &operator[](const std::size_t idx) { return data_[idx]; }
-  PORTABLE_FORCEINLINE_FUNCTION
-  Real &operator[](const int idx) { return data_[idx]; }
   static inline constexpr std::size_t size() { return sizeof...(Ts); }
 
  private:
-  std::array<Real, sizeof...(Ts)> data_;
+  Data_t data_;
 };
-
+// uses an array
+template <typename... Ts>
+using VariadicIndexer = VariadicIndexerBase<std::array<Real, sizeof...(Ts)>, Ts...>;
+// uses a Real*
+template <typename... Ts>
+using VariadicPointerIndexer = VariadicIndexerBase<Real *, Ts...>;
 } // namespace IndexerUtils
 
 namespace IndexableTypes {
-SINGULARITY_DECLARE_INDEXABLE_TYPE(MeanIonizationState);
-SINGULARITY_DECLARE_INDEXABLE_TYPE(LogDensity);
-SINGULARITY_DECLARE_INDEXABLE_TYPE(LogTemperature);
-SINGULARITY_DECLARE_INDEXABLE_TYPE(MeanAtomicMass);
-SINGULARITY_DECLARE_INDEXABLE_TYPE(MeanAtomicNumber);
-SINGULARITY_DECLARE_INDEXABLE_TYPE(ElectronFraction);
+struct MeanIonizationState {};
+struct LogDensity {};
+struct LogTemperature {};
+struct MeanAtomicMass {};
+struct MeanAtomicNumber {};
+struct ElectronFraction {};
 } // namespace IndexableTypes
 } // namespace singularity
 #endif // SINGULARITY_EOS_BASE_INDEXABLE_TYPES_
--- a/singularity-eos/base/indexable_types.hpp
+++ b/singularity-eos/base/indexable_types.hpp
@@ -1,0 +1,94 @@
+//------------------------------------------------------------------------------
+// © 2025. Triad National Security, LLC. All rights reserved.  This
+// program was produced under U.S. Government contract 89233218CNA000001
+// for Los Alamos National Laboratory (LANL), which is operated by Triad
+// National Security, LLC for the U.S.  Department of Energy/National
+// Nuclear Security Administration. All rights in the program are
+// reserved by Triad National Security, LLC, and the U.S. Department of
+// Energy/National Nuclear Security Administration. The Government is
+// granted for itself and others acting on its behalf a nonexclusive,
+// paid-up, irrevocable worldwide license in this material to reproduce,
+// prepare derivative works, distribute copies to the public, perform
+// publicly and display publicly, and to permit others to do so.
+//------------------------------------------------------------------------------
+
+#ifndef SINGULARITY_EOS_BASE_INDEXABLE_TYPES_
+#define SINGULARITY_EOS_BASE_INDEXABLE_TYPES_
+
+#include <array>
+#include <type_traits>
+#include <utility>
+
+#include <ports-of-call/portability.hpp>
+
+#define SINGULARITY_DECLARE_INDEXABLE_TYPE(TYPE_NAME)                                    \
+  struct TYPE_NAME : public singularity::IndexableTypes::IndexableTypesBase<TYPE_NAME> {}
+
+namespace singularity {
+namespace IndexableTypes {
+// Base class for indexable types that uses CRTP
+template <typename CRTP>
+struct IndexableTypesBase {
+  template <typename, typename = void>
+  struct IsOverloadedIn : std::false_type {};
+  template <typename T>
+  struct IsOverloadedIn<T, std::void_t<decltype(std::declval<T>()[std::declval<CRTP>()])>>
+      : std::true_type {};
+};
+} // namespace IndexableTypes
+
+namespace IndexerUtils {
+// Convenience function for accessing an indexer by either type or
+// natural number index depending on what is available
+template <typename T, typename Indexer_t>
+inline auto &Get(Indexer_t &&lambda, std::size_t idx = 0) {
+  if constexpr (T::template IsOverloadedIn<Indexer_t>::value) {
+    return lambda[T()];
+  } else {
+    return lambda[idx];
+  }
+}
+
+// This is a convenience struct to easily build a small indexer with
+// a set of indexable types. Note however that accessing with
+// indexable types is an O(N) operation with this implementation.
+// You probably don't want to use this in real code.
+template <typename... Ts>
+class VariadicIndexer {
+ public:
+  VariadicIndexer() = default;
+  template <typename T>
+  PORTABLE_FORCEINLINE_FUNCTION Real &operator[](const T &t) {
+    constexpr bool contains_v = (std::is_same_v<T, Ts> || ...);
+    static_assert(contains_v, "Indexable type must be in indexer type list!");
+
+    // get the index of T in the variadic type list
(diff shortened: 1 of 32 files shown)


Yurlungur:
I also revise the PTE solver to now actually access the lambda indexer, unless it is passed a `NullIndexer` in which case it will use the chache.

Yurlungur:
One thing we may wish to consider down the line: *should* we be backwards compatible with integer indexing in lambdas? Or should we eventually remove that optionality? Currently, that's used in the xrage code `get_sg_*`. But it could potentially be removed by adding an appropriately overloaded cache object. Probably a discussion for another time. The integer-based indexing plays nicer with the fortran host codes.

jhp-lanl:
> One thing we may wish to consider down the line: _should_ we be backwards compatible with integer indexing in lambdas? Or should we eventually remove that optionality? Currently, that's used in the xrage code `get_sg_*`. But it could potentially be removed by adding an appropriately overloaded cache object. Probably a discussion for another time. The integer-based indexing plays nicer with the fortran host codes.

(I'm reviewing right now) My opinion is that we should by default use a named approach within our code. I'd have to think more about what would be best for the vector interface though.

Yurlungur:
> (I'm reviewing right now) My opinion is that we should by default use a named approach within our code. I'd have to think more about what would be best for the vector interface though.

Yeah agreed. This isn't something we need to change right now. But if we break backwards compatibility some of the template/sfinae complexity I implement here can go away. I'm reluctant to completely remove it right now though, as the backwards compatibility gives us time to transition and test how it feels and where it might break.

Yurlungur:
Tests now passing on re-git.

jhp-lanl (commented):
I think it's worth considering breaking out the square bracket invocable check from the base class. I think that would simplify some of this code significantly.

singularity-eos/base/indexable_types.hpp
```diff
@@ -0,0 +1,90 @@
...
+#include <ports-of-call/portability.hpp>
+#include <singularity-eos/base/variadic_utils.hpp>
+
+#define SINGULARITY_DECLARE_INDEXABLE_TYPE(TYPE_NAME)                                    \
+  struct TYPE_NAME : public singularity::IndexableTypes::IndexableTypesBase<TYPE_NAME> {}
+
+namespace singularity {
+namespace IndexableTypes {
+// Base class for indexable types that uses CRTP
+template <typename CRTP>
```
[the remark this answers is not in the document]
Yurlungur: I think that all that's necessary to make it CRTP is the recursion of templating the base class on the derived class. But that's a nitpick.

singularity-eos/base/indexable_types.hpp
```diff
@@ -0,0 +1,90 @@
...
+    return data_[idx];
+  }
+  PORTABLE_FORCEINLINE_FUNCTION
+  Real &operator[](const std::size_t idx) { return data_[idx]; }
+  PORTABLE_FORCEINLINE_FUNCTION
+  Real &operator[](const int idx) { return data_[idx]; }
+  static inline constexpr std::size_t size() { return sizeof...(Ts); }
+
+ private:
+  std::array<Real, sizeof...(Ts)> data_;
```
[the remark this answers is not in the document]
Yurlungur: It does. But also that's a good idea. Yes I can add a version that takes a block of pointers.

singularity-eos/base/indexable_types.hpp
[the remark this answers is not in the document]
jhp-lanl: Fair enough. See my other comments about whether this inheritance style is really necessary.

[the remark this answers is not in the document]
Yurlungur: Also the CRTP is gone if I split out is_indexable as you suggest.

singularity-eos/base/indexable_types.hpp
```diff
@@ -0,0 +1,90 @@
...
+      : std::true_type {};
+};
+} // namespace IndexableTypes
+
+namespace IndexerUtils {
+// Convenience function for accessing an indexer by either type or
+// natural number index depending on what is available
+template <typename T, typename Indexer_t>
+PORTABLE_FORCEINLINE_FUNCTION auto &Get(Indexer_t &&lambda, std::size_t idx = 0) {
+  if constexpr (T::template IsOverloadedIn<Indexer_t>::value) {
```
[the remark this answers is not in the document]
Yurlungur: Yeah this is cleaner and more flexible. I'll change it to split it out like this.

singularity-eos/base/indexable_types.hpp
```diff
@@ -0,0 +1,90 @@
...
+    std::size_t idx = 0;
+    std::size_t i = 0;
+    ((std::is_same_v<T, Ts> ? idx = i : ++i), ...);
+
+    return data_[idx];
+  }
+  PORTABLE_FORCEINLINE_FUNCTION
+  Real &operator[](const std::size_t idx) { return data_[idx]; }
+  PORTABLE_FORCEINLINE_FUNCTION
+  Real &operator[](const int idx) { return data_[idx]; }
```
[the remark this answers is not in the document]
Yurlungur: It's not necessary---I was just being paranoid. I will remove it.

singularity-eos/closure/mixed_cell_models.hpp
```diff
@@ -461,6 +464,17 @@ class PTESolverBase {
...
+  auto GetLambda(std::size_t m) const {
+    if constexpr (needs_cache_) {
+      return Cache[m];
+    } else {
+      return lambda[m];
+    }
+  }
+
+  static constexpr const bool needs_cache_ =
+      std::is_same<LambdaIndexer, NullIndexer>::value;
```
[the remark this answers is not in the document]
Yurlungur: The idea here is that if the lambda indexer is passed in as a null indexer, we still need a cache---we can't pass in nullptr. That's because the PTE solver may be caching log density and log temperature for root finds for SpinerEOS. That's what this mechanism is for. 

That said, maybe the code base has moved beyond needing this, as I think everywhere where we call the PTE solver, we manage the cache ourselves. If we are happy with always managing the cache ourselves, then this can be removed. We can just always call the lambda indexer.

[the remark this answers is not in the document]
Yurlungur: I am going to remove this special logic and assume that the lambda is always manually managed externally. `get_sg_*` and friends already do this, so long as the size of the scratch array is properly computed. I think it's actually long past time we remove this weird internal logic where the PTE solver manages its own cache. (Which we can blame on me... I added it in the first place as a hack.)

test/test_closure_pte.cpp
```diff
@@ -109,7 +109,7 @@ bool run_PTE_from_state(const int num_pte, EOS *v_EOS, const Real spvol_bulk,
   portableCopyToDevice(v_pressures, pressures.data(), bytes);
 
   // Allocate scratch space for the PTE solver
-  const int pte_solver_scratch_size = RequiredScratch(num_pte);
+  const int pte_solver_scratch_size = RequiredScratch(num_pte, true);
```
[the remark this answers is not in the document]
Yurlungur: This optional argument is now removed.

test/test_pte_ideal.cpp
```diff
@@ -83,8 +94,8 @@ PORTABLE_INLINE_FUNCTION Real set_state(Real rho_nom, Real sie_nom, RealIndexer
 template <typename EOS_Indexer_t>
 int ComparePTEs(EOS_Indexer_t eoss, const std::size_t NEOS, const std::size_t NTRIAL) {
   constexpr Real EPS = 1e-5;
-  const std::size_t nscratch_rt = PTESolverRhoTRequiredScratch(NEOS);
-  const std::size_t nscratch_pt = PTESolverPTRequiredScratch(NEOS);
+  const std::size_t nscratch_rt = PTESolverRhoTRequiredScratch(NEOS, false);
+  const std::size_t nscratch_pt = PTESolverPTRequiredScratch(NEOS, false);
```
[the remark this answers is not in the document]
Yurlungur: Optional argument is now removed.

Yurlungur:
Thanks for the review @jhp-lanl ! I think all your comments should be resolved now.

singularity-eos/eos/modifiers/zsplit_eos.hpp
```diff
@@ -268,7 +274,16 @@ class ZSplit : public EosBase<ZSplit<ztype, T>> {
...
-    return std::max(0.0, lambda[t_.nlambda()]);
+    using namespace variadic_utils;
+    if (is_nullptr(lambda)) {
+      PORTABLE_THROW_OR_ABORT(
+          "StellarCollapse: lambda must contain Ye and 1 space for caching.\n");
+    }
+    if constexpr (is_indexable_v<Indexer_t, IndexableTypes::MeanIonizationState>) {
+      return std::max(0.0, lambda[IndexableTypes::MeanIonizationState()]);
+    } else {
+      return std::max(0.0, lambda[t_.nlambda()]);
```
dholladay00, line 285: Does this work if I nest modifiers each of which requires lambdas?

Yurlungur: It does. In fact, that's the idea. With nested modifiers, the lambdas work as a stack---you always treat the end of the lambda array as yours and then work your way inward to the base object.

jhp-lanl (approved):

singularity-eos/eos/modifiers/zsplit_eos.hpp
jhp-lanl: If we had two modifiers (or an EOS and a modifer) that require the same lambda information (I can't think of a good reason right now), the typed indexing would allow them to both use the same information though. With integer indexing, the information would need to be duplicated, yes?

Yurlungur: :+1: exactly

singularity-eos/eos/eos_electrons.hpp
```diff
@@ -209,8 +214,8 @@ class IdealElectrons : public EosBase<IdealElectrons> {
  private:
   template <typename Indexer_t = Real *>
   PORTABLE_INLINE_FUNCTION Real _Cv(Indexer_t &&lambda) const {
-    const Real Z = std::max(lambda[Lambda::Zi], static_cast<Real>(0.0));
-    return _Cvbase * Z;
+    const Real Z = IndexerUtils::Get<IndexableTypes::MeanIonizationState>(lambda);
+    return _Cvbase * std::max(Z, static_cast<Real>(0.0));
```
lroberts36, line 218: Should this be 
```c++
const Real Z = IndexerUtils::Get<IndexableTypes::MeanIonizationState>(lambda, Lambda::Zi);
```
for backwards compatibility?

singularity-eos/eos/eos_spiner.hpp
```diff
@@ -1889,7 +1907,9 @@ SpinerEOSDependsRhoSie::FillEos(Real &rho, Real &temp, Real &energy, Real &press
     }
   } else {
     lRho = toLog_(rho, lRhoOffset_);
-    if (!variadic_utils::is_nullptr(lambda)) lambda[0] = lRho;
+    if (!variadic_utils::is_nullptr(lambda)) {
+      IndexerUtils::Get<IndexableTypes::LogDensity>(lambda) = lRho;
```
lroberts36, line 1911: Include the `Lambda::lRho` argument explicitly for clarity, and similarly below.

singularity-eos/eos/modifiers/zsplit_eos.hpp
```diff
@@ -224,6 +226,10 @@ class ZSplit : public EosBase<ZSplit<ztype, T>> {
 
   PORTABLE_INLINE_FUNCTION
   int nlambda() const noexcept { return 1 + t_.nlambda(); }
+  template <typename Indexable>
+  static inline constexpr bool NeedsLambda() {
+    return std::is_same<Indexable, IndexableTypes::MeanIonizationState>::value;
```
lroberts36, line 231: Should this also have `|| T::template NeedsLambda<Indexable>();`?

singularity-eos/eos/modifiers/zsplit_eos.hpp
```diff
@@ -268,7 +274,16 @@ class ZSplit : public EosBase<ZSplit<ztype, T>> {
  private:
   template <typename Indexer_t = Real *>
   PORTABLE_FORCEINLINE_FUNCTION Real GetIonizationState_(Indexer_t &&lambda) const {
-    return std::max(0.0, lambda[t_.nlambda()]);
+    using namespace variadic_utils;
+    if (is_nullptr(lambda)) {
+      PORTABLE_THROW_OR_ABORT(
+          "StellarCollapse: lambda must contain Ye and 1 space for caching.\n");
```
lroberts36, line 280: This message looks like a copy and paste error.

test/CMakeLists.txt
```diff
@@ -32,12 +32,13 @@ add_executable(
   eos_infrastructure_tests
   catch2_define.cpp
   eos_unit_test_helpers.hpp
-  test_eos_modifiers.cpp
+  #test_eos_modifiers.cpp
```
lroberts36, line 35: Should this be uncommented or should the line be removed?

test/test_indexable_types.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+using namespace singularity::IndexerUtils;
+using namespace singularity::IndexableTypes;
+
+using Lambda_t = VariadicIndexer<MeanIonizationState, ElectronFraction, LogDensity>;
+
+class ManualLambda_t {
+ public:
+  ManualLambda_t() = default;
+  Real &operator[](const std::size_t idx) { return data_[idx]; }
```
lroberts36, line 33: Will things work if a lambda is defined without `Real &operator[](const std::size_t idx);`?

lroberts36 (approved):
LGTM, just a few small questions/nitpicks. This turned out to be a clean way of dealing with general lambdas that should allow for more flexibility downstream. You know me though, I am always a fan of using types like this.

singularity-eos/eos/modifiers/zsplit_eos.hpp
Yurlungur: Yes. Good catch!

test/test_indexable_types.cpp
Yurlungur: Yes---so long as it contains all types needed.

PR #481, opened 2025-03-12
Remove mutable from spiner-eos and fillin docs a bit

Yurlungur:
## PR Summary

@dholladay00 note that the `mutable` members in the `Spiner` equations of state might still cause problems. I don't know if this is happening, but it's about time we remove the undefined behavior.

The recent PR #470 that introduced indexable types suggests another way to retain diagnostic capability without relying on unsafe paradigms. I optionally add indexable types for the two diagnostic variables which were mutable. They can now optionally be written to the lambda if the lambda is overloaded for them. Unlike other variables, which have backwards compatibility with integer-based indexing, these diagnostics are *only* available if the lambda supports the indexable type for them. This reflects that this isn't really intended to be a public facing API--it's a diagnostic tool used for developers to debug.

I considered completely deleting the logic and just using print statements if we ever need this optionality again. But this seemed like a relatively simple, thread-safe way to stash information. So I went with that.

## PR Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
 ### Changed (changing behavior/API/variables/...)
 
 ### Infrastructure (changes irrelevant to downstream codes)
+- [[PR481]](https://github.com/lanl/singularity-eos/pull/481) Move mutable diagnostic variables in singularity-eos into optional indexable types in a lambda.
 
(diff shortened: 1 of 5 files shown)


Yurlungur:
This may or may not be related to the issues @jhp-lanl saw in #375

jhp-lanl (approved):
Looks good to me

PR #483, opened 2025-03-18
Added optional argument to fortran interface

gopsub:
Added capability to pass optional arrays via fortran interface. This is required for implementation of EoS-es that use additional quantities (such as burn fraction) to evaluate Pressure


gopsub:
@Yurlungur , @dholladay00 , As discussed yesterday, I've taken a crack at the optional parameters for the fortran interface. I've modified just the one  function for now. Please let me know if any changes are needed. If this looks OK, I will proceed with the remainder. Thanks.

singularity-eos/eos/singularity_eos.cpp
```diff
@@ -19,6 +19,34 @@
...
+class lambdaIndexer {
+ public:
+  lambdaIndexer(int n, double *data) : n_(n), data_(data) {}
+
+  double *operator[](int i) const { return &(data_[n_ * i]); }
+
+ private:
+  int n_;
+  double *data_;
+};
```
Yurlungur, line 48: I think we should probably name this something else to disambiguate from others like this one. Maybe `LambdaIndexer2D`?

singularity-eos/eos/singularity_eos.cpp
```diff
@@ -342,8 +370,14 @@ int init_sg_eospac(const int matindex, EOS *eos, const int id,
 
 int get_sg_PressureFromDensityInternalEnergy(int matindex, EOS *eos, const double *rhos,
                                              const double *sies, double *pressures,
-                                             const int len) {
-  eos[matindex].PressureFromDensityInternalEnergy(rhos, sies, pressures, len);
+                                             const int len, const int stride = -1,
+                                             double *lambda_data = NULL) {
+  if (stride != -1 && lambda_data != NULL) {
```
Yurlungur, line 375: In C++ `nullptr` is preferred over `NULL`.

Yurlungur (approved):
This looks good to me. Thanks @gopsub . @dholladay00 should also look though. Also should we add a test perhaps? It doesn't look like we have much in the way of test coverage for the fortran interface though...

singularity-eos/eos/singularity_eos.cpp
gopsub: Renamed.

gopsub:
> @jhp-lanl this is in support of the Crest EOS on the Fortran side.
> 
> @gopsub it looks like the PTE test is failing on clang, but I don't think that's your fault. I will look at it.

OK -- thanks for looking

Yurlungur:
Weird... clang test now passing. Maybe that segmentation fault was a red herring. I also can't reproduce on my machine.

gopsub:
> This looks good to me. Thanks @gopsub . @dholladay00 should also look though. Also should we add a test perhaps? It doesn't look like we have much in the way of test coverage for the fortran interface though...

The new feature is used only by the extras side. Do we have anything (either current or planned) that uses this new feature on this side?

gopsub:
> Weird... clang test now passing. Maybe that segmentation fault was a red herring. I also can't reproduce on my machine.

Stranger things have happened.

singularity-eos/eos/singularity_eos.cpp
```diff
@@ -19,6 +19,34 @@
...
+
+which will return the memory address of the element n*i of the array lambda
+===============================================
+*/
+
+class lambdaIndexer2D {
+ public:
+  lambdaIndexer2D(int n, double *data) : n_(n), data_(data) {}
+
+  double *operator[](int i) const { return &(data_[n_ * i]); }
```
Yurlungur, line 43: This needs device markings, which is why tests are failing on re-git
```suggestion
  PORTABLE_FORCEINLINE_FUNCTION
  double *operator[](int i) const { return &(data_[n_ * i]); }
```

gopsub: > Added

The re-git CI is still failing. Not sure what the error means

singularity-eos/eos/eos_helmholtz.hpp
```diff
@@ -700,7 +700,7 @@ class Helmholtz : public EosBase<Helmholtz> {
                                        Indexer_t &&lambda, Real &rho, Real &sie) const {
     using RootFinding1D::regula_falsi;
     using RootFinding1D::Status;
-    PORTABLE_REQUIRE(temp > = 0, "Non-negative temperature required");
+    PORTABLE_REQUIRE(temp >= 0, "Non-negative temperature required");
```
Yurlungur, line 703: syntactic error. Surprised I didn't catch this before.

test/test_pte.cpp
```diff
@@ -229,7 +229,11 @@ auto TestPTE(const std::string name, const std::size_t nscratch_vars,
   }
   std::cout << std::endl;
 
+#ifdef PORTABILITY_STRATEGY_KOKKOS
+  return rho_v;
+#else
   return rho_d;
+#endif // PORTABILITY_STRATEGY_KOKKOS
```
Yurlungur, line 236: When Kokkos is on, this was actually not valid, because `rho_d` points to memory managed by Kokkos. This resolves that issue.

Yurlungur:
Tests now passing on re-git.

jhp-lanl (approved):

gopsub:
Hi all: Is this ready for merge?

jhp-lanl:
> Hi all: Is this ready for merge?

@Yurlungur I'll let you merge in case there are any lingering issues

singularity-eos/eos/singularity_eos.f90
```diff
@@ -720,14 +724,21 @@ end function init_sg_eospac_f
...
+    eos, rhos, sies, pressures, len, stride, lambda_data) &
     result(err)
     integer(c_int), intent(in) :: matindex, len
     real(kind=8), dimension(:,:,:), intent(in), target:: rhos, sies
     real(kind=8), dimension(:,:,:), intent(inout), target:: pressures
     type(sg_eos_ary_t), intent(in)    :: eos
-    err = get_sg_PressureFromDensityInternalEnergy(matindex-1, &
-           eos%ptr, c_loc(rhos(1,1,1)), c_loc(sies(1,1,1)), c_loc(pressures(1,1,1)), len)
+    integer(c_int), intent(in), optional :: stride
+    real(kind=8), dimension(:,:,:,:), intent(inout), target, optional::lambda_data
```
dholladay00, line 734: Why is this assumed to be a rank 4 array? Won’t that dictate the stride?

gopsub: As far as I understood this, the stride is the same as the number of elements in the first rank of the array, and the other three ranks contain the actual data. So, for stride=2, and using hypothetical pagosa variables X and Y as the extra data I'm sending into singularity, I would have to copy pagosa data as

`lambda_data(1, :, :, :) = X(:, :, :)`

`lambda_data(2, :, :, :) = Y(:, :, :)`

Is this not right?

dholladay00: How difficult would it be to make this input a rank 2 by flattening the rank3 arrays into a rank 1 (then you have the size of the first rank the stride and the 2nd rank is the num_cells length thing)?

Yurlungur: On the C++ side it's just a pointer so I think there's flexibility on the fortran side. Up to you for however you want to do it @gopsub .

Yurlungur:
I'd like to see the discussion between @dholladay00 and @gopsub resolved. But otherwise I'm ready to click the button.

singularity-eos/eos/singularity_eos.f90
gopsub: > How difficult would it be to make this input a rank 2 by flattening the rank3 arrays into a rank 1 (then you have the size of the first rank the stride and the 2nd rank is the num_cells length thing)?

It wouldn't be difficult, but stylistically, the other rank 3 arrays (e.g. rhos, sies) are not collapsed into a rank 1. Is there a strong reason to break this pattern? Moreover, as @Yurlungur said, the C++ side gets the pointer to the first element.

Yurlungur:
@dholladay00 and @gopsub feel free to continue your discussion. I'm going to click merge on this now. We can revise based on your discussion if we need to.

singularity-eos/eos/singularity_eos.f90
dholladay00: I see now. The rank 3 arrays is a bit SAP specific but given that that is consistent with other arrays in this routine, that is ok.

PR #486, opened 2025-04-04
Move scratch_size and max_scratch_size into base class

jhp-lanl:
## PR Summary

Moves the `scratch_size()` and `max_scratch_size()` from the EOS to the base class.

For almost all EOS, these are zero, and we don't need to keep duplicating this as we add EOS. It makes more sense to just add special versions for specific use cases.

I also made the inputs const for possible performance benefits

## PR Checklist


jhp-lanl:
Note that I think the EOSPAC tests are the only tests that rely on these member functions, so the LANL CI tests should pass before this is merged.

Yurlungur (approved):
Nice cleanup. :+1:

jhp-lanl:
LANL CI passes. Merging

PR #488, opened 2025-04-08
Bugfix --- stride needs to have "value" attribute

gopsub:
## PR Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@
 
 ### Infrastructure (changes irrelevant to downstream codes)
 - [[PR481]](https://github.com/lanl/singularity-eos/pull/481) Move mutable diagnostic variables in singularity-eos into optional indexable types in a lambda.
+- [[PR486]](https://github.com/lanl/singularity-eos/pull/486) Use default behavior for max_scratch_size and scratch_size in the eos base class
 
 ### Removed (removing behavior/API/varaibles/...)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - [[PR444]](https://github.com/lanl/singularity-eos/pull/444) Add Z split modifier and electron ideal gas EOS
 
 ### Fixed (Repair bugs, etc)
+- [[PR485]](https://github.com/lanl/singularity-eos/pull/485) Fix segfault in Fortran interface related to EOSPAC initialization
 - [[PR478]](https://github.com/lanl/singularity-eos/pull/478) Fix bug in KPT test. Add more extensive clang build to github CI matrix.
 - [[PR473]](https://github.com/lanl/singularity-eos/pull/473) Resolve memory issue. Thanks for the catch, Richard!
 - [[PR468]](https://github.com/lanl/singularity-eos/pull/468) Move definition of def_en and def_v to an implementation file
--- a/singularity-eos/eos/eos_base.hpp
+++ b/singularity-eos/eos/eos_base.hpp
@@ -86,7 +86,9 @@ char *StrCat(char *destination, const char *source) {
   using EosBase<__VA_ARGS__>::EntropyFromDensityTemperature;                             \
   using EosBase<__VA_ARGS__>::EntropyFromDensityInternalEnergy;                          \
   using EosBase<__VA_ARGS__>::GibbsFreeEnergyFromDensityTemperature;                     \
-  using EosBase<__VA_ARGS__>::GibbsFreeEnergyFromDensityInternalEnergy;
+  using EosBase<__VA_ARGS__>::GibbsFreeEnergyFromDensityInternalEnergy;                  \
+  using EosBase<__VA_ARGS__>::scratch_size;                                              \
+  using EosBase<__VA_ARGS__>::max_scratch_size;
 
 // This macro adds these methods to a derived class. Due to scope,
 // these can't be implemented in the base class, unless we make
@@ -803,6 +805,14 @@ class EosBase {
 
   PORTABLE_INLINE_FUNCTION
   Real RhoPmin(const Real temp) const { return 0.0; }
+
+  static inline unsigned long scratch_size(const std::string method,
+                                           const unsigned int nelements) {
+    return 0;
+  }
+  static inline unsigned long max_scratch_size(const unsigned int nelements) {
+    return 0;
+  }
 
   // JMM: EOS's which encapsulate a mix or reactions may wish to vary
   // this.  For example, Helmholtz and StellarCollapse. This isn't the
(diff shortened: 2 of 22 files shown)


jhp-lanl:
CI is failing with the following:
```
 Error: Variable ‘stride’ at (1) cannot have both the OPTIONAL and the VALUE attribute because procedure ‘get_sg_bulkmodulusfromdensityinternalenergy’ is BIND(C)
```

gopsub:
> CI is failing with the following:
> 
> ```
>  Error: Variable ‘stride’ at (1) cannot have both the OPTIONAL and the VALUE attribute because procedure ‘get_sg_bulkmodulusfromdensityinternalenergy’ is BIND(C)
> ```

It builds on my end, and correctly passes on the value of stride. Without the "VALUE" attribute, the reference is passed. What's the solution?

jhp-lanl:
> > CI is failing with the following:
> > ```
> >  Error: Variable ‘stride’ at (1) cannot have both the OPTIONAL and the VALUE attribute because procedure ‘get_sg_bulkmodulusfromdensityinternalenergy’ is BIND(C)
> > ```
> 
> It builds on my end, and correctly passes on the value of stride. Without the "VALUE" attribute, the reference is passed (not desirable). What's the solution?

Are you compiling with Intel? It seems to play a bit fast and loose with the Fortran standard.

I think the fundamental issue is that in C, an optional variable is given a default value. In that sense, it's never _not_ declared, whereas in Fortran the `OPTIONAL` keyword says that the variable may not _ever_ have a value if it's not provided as an input argument. From the C perspective, there's no logical way to resolve this.

It's possible there's an easier fix, but my suggestion would be to move the default value from the C side to the Fortran wrapper side. So in `get_sg_PressureFromDensityInternalEnergy_f()` use the check on whether `stride` is present or not to set a default value of -1 if it's not present. Then the C function just always takes a stride.

jhp-lanl:
Another possibility is to define two C functions with different names, one to accept a stride (and `lambda_data`) and one that doesn't. Then create two C-interoperable routines in Fortran and then pick which you call. On the C side, this would split the `if(stride...` statement so that it's just two routines with different behavior. The branching then occurs only on the Fortran side in which function is called.

gopsub:
How's this? Sorry it's a bit cluttered because I worked off the gss/entropy branch to get all 3 functions that need lambdas

Yurlungur (approved):
Good catch! I'm happy with this version if @jhp-lanl is.

jhp-lanl (approved):
I think this will work

jhp-lanl:
Let me at least start the LANL tests to make sure this all works

jhp-lanl:
It looks like the LANL CI is having an issue with the different definitions:
```
ftn-855 ftn: ERROR SINGULARITY_EOS, File = ../../../../usr/WS2/jeffreyp/xcap-gitlab-ci/jeffreyp/builds/dDssG5uCy/002/xcap/oss/singularity-eos/singularity-eos/eos/singularity_eos.f90, Line = 27, Column = 8 
  The compiler has detected errors in module "SINGULARITY_EOS".  No module information file will be created for this module.
ftn-1295 ftn: ERROR GET_SG_ENTROPYFROMDENSITYINTERNALENERGY_LAMBDA, File = ../../../../usr/WS2/jeffreyp/xcap-gitlab-ci/jeffreyp/builds/dDssG5uCy/002/xcap/oss/singularity-eos/singularity-eos/eos/singularity_eos.f90, Line = 272, Column = 8 
  "get_sg_EntropyFromDensityInternalEnergy" is defined or referenced at line 260 (/usr/workspace/jeffreyp/xcap-gitlab-ci/jeffreyp/builds/dDssG5uCy/002/xcap/oss/singularity-eos/singularity-eos/eos/singularity_eos.f90) and here.  The number of arguments do not match.  Expected 6, but found 8.
ftn-1295 ftn: ERROR GET_SG_PRESSUREFROMDENSITYINTERNALENERGY_LAMBDA, File = ../../../../usr/WS2/jeffreyp/xcap-gitlab-ci/jeffreyp/builds/dDssG5uCy/002/xcap/oss/singularity-eos/singularity-eos/eos/singularity_eos.f90, Line = 298, Column = 8 
  "get_sg_PressureFromDensityInternalEnergy" is defined or referenced at line 286 (/usr/workspace/jeffreyp/xcap-gitlab-ci/jeffreyp/builds/dDssG5uCy/002/xcap/oss/singularity-eos/singularity-eos/eos/singularity_eos.f90) and here.  The number of arguments do not match.  Expected 6, but found 8.
ftn-1295 ftn: ERROR GET_SG_BULKMODULUSFROMDENSITYINTERNALENERGY_LAMBDA, File = ../../../../usr/WS2/jeffreyp/xcap-gitlab-ci/jeffreyp/builds/dDssG5uCy/002/xcap/oss/singularity-eos/singularity-eos/eos/singularity_eos.f90, Line = 335, Column = 8 
  "get_sg_BulkModulusFromDensityInternalEnergy" is defined or referenced at line 323 (/usr/workspace/jeffreyp/xcap-gitlab-ci/jeffreyp/builds/dDssG5uCy/002/xcap/oss/singularity-eos/singularity-eos/eos/singularity_eos.f90) and here.  The number of arguments do not match.  Expected 6, but found 8.
  ```

jhp-lanl:
This might be a bug in the Fortran compiler on the MI300 and MI250 machines.

I wonder if it's getting confused about the C vs Fortran definitions of the function. You might change `get_sg_PressureFromDensityInternalEnergy` to `get_sg_PressureFromDensityInternalEnergyNoLambda` on the Fortran side and we can see if that works.

jhp-lanl (changes requested):
Once this passes on the MI250/MI300 machines I think this is good to go

jhp-lanl:
Testing on the LANL CI now

jhp-lanl:
Same issue 😖 @gopsub . I think the C interoperable function needs to be passed all 8 arguments all the time.

I would try one of two solutions:
1. Rather than using an `if` statement in the C interoperable function to select the behavior, you can just have two versions of the function that do the different behaviors. Then on the Fortran side, you can bind an interoperable function to each version and then use the logic in the wrapper function to determine which one you call depending on whether the optional arguments are provided or not. You'd need to add an additional C function, but the advantage of this approach would be that the logic would be cleanly handled on the Fortran side. The disadvantage would be that the interface is more complicated.
2. Keep one version of the C function and make the optional arguments mandatory. Then you can use the Fortran wrapper logic to select the values passed to the function and handle those values the exact same way you already do. You can delete one of the Fortran interoperable functions with this approach. The advantage of this approach is that you have minimal interface functions, but the logic is duplicated on the C and Fortran sides.

Yurlungur:
I think option #2 is probably slightly preferable IMO, as it reduces code duplication, and this function essentially only exists for the fortran API anyway.

gopsub:
Took a crack at option 2

jhp-lanl:
Running the CI on LANL

jhp-lanl:
Tests pass!

jhp-lanl (approved):

PR #489, opened 2025-04-09
Move nlambda default behavior into base class

jhp-lanl:
## PR Summary

Does what it says...

## PR Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@
 ### Infrastructure (changes irrelevant to downstream codes)
 - [[PR481]](https://github.com/lanl/singularity-eos/pull/481) Move mutable diagnostic variables in singularity-eos into optional indexable types in a lambda.
 - [[PR486]](https://github.com/lanl/singularity-eos/pull/486) Use default behavior for max_scratch_size and scratch_size in the eos base class
+- [[PR489]](https://github.com/lanl/singularity-eos/pull/489) Move default behavior for nlambda to the eos base class
 
 ### Removed (removing behavior/API/varaibles/...)
 
--- a/python/module.hpp
+++ b/python/module.hpp
@@ -346,7 +346,7 @@ py::class_<T> eos_class(py::module_ & m, std::string name) {
 
     .def("MinimumDensity", &T::MinimumDensity)
     .def("MinimumTemperature", &T::MinimumTemperature)
-    .def_property_readonly_static("nlambda", &T::nlambda)
+    .def_property_readonly_static("nlambda", [](py::object) { return T::nlambda(); })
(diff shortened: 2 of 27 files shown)


jhp-lanl:
Note: I didn't include a `using` statement in the `SG_ADD_BASE_CLASS_USINGS` macro because the intention is to hide the `nlambda` function from the base class when there are more than zero lambdas. With a `using` statement, the function becomes ambiguously defined rather than clearly shadowed.

We have more of these cases in the code where it might be useful to hide the base class version of a function, but we wouldn't be able to do so because of excessive `using` statements. That said, I think it's probably fine to just leave them until we want to hide or shadow the base class version with a different implementation in the derived class.

jhp-lanl:
I'll double check. Since I'm changing things, @Yurlungur what are your thoughts on just making these all `constexpr static inline` instead of using the `PORTABLE_INLINE_FUNCTION` decorator?

Yurlungur:
If it works on device, I'm fine with that. If I recall, we were bitten by constexpr not working on device as we expected previously.

jhp-lanl:
From my reading, if the expression is compile time constant, it should be inlined and will be available on device. I can't think of a reason why this function _wouldn't_ be compile time constant.

jhp-lanl:
> Well lets just give it a try and see if tests pass. (May need to add a test to stress this)

I ran into an issue with the `Variant` and the modifers. The main issue was that the EOS object isn't `constexpr`. In the modifier case, I could have called the static version from the type rather than the EOS object itself, but I felt like I was battling too much here.

jhp-lanl:
Okay @Yurlungur I think everything should be the same now. I _think_ I could have made `static` or `constexpr static` work, but it would have meant that the Variant versions would _not_ be `constexpr` while the EOS objects would (and the modifers would have to use the type rather than the EOS object). I think I'll just stick with what works for now. Hopefully the compiler can still inline most of this

Yurlungur:
> Okay @Yurlungur I think everything should be the same now. I _think_ I could have made `static` or `constexpr static` work, but it would have meant that the Variant versions would _not_ be `constexpr` while the EOS objects would (and the modifers would have to use the type rather than the EOS object). I think I'll just stick with what works for now. Hopefully the compiler can still inline most of this

Sounds good. Too bad that ended up being a bit messy.

jhp-lanl:
I had some issues with the python API, but that appears to be working with some minor changes. Will merge when all tests pass (maybe with the exception of the Ampere nodes since the queue times for those are pretty long today).

jhp-lanl:
@Yurlungur tests pass. If you approve, feel free to merge

Yurlungur (approved):
LGTM

Issue #490, opened 2025-04-09
Add docs for the fortran API


Yurlungur:

Yurlungur:
Partially resolved by #504, but still work to do here long term.

PR #504, opened 2025-05-07
Add Basic Fortran Interface Documentation

Yannicked:
Adds rudimentary Fortran interface documentation.

## PR Summary

This PR adds documentation for all available Fortran functions, and tries to explain the arguments of each function. I've also added two examples at the bottom that use some of the functions.
The function signatures might be a bit verbose, let me know what you think. I can make it more concise.

Addresses #490

## PR Checklist

### DIFF
--- a/doc/sphinx/src/fortran.rst
+++ b/doc/sphinx/src/fortran.rst
@@ -32,8 +32,6 @@ The ``sg_eos_ary_t`` type is a wrapper around a C pointer to the underlying EOS 
 Initialization Functions
 -----------------------
 
-The following functions are available for initializing EOS models:
-
 .. code-block:: fortran
 
    integer function init_sg_eos_f(nmat, eos) result(err)
@@ -42,6 +40,11 @@ The following functions are available for initializing EOS models:
    end function init_sg_eos_f
 
 Initializes the EOS array with the specified number of materials.
+
+The following functions are available for initializing EOS models:
+
+Ideal Gas
+^^^^^^^^^
 
 .. code-block:: fortran
 
@@ -63,6 +66,9 @@ Initializes an Ideal Gas EOS model at the specified material index.
 - ``Cv``: Specific heat at constant volume
 - ``sg_mods_enabled``: Optional array for enabling modifiers
 - ``sg_mods_values``: Optional array for modifier values
+
+Gruneisen
+^^^^^^^^^
 
 .. code-block:: fortran
 
@@ -93,6 +99,9 @@ Initializes a Gruneisen EOS model at the specified material index.
 - ``sg_mods_enabled``: Optional array for enabling modifiers
 - ``sg_mods_values``: Optional array for modifier values
 
+Jones-Wilkins-Lee
+^^^^^^^^^^^^^^^^^
+
 .. code-block:: fortran
 
    integer function init_sg_JWL_f(matindex, eos, A, B, R1, R2, w, rho0, Cv, &
@@ -117,6 +126,9 @@ Initializes a JWL (Jones-Wilkins-Lee) EOS model at the specified material index.
 - ``sg_mods_enabled``: Optional array for enabling modifiers
 - ``sg_mods_values``: Optional array for modifier values
 
+Davis Products
+^^^^^^^^^^^^^^
+
 .. code-block:: fortran
 
    integer function init_sg_DavisProducts_f(matindex, eos, a, b, k, n, vc, pc, &
@@ -140,6 +152,9 @@ Initializes a Davis Products EOS model at the specified material index.
 - ``Cv``: Specific heat at constant volume
 - ``sg_mods_enabled``: Optional array for enabling modifiers
 - ``sg_mods_values``: Optional array for modifier values
+
+Davis Reactants
+^^^^^^^^^^^^^^^
 
 .. code-block:: fortran
 
@@ -171,6 +186,9 @@ Initializes a Davis Reactants EOS model at the specified material index.
 - ``sg_mods_enabled``: Optional array for enabling modifiers
 - ``sg_mods_values``: Optional array for modifier values
 
+Noble-Abel
+^^^^^^^^^^
+
 .. code-block:: fortran
 
    integer function init_sg_NobleAbel_f(matindex, eos, gm1, Cv, &
@@ -195,6 +213,9 @@ Initializes a Noble-Abel EOS model at the specified material index.
 - ``sg_mods_enabled``: Optional array for enabling modifiers
 - ``sg_mods_values``: Optional array for modifier values
 
+Stiff Gas
+^^^^^^^^^
+
 .. code-block:: fortran
 
    integer function init_sg_StiffGas_f(matindex, eos, gm1, Cv, &
@@ -218,6 +239,9 @@ Initializes a Stiff Gas EOS model at the specified material index.
 - ``qq``: Energy shift parameter
 - ``sg_mods_enabled``: Optional array for enabling modifiers
 - ``sg_mods_values``: Optional array for modifier values
+
+SAP Polynomial
+^^^^^^^^^^^^^^
 
 .. code-block:: fortran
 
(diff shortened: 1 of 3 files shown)


doc/sphinx/src/fortran.rst
```diff
@@ -0,0 +1,693 @@
...
+The following functions are available for initializing EOS models:
+
+.. code-block:: fortran
+
+   integer function init_sg_eos_f(nmat, eos) result(err)
+     integer(kind=c_int), intent(in) :: nmat
+     type(sg_eos_ary_t), intent(in)  :: eos
+   end function init_sg_eos_f
+
+Initializes the EOS array with the specified number of materials.
```
Yurlungur, line 44: There should probably be some discussion that only the EOS's available in the variant are defined, which depends on how the code is built. For example `init_sg_SpinerDependsRhoT` is only available if the code is built with spiner and HDF5, EOSPAC is only built if EOSPAC is enabled, and similarly Helmholtz is only built if Helmholtz is enabled.

doc/sphinx/src/fortran.rst
```diff
@@ -0,0 +1,693 @@
...
+     result(err)
+     integer(c_int), intent(in) :: matindex, len
+     real(kind=8), dimension(:,:,:), intent(in), target:: rhos, sies
+     real(kind=8), dimension(:,:,:), intent(inout), target:: entropies
+     type(sg_eos_ary_t), intent(in)    :: eos
+     integer(c_int), intent(in), optional :: stride
+     real(kind=8), dimension(:,:,:,:), intent(inout), target, optional::lambda_data
+   end function get_sg_EntropyFromDensityInternalEnergy_f
+
+Calculates entropy from density and specific internal energy.
```
Yurlungur, line 359: It may also be worth noting that some equations of state are not thermodynamically complete, and may lack an entropy.

doc/sphinx/src/fortran.rst
```diff
@@ -0,0 +1,693 @@
...
+     real(kind=8), dimension(:,:), target, optional, intent(inout) :: frac_cv
+     real(kind=8),                         optional, intent(in)    :: mass_frac_cutoff
+   end function get_sg_eos_f
+
+Main EOS evaluation function for mixed cells. This function is used to calculate various thermodynamic quantities for multiple materials in mixed cells.
+
+- ``nmat``: Number of materials
+- ``ncell``: Number of cells
+- ``cell_dim``: Cell dimension
+- ``option``: Option flag for calculation
```
Yurlungur, line 481: In the fullness of time, this function may need significantly more discussion. But I wouldn't add that now.

I also wanted to let you know know this particular function may change, as it was originally written with one downstream code in mind. So we're exploring what this should look like in general. (And feedback is welcome.)

doc/sphinx/src/fortran.rst
```diff
@@ -0,0 +1,693 @@
...
+       if (res /= 0) then
+           print *, "Error finalizing EOS:", res
+           stop
+       end if
+   
+   end program tabulated_eos_example
+
+.. note::
+   
+   When using the ``get_sg_eos_f`` function, temperature inputs should be provided in electron volts (eV), not Kelvin. This is different from some other functions in the interface that expect temperature in Kelvin.
```
Yurlungur, line 692: This is an example of something we might want to change down the line. :)

Yurlungur (changes requested):
Thanks for adding this @Yannicked ! Can you once again also add a changelog entry? Otherwise, I am happy to merge, modulo some minor nitpicks above. Frankly, I think any documentation is better than none, and I'm grateful you went ahead and added this. So I don't feel the need to iterate much on this MR. We can add more docs later in separate MRs as they get written.

Yurlungur:
@jhp-lanl @dholladay00 please skim for correctness if you have a moment.

Yurlungur (approved):

doc/sphinx/src/fortran.rst
jhp-lanl: This is actually code that maybe shouldn't live in `singularity-eos` to be honest since it's very specific.

I'd advocate that we should expose a very general Fortran interface, but that the `get_sg_eos` interface is waaaaay too specific. There's an advantage to having it in our code since we test its capabilities, but I think we should try to capture those some other way and/or explore other ways of testing the library against this interface.

Yurlungur: Yeah absolutely. That said, if there is a desire to access PTE on the fortran side, we should think how we'd like the generic interface to look.

doc/sphinx/src/fortran.rst
```diff
@@ -0,0 +1,742 @@
...
+     result(err)
+     integer(c_int), value, intent(in) :: matindex
+     type(sg_eos_ary_t), intent(in)    :: eos
+     real(kind=8), value, intent(in)   :: rho0, a0, a1, a2c, a2e, a3,&
+              b0, b1, b2c, b2e, b3
+     integer(kind=c_int), dimension(:), target, optional, intent(inout) :: sg_mods_enabled
+     real(kind=8), dimension(:), target, optional, intent(inout)        :: sg_mods_values
+   end function init_sg_SAP_Polynomial_f
+
+Initializes a SAP (Separate Analytic Polynomials) Polynomial EOS model at the specified material index.
```
jhp-lanl, line 260: I'm not sure this is what SAP stands for in this context 😉 . We should maybe consider renaming in the future, but suffice it to say that it's a specific implementation of _a_ polynomial EOS.

doc/sphinx/src/fortran.rst
```diff
@@ -0,0 +1,742 @@
...
+- ``T0``: Reference temperature
+- ``A, B, C``: Davis Reactants parameters
+- ``G0``: Gruneisen parameter
+- ``Z``: Davis Reactants parameter
+- ``alpha``: Davis Reactants parameter
+- ``Cv0``: Specific heat at constant volume
+- ``sg_mods_enabled``: Optional array for enabling modifiers
+- ``sg_mods_values``: Optional array for modifier values
+
+Noble-Abel
```
jhp-lanl, line 189: Only available with `SINGULARITY_USE_V_AND_V_EOS`

doc/sphinx/src/fortran.rst
```diff
@@ -0,0 +1,742 @@
...
+- ``matindex``: Material index (1-based in Fortran)
+- ``eos``: EOS array
+- ``gm1``: Gamma minus 1 (γ-1)
+- ``Cv``: Specific heat at constant volume
+- ``bb``: Covolume parameter
+- ``qq``: Energy shift parameter
+- ``sg_mods_enabled``: Optional array for enabling modifiers
+- ``sg_mods_values``: Optional array for modifier values
+
+Stiff Gas
```
jhp-lanl, line 216: Only available with `SINGULARITY_USE_V_AND_V_EOS`

doc/sphinx/src/fortran.rst
```diff
@@ -0,0 +1,742 @@
...
+- ``matindex``: Material index (1-based in Fortran)
+- ``eos``: EOS array
+- ``gm1``: Gamma minus 1 (γ-1)
+- ``Cv``: Specific heat at constant volume
+- ``Pinf``: Pressure shift parameter
+- ``qq``: Energy shift parameter
+- ``sg_mods_enabled``: Optional array for enabling modifiers
+- ``sg_mods_values``: Optional array for modifier values
+
+SAP Polynomial
```
jhp-lanl, line 243: Only available with `SINGULARITY_USE_V_AND_V_EOS`

doc/sphinx/src/fortran.rst
jhp-lanl: This is noted below
```
Some EOS models do not support Entropy. For more details see :ref:`models <Implemented EOS models>`
```

jhp-lanl (commented):
Great work!

doc/sphinx/src/fortran.rst
jhp-lanl: True enough! I think we can discuss future possibilities now that we are going to have three different Fortran codes interfacing

Issue #390, opened 2024-07-16
Figure out PTE convergence issues


jhp-lanl:
In https://github.com/lanl/singularity-eos/pull/383 I added two tests where the PTE solver hadn't been converging. I fixed the issue with the first test, but I have yet to fix the second test.

This is the second test that doesn't converge (I believe with underflow in the temperature):
```c++
  GIVEN("Two equations of state") {
    // Set up the three EOS
    constexpr int num_eos = 2;
    std::array<EOS, num_eos> eos_arr;
    // Reference state
    constexpr Real P0 = 1.0e6; // 1 bar
    constexpr Real T0 = 296;   // K
    // Ideal gas air
    constexpr Real rho0_air = 1e-03; // g/cc
    constexpr Real Gruneisen_air = 0.4;
    constexpr Real CV_air = P0 / rho0_air / (Gruneisen_air * T0);
    eos_arr[0] = IdealGas(Gruneisen_air, CV_air);
    // Spiner copper EOS
    constexpr int Cu_matid = 3337;
    const std::string eos_file = "../materials.sp5";
    eos_arr[1] = SpinerEOSDependsRhoT(eos_file, Cu_matid);

    GIVEN("A state that would cause a negative temperature in the PTE solver") {
      // Define the state
      constexpr Real spvol_bulk = 4.010467628234189e-01;
      constexpr Real sie_bulk = 3.290180957185173e+07;
      const std::array<const Real, num_eos> mass_frac = {0.000312273191678158,
                                                         0.999687726808322};

      THEN("The PTE solver should converge") {
        EOS *v_EOS = copy_eos_arr_to_device(num_eos, eos_arr);
        const bool pte_converged =
            run_PTE_from_state(num_eos, v_EOS, spvol_bulk, sie_bulk, mass_frac);
        // TODO: make this test converge
        // CHECK(pte_converged);
        finalize_eos_arr(num_eos, v_EOS);
      }
    }
  }
```

Yurlungur:
Some progress in #506 but still unresolved. I think someone needs to chase this down with print statements to see what's failing. The PTE 2mat example added in #506 however should help with that.

PR #506, opened 2025-05-13
[WIP] pte example and robustness fix

Yurlungur:
## PR Summary

Recently I was asked for a C++ example of how to use the PTE solvers, which is sorely lacking. I have also been recently chasing down some robustness issues in the PTE for an HEDP problem run in riot. I decided to kill two birds with one stone. This MR adds two new examples to the examples folder:
- `pte_2mat.cpp` sets up and runs the PTE solver on two materials specified by the user on a single cell. It is commented to explain what's going on. I wrote it for use with Spiner but it is easy to modify to use with eospac.
- `map_pt_space.cpp` takes two materials and walks through valid sets of volume fractions. At each one, it computes the temperature so that the total energy is correct, and outputs the residual as well as the P/T of each material at each point. This produces a "map" of the PT landscape for the material pair.

Exploration of this space revealed two "failure modes" of the PTE solvers, which I also resolve in this MR:
1. In the line search, because we also look at physical ranges of validity, it's possible for the solver to take steps of size zero for many iterations. I add a check that now catches this case and fails out early.
2. When there is a saddle point in the residual, the PTE solver can fail to converge as it will travel down the wrong path through the space. Empirically, it seems to help to have a better (and usually larger) temperature guess. To help with that, I added a routine that takes one (or more) newton-step towards finding 
$u_{tot} - \sum_m \alpha_m \rho_m \varepsilon_m (\rho_m, T_{guess}) = 0$
to modify the initial temperature guess, where $\rho_m$ is the set of initial density guesses. The update is only accepted if the temperature increases due to the Newton step. This one extra 1D Newton step seems to work shockingly well. Convergence appears faster and PTE inputs that previously failed to converge now do converge successfully. I suspect part of the secret sauce here is that pressure structure is reduced at high temperature, so you can escape structure that would otherwise trap you in a local extremum.

This should still probably be tested a bit to check robustness, but I think it will help in situations that the solvers previously struggled in.

## PR Checklist

### DIFF
--- a/example/map_pt_space.cpp
+++ b/example/map_pt_space.cpp
@@ -1,0 +1,275 @@
+//------------------------------------------------------------------------------
+// © 2025. Triad National Security, LLC. All rights reserved.  This
+// program was produced under U.S. Government contract 89233218CNA000001
+// for Los Alamos National Laboratory (LANL), which is operated by Triad
+// National Security, LLC for the U.S.  Department of Energy/National
+// Nuclear Security Administration. All rights in the program are
+// reserved by Triad National Security, LLC, and the U.S. Department of
+// Energy/National Nuclear Security Administration. The Government is
+// granted for itself and others acting on its behalf a nonexclusive,
+// paid-up, irrevocable worldwide license in this material to reproduce,
+// prepare derivative works, distribute copies to the public, perform
+// publicly and display publicly, and to permit others to do so.
+//------------------------------------------------------------------------------
+
+// C headers
+#include <cmath>
+#include <cstdio>
+
+// C++ headers
+#include <iostream>
+#include <memory>
+#include <string>
+
+// This library contains portable utilities
+#include <ports-of-call/portability.hpp>
+#include <ports-of-call/portable_errors.hpp>
+
+// This contains useful tools for preventing things like divide by zero
+#include <singularity-eos/base/robust_utils.hpp>
+// 1D root finding
+#include <singularity-eos/base/root-finding-1d/root_finding.hpp>
+// Needed to import the eos models
+#include <singularity-eos/eos/eos.hpp>
+
+// This library contains the spiner table object, which we will use to
+// store our output
+#include <spiner/databox.hpp>
+#include <spiner/interpolation.hpp>
+
+// These are the specializations of spiner we will use
+using DataBox = Spiner::DataBox<Real>;
+using RegularGrid1D = Spiner::RegularGrid1D<Real>;
+using Spiner::DBDeleter;
+
+// Number of equations of state to use, always 2
+constexpr std::size_t NEOS = 2;
+
+// Set the EOS you want to use here.
+using EOS = singularity::SpinerEOSDependsRhoT;
+
+int main(int argc, char *argv[]) {
+  if (argc < 6) {
+    std::cerr << "Usage: " << argv[0]
+              << " residual_savename rhobar1 rhobar2 sietot nvfrac eosargs..."
+              << std::endl;
+    std::exit(1);
+  }
+
+  // This is needed for Kokkos
+#ifdef PORTABILITY_STRATEGY_KOKKOS
+  Kokkos::initialize();
+#endif
+  {
+    // File we'll save the residual map to, in ascii. P/T trajectories will be printed to
+    // to stdout
+    const std::string savename = argv[1];
+    // alpha rho for each material where here alpha is volume
+    // fraction and rho is microphysical material density.
+    const Real rhobar1 = atof(argv[2]);
+    const Real rhobar2 = atof(argv[3]);
+    // total specific internal energy in the problem.
+    const Real sietot = atof(argv[4]);
+    // number of vfrac points
+    const std::size_t nvfrac = atoi(argv[5]);
+
+    // Below are the arguments for an individual equation of
+    // state. Here I'm assuming SpinerEOSDependsRhoT but you could
+    // modify for your needs.
+    const std::string materialsfile = argv[6];
+    const int matids[] = {atoi(argv[7]), atoi(argv[8])};
+
+    // Now let's load up the EOS's.
+    EOS eos1 = EOS(materialsfile, matids[0]);
+    EOS eos2 = EOS(materialsfile, matids[1]);
+
+    // The total bulk density
+    const Real rhotot = rhobar1 + rhobar2;
+    const Real utot = rhotot * sietot;
+
+    // We will be walking through volume fractions so let's grid that
+    // up using a Spiner RegularGrid1D. This is the grid of alpha1s. alpha2 = 1 - alpha1.
+    constexpr Real lvfrac_min = -5;
+    constexpr Real lvfrac_max = -0.5;
+    RegularGrid1D lvfracs(lvfrac_min, lvfrac_max, nvfrac);
+
+    // We will be doing root finds in T space so we need to know what
+    // our bounds are. There is often a minimum temperature for an EOS
+    // but not a maximum.
+    const Real Tmin = std::max(eos1.MinimumTemperature(), eos2.MinimumTemperature());
+    const Real Tmax = 1e10; // change this if needed
+
+    // We also want to store quantities of interest as we walk the volume fractions
+    // This is a way of managing databoxes so that memory is automatically managed
+    // Temperature
+    std::unique_ptr<DataBox, DBDeleter> pTs(
+        new DataBox(Spiner::AllocationTarget::Device, nvfrac));
+    DataBox Ts = *pTs; // A shallow coppy of pTs. Unmanaged memory.
+    // Pressure for materials 1 and 2
+    std::unique_ptr<DataBox, DBDeleter> pPs(
+        new DataBox(Spiner::AllocationTarget::Device, nvfrac, NEOS));
+    DataBox Ps = *pPs;
+    // vfrac, energy, and pressure residuals for the trajectory
+    std::unique_ptr<DataBox, DBDeleter> pTrajRes(
+        new DataBox(Spiner::AllocationTarget::Device, nvfrac, 3));
+    DataBox trajRes = *pTrajRes;
+
+    // We also need to ensure whatever internal tabulated data for
+    // each EOS is available on device.
+    eos1 = eos1.GetOnDevice();
+    eos2 = eos2.GetOnDevice();
+
+    // Ok let's walk the grid. We will do so on device with a portableFor loop
+    portableFor(
+        "Compute P and T for each alpha", 0, nvfrac, PORTABLE_LAMBDA(const int i) {
+          // volume fractions
+          const Real lvfrac = lvfracs.x(i);
+          const Real vfrac = std::pow(10., lvfrac);
+          const Real alpha1 = vfrac;
+          const Real alpha2 = 1. - vfrac;
+
+          // microphysical densities
+          Real rho1 = singularity::robust::ratio(rhobar1, alpha1);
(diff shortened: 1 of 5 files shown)


singularity-eos/closure/mixed_cell_models.hpp
```diff
@@ -275,10 +275,29 @@ class PTESolverBase {
...
+      Real usum = 0;
+      for (std::size_t m = 0; m < nmat; ++m) {
+        Real sie = eos[m].InternalEnergyFromDensityTemperature(rho[m], T, lambda[m]);
+        Real cv = eos[m].SpecificHeatFromDensityTemperature(rho[m], T, lambda[m]);
+        Real usum = rhobar[m] * sie;
+        dudt += rhobar[m] * cv;
+      }
+      return T - (usum - utotal_scale * uscale) / dudt;
+    };
+
```
mauneyc-LANL, line 295: Wouldn't just checking `dudt < 0` be equivalent? In effect, you're stepping the guess towards the local optimum.

I suspect the effect here is doing a partial flattening of the energy surface - similar to something like [basin hopping](https://en.wikipedia.org/wiki/Basin-hopping).

Yurlungur: What do you mean by checking `dudt < 0`?

mauneyc-LANL: Is this worth putting this behind a flag? It's conceivable that downstream codes may have techniques for generating guesses that could conflict.

Yurlungur: yeah, I'd be supportive of hiding it behind a flag.

singularity-eos/closure/mixed_cell_models.hpp
```diff
@@ -1785,6 +1805,18 @@ PORTABLE_INLINE_FUNCTION SolverStatus PTESolver(System &s) {
     Real scale = s.ScaleDx();
     PORTABLE_REQUIRE(scale <= 1.0, "PTE Solver is attempting to increase the step size");
 
+    // If scale is very small, we may be iterating around the regime
+    // of validity and not making progress. Try to catch this, fail
+    // out early, and report this as a failure, so fallbacks can take
+    // over in a host code. We check that this has happened for at
+    // least two iterations, in case we can escape in one.
+    if ((std::abs(scale) < 1e-16) && (std::abs(scale_last) < 1e-16)) {
```
mauneyc-LANL, line 1817: Use a named constant, preferably something like `pte::ls::minimum_scale_magnitude`.

mauneyc-LANL: Suggestion for this

```c++
ntry = 0;
/* ... */
if (std::abs (scale) < pte::min_scale ) ++ntry;
if (ntry > params.small_step_trys )
{ /* ...fails....*/ }
```

mauneyc-LANL (approved):
See comments for suggestions.

singularity-eos/closure/mixed_cell_models.hpp
mauneyc-LANL: > What do you mean by checking `dudt < 0`?

You accept the new guess if `T` increases, so in effect you're checking if `dudt` is negative (unless i'm mistaken and the numerator can be negative?)

Yurlungur: The numerator can be either sign. Both because the current total energy can be either side of the target and because the total (or target) energy can be negative

singularity-eos/closure/mixed_cell_models.hpp
Yurlungur: Done.

singularity-eos/closure/mixed_cell_models.hpp
Yurlungur: > Is this worth putting this behind a flag? It's conceivable that downstream codes may have techniques for generating guesses that could conflict.

Done.

jhp-lanl:
A couple comments:
1. If it's not already covered by the examples, can you add any hard states to the https://github.com/lanl/singularity-eos/blob/main/test/test_closure_pte.cpp
2. Can you test the commented-out test there and see if your improvements cause it to converge? See https://github.com/lanl/singularity-eos/issues/390

example/map_pt_space.cpp
```diff
@@ -0,0 +1,276 @@
...
+// These are the specializations of spiner we will use
+using DataBox = Spiner::DataBox<Real>;
+using RegularGrid1D = Spiner::RegularGrid1D<Real>;
+using Spiner::DBDeleter;
+
+// Number of equations of state to use, always 2
+constexpr std::size_t NEOS = 2;
+
+// Set the EOS you want to use here.
+using EOS = singularity::SpinerEOSDependsRhoT;
```
jhp-lanl, line 49: Do you want to define a custom variant?

example/map_pt_space.cpp
```diff
@@ -0,0 +1,276 @@
...
+// National Security, LLC for the U.S.  Department of Energy/National
+// Nuclear Security Administration. All rights in the program are
+// reserved by Triad National Security, LLC, and the U.S. Department of
+// Energy/National Nuclear Security Administration. The Government is
+// granted for itself and others acting on its behalf a nonexclusive,
+// paid-up, irrevocable worldwide license in this material to reproduce,
+// prepare derivative works, distribute copies to the public, perform
+// publicly and display publicly, and to permit others to do so.
+//------------------------------------------------------------------------------
+
```
jhp-lanl, line 14: Would it make sense to briefly describe the purpose of this example?

example/map_pt_space.cpp
```diff
@@ -0,0 +1,276 @@
...
+    DataBox Ts_h = *pTs_h;
+    portableCopyToHost(Ts_h.data(), Ts.data(), Ts.sizeBytes());
+
+    // Get max temperature achieved. Note this could have been done in
+    // a single step via a portableReduce, but we split it out for clarity.
+    Real Tmax_walked = Ts_h.max();
+
+#ifdef PORTABILITY_STRATEGY_KOKKOS
+    Kokkos::fence();
+#endif
```
jhp-lanl, line 187: Is this fence necessary? You're already interacting with `Ts_h` above after the copy to host (which I think is blocking).

example/map_pt_space.cpp
```diff
@@ -0,0 +1,276 @@
...
+    }
+
+    eos1.Finalize();
+    eos2.Finalize();
+  }
+#ifdef PORTABILITY_STRATEGY_KOKKOS
+  Kokkos::finalize();
+#endif
+  return 0;
+}
```
jhp-lanl, line 276: Thanks for codifying this in an example. I wish we had done that with the EOSPAC and Spiner comparison code! We should keep it here whenever that is redone.

example/pte_2mat.cpp
```diff
@@ -0,0 +1,188 @@
...
+// National Security, LLC for the U.S.  Department of Energy/National
+// Nuclear Security Administration. All rights in the program are
+// reserved by Triad National Security, LLC, and the U.S. Department of
+// Energy/National Nuclear Security Administration. The Government is
+// granted for itself and others acting on its behalf a nonexclusive,
+// paid-up, irrevocable worldwide license in this material to reproduce,
+// prepare derivative works, distribute copies to the public, perform
+// publicly and display publicly, and to permit others to do so.
+//------------------------------------------------------------------------------
+
```
jhp-lanl, line 14: Again, I think it might be useful to have a short description of the example at the top.

example/pte_2mat.cpp
```diff
@@ -0,0 +1,188 @@
...
+// Needed to import the eos models
+#include <singularity-eos/eos/eos.hpp>
+
+// Number of equations of state to use. For this example, it's 2.
+constexpr std::size_t NEOS = 2;
+// Number of PTE solves to do
+constexpr std::size_t NTRIAL = 1;
+
+// Set the EOSs you want to use here.
+using EOS = singularity::Variant<singularity::SpinerEOSDependsRhoT>;
```
jhp-lanl, line 42: Again, does it make sense to make a custom variant?

Yurlungur:
>     2. Can you test the commented-out test there and see if your improvements cause it to converge? See [Figure out PTE convergence issues #390](https://github.com/lanl/singularity-eos/issues/390)

Good idea. Will try it.

jhp-lanl (approved):

example/map_pt_space.cpp
Yurlungur: Done.

example/pte_2mat.cpp
Yurlungur: Done.

example/pte_2mat.cpp
Yurlungur: What do you mean? This is a custom variant? Do you mean I should be just using the standard one? Making it custom is just for faster compile times.

example/map_pt_space.cpp
Yurlungur: Switched to a variant

example/map_pt_space.cpp
Yurlungur: You're right. Removed.

example/map_pt_space.cpp
```diff
@@ -0,0 +1,288 @@
...
+          const Real alpha1 = vfrac;
+          const Real alpha2 = 1. - vfrac;
+
+          // microphysical densities
+          Real rho1 = singularity::robust::ratio(rhobar1, alpha1);
+          Real rho2 = singularity::robust::ratio(rhobar2, alpha2);
+
+          const Real u1 = rhobar1 * eos1.InternalEnergyFromDensityTemperature(rho1, T);
+          const Real u2 = rhobar2 * eos2.InternalEnergyFromDensityTemperature(rho2, T);
+          residuals(j, i, 0) = u1 + u2 - utot, utot;
```
dholladay00, line 229: Is there some weird application of the `,` operator here?

Yurlungur: I'm shocked that compiled lol. That should just be `residuals(j, i, 0) = u1 + u2 - utot;`. Thanks for the catch!

example/pte_2mat.cpp
jhp-lanl: For some reason I thought it was the standard variant... my bad!

Yurlungur:
OK Tests now pass on MI300. This MR now depends on #510 which must be merged first.

Yurlungur:
> A couple comments:
> 
>     1. If it's not already covered by the examples, can you add any hard states to the https://github.com/lanl/singularity-eos/blob/main/test/test_closure_pte.cpp
> 
>     2. Can you test the commented-out test there and see if your improvements cause it to converge? See [Figure out PTE convergence issues #390](https://github.com/lanl/singularity-eos/issues/390)

I also still need to do this, which I would like to do before merge.

Yurlungur:
>     1. If it's not already covered by the examples, can you add any hard states to the https://github.com/lanl/singularity-eos/blob/main/test/test_closure_pte.cpp

Done.

>     2. Can you test the commented-out test there and see if your improvements cause it to converge? See [Figure out PTE convergence issues #390](https://github.com/lanl/singularity-eos/issues/390)

Unfortunately this still doesn't converge. Will have to attack this in a later MR...

jhp-lanl:
> Unfortunately this still doesn't converge. Will have to attack this in a later MR...

Worth a try! Thanks for checking that. Hopefully soon we'll figure out how to get that state to converge...

Issue #508, opened 2025-05-19
EOS copies iompacting performance with small array calls


Yannicked:
I’m seeing a performance issue when calling the EOS routines on many small arrays (≈8 elements each). Profiling shows that every call to
```c++
my_eos.InternalEnergyFromDensityTemperature(density, temperature)
```
invokes the EOS copy constructor. Since the SpinerEOS object is about 20 KB (per sizeof), these repeated copies add significant overhead. It appears that PR #106 introduced an explicit copy in this path.

Yurlungur:
This copy is necessary on GPUs as references and pointers don't span the two memory spaces. We could potentially remove this copy in the serial case... 

I am a little reluctant to add such a branch, as it will make debugging more difficult. But maybe the switch can be based on portability strategy... I.e., for Kokkos, behavior would be unchanged but for PORTABILITY_STRATEGY_NONE we could short circuit this. Would that work?

Yurlungur:
Alternatively, if you are trying to run on GPUs, you are passing the kernel launch overhead. We should talk about your specific use case, as you may want to restructure your code.

Yannicked:
I’m running in serial right now, so no GPU (yet). What if we used [`KOKKOS_CLASS_LAMBDA`](https://kokkos.org/kokkos-core-wiki/API/core/macros-special/host_device_macros.html#kokkos-class-lambda) (i.e. `[=, *this]`, instead of `[=]`) to capture and copy `*this`? Would that skip the extra copy in serial (`PORTABILITY_STRATEGY_NONE`) while still working on device?

jhp-lanl:
I'm a bit confused here... why does the entire class need to be copy captured? It seems like we need to copy it to device and then copy capture the _pointer_ to the class on the device.

It sounds from @Yannicked 's original post that a class copy is being created for every lambda capture.

Yurlungur:
> I'm a bit confused here... why does the entire class need to be copy captured? 

There's a few entangled pieces of this:

1. To access the member functions on device, we at least need to do `[this]`  because otherwise the `this` pointer isn't in scope, and thus class methods will not be available. That's a C++ism
2. But we can't capture `this` on device because it points to the wrong memory. Hence why `KOKKOS_CLASS_LAMBDA` captures `*this`. But that has the effect of capturing the entire class by value, and doing an extra copy.
[...]

PR #511, opened 2025-05-28
sometimes capture by reference

Yurlungur:
## PR Summary

As discussed in #508 @Yannicked reports problems with the vector API for small vectors. This was partially resolved by #509 . In general, it would be better to use the cell based API in this case. However, *specifically* for CPU, this can be resolved by capturing by reference instead of by value in the `portableFor`. This MR adds this optionally. It is off by default and comes with significant warning labels. 

This is marked WIP as I'm not sure we want to pursue this. It is, as far as I can tell, safe here. But does fundamentally change the semantics of these functions in some cases, which I consider a risk as far as debugging and maintainability. 

An alternative approach would be to expose scalar fortran functions. That would be a lot more boiler plate, but potentially more maintainable. Curious to hear folks thoughts. @dholladay00 @jhp-lanl .

Note that getting scalar fortran functions to bind onto device is apparently a challenge, as C/fortran interop is a fundamentally host-side thing. Going this path would at least let us punt on that until that machinery is potentially more mature.

## PR Checklist


jhp-lanl:
> Note that getting scalar fortran functions to bind onto device is apparently a challenge, as C/fortran interop is a fundamentally host-side thing. Going this path would at least let us punt on that until that machinery is potentially more mature.

I'm not sure we want to support fortran device binding unless there's a dedicated person willing to invest in that. But it sounds like a scalar fortran interface for the host _would_ be useful, yes?

Yurlungur:
> > Note that getting scalar fortran functions to bind onto device is apparently a challenge, as C/fortran interop is a fundamentally host-side thing. Going this path would at least let us punt on that until that machinery is potentially more mature.
> 
> I'm not sure we want to support fortran device binding unless there's a dedicated person willing to invest in that. But it sounds like a scalar fortran interface for the host _would_ be useful, yes?

Yes I think that's right. I'm not sure how many LOC it would be but I think it would be a useful addition.

PR #521, opened 2025-07-18
Added Transform template to SIe -- issues with movie tolog and fromlog functions to separate file, and with format;

jbasabe:
## PR Checklist

- [ ] Adds a test for any bugs fixed. Adds tests for new features.
- [ ] Format your changes by using the `make format` command after configuring with `cmake`.
- [ ] Document any new features, update documentation for changes made.
- [ ] Make sure the copyright notice on any files you modified is up to date.
- [ ] After creating a pull request, note it in the CHANGELOG.md file.
- [ ] LANL employees: make sure tests pass both on the github CI and on the Darwin CI

If preparing for a new release, in addition please check the following:
- [ ] Update the version in cmake.
- [ ] Move the changes in the CHANGELOG.md file under a new header for the new release, and reset the categories.
- [ ] Ensure that any `when='@main'` dependencies are updated to the release version in the package.py

PR #538, opened 2025-08-07
[WIP] Add ability to look at rho at Pmin and make it more robust

Yurlungur:
…And clean up spiner sie a bit

This is a bit more of chasing down bugs in riot. The main issue here is that if you work with Maxwell constructed tables. Maxwell constructed tables *without* a tilt are just as bad as a spinodal region, because a flat gradient is just as bad as a gradient going the wrong direction, so we want to avoid the Maxwell constructed region. To do so, I modify the way Spiner computes `RhoPMin` to be more robust to this by:
1. Checking dP/drho as well as just looking for minimum rho
2. Checking distance from P(rho=0)
3. Enforcing that dRhoPMin(T)/dT <= 0

I also add some introspection to the spiner tables so that they can report this, and do some minor cleanup in the spiner code. Note this is going to require a small, but not automatic merge with #521 .

Here's results that it works better. Blue curve is original RhoPMin and orange is fixed. 
rho_p_min_analysis

## PR Checklist

### DIFF
--- a/singularity-eos/base/root-finding-1d/root_finding.hpp
+++ b/singularity-eos/base/root-finding-1d/root_finding.hpp
@@ -265,7 +265,7 @@ PORTABLE_INLINE_FUNCTION Status newton_raphson(const T &f, const Real ytarget,
     std::tie(yg, dfunc) = f(_x); // C++11 tuple unpacking
 
     // check if we are converged already
-    if (std::abs(yg - ytarget) < (ytol * ytarget)) break;
+    if (std::abs(yg - ytarget) < std::abs(ytol * ytarget)) break;
 
     // not converged; compute the next step
     _xold = _x;


Yurlungur:
Note this may be a reason to revive #201

jhp-lanl:
I'm a bit confused... since there is no tilt in the Maxwell-construction, shouldn't the minimum be at zero density always?

I assume the PTE solver just barfs on $\left(\frac{\partial P}{\partial \rho}\right)_T = 0$ as a rule. Would it instead make sense to ascribe a minimum value to this derivative so we escape the vapor dome and instead land on either side?

Maxwell-constructed tables are usually materials for which the vapor phase can't be ignored (there are also a couple tables that include tension isotherms along with Maxwell-constructed isotherms, but these are more rare). As a result, the physical PTE solution could very well be in the vapor domain or the condensed domain. The volume change associated with the phase transition will make finding a PTE solution difficult though, and mixed-phase solutions could be valid.

Yurlungur:
> I'm a bit confused... since there is no tilt in the Maxwell-construction, shouldn't the minimum be at zero density always?

Yeah, physically. So even the old behavior was wrong.

> I assume the PTE solver just barfs on $\left(\frac{\partial P}{\partial \rho}\right)_T = 0$ as a rule. 

Yep exactly, get stuck in the vapor dome and the code just cycles forever. 

That's why my solution here was to change the meaning of RhoPMin to be the highest density (at a given temperature) for which dP/drho > 0.

> Would it instead make sense to ascribe a minimum value to this derivative so we escape the vapor dome and instead land on either side?

Interesting thought. That could be an alternative option, I suppose. I'd prefer to still put this in, though, at least as a stopgap, as I can see it works well and makes the PTE solver significantly more robust. I think RhoPMin should be monotone no matter what, for example.
 
> Maxwell-constructed tables are usually materials for which the vapor phase can't be ignored (there are also a couple tables that include tension isotherms along with Maxwell-constructed isotherms, but these are more rare). As a result, the physical PTE solution could very well be in the vapor domain or the condensed domain. The volume change associated with the phase transition will make finding a PTE solution difficult though, and mixed-phase solutions could be valid.

Yeah, I understand what you mean. I think we should table that, though, as it might be a longer exploration.

Yurlungur:
> > I'm a bit confused... since there is no tilt in the Maxwell-construction, shouldn't the minimum be at zero density always?
> 
> Yeah, physically. So even the old behavior was wrong.

Actually thinking about this more, isn't it ambiguous? If the minimum pressure is at multiple points, isn't any one of them a valid `RhoPMin`? In any case, we certainly want it to be deterministic and the most stable behavior seems to be at the edge of the vapor dome/Maxwell constructed region.

Yurlungur:
@jhp-lanl have you had a chance to review this more concretely?

Yurlungur:
> By thermodynamic construction, the Maxwell construction is at the vapor pressure, which is only the pressure of the vapor in equilibrium with the condensed phase. This is always positive, and if you expand the vapor to zero density, the pressure will decrease to zero as it approaches an ideal gas. So for a true Maxwell construction, the flat region is always at a pressure greater than the minimum pressure for that isotherm.

Physically yes---agreed. What I meant though was if I open up the table, there is a flat line of 0 pressure extending from rho = 0 to some density. Any point along that line is "rho of Pmin." But that's because the EOS has been modified by the Maxwell construction. Of course, this is at a pressure greater than the physical minimum pressure, since the tension state has been eliminated.

jhp-lanl:
> > By thermodynamic construction, the Maxwell construction is at the vapor pressure, which is only the pressure of the vapor in equilibrium with the condensed phase. This is always positive, and if you expand the vapor to zero density, the pressure will decrease to zero as it approaches an ideal gas. So for a true Maxwell construction, the flat region is always at a pressure greater than the minimum pressure for that isotherm.
> 
> Physically yes---agreed. What I meant though was if I open up the table, there is a flat line of 0 pressure extending from rho = 0 to some density. Any point along that line is "rho of Pmin." But that's because the EOS has been modified by the Maxwell construction. Of course, this is at a pressure greater than the physical minimum pressure, since the tension state has been eliminated.

Under most circumstances though, this _isn't_ the case. Looking at the 311 table for 3337 in particular, you can see that this only occurs for the zero-K isotherm since the vapor essentially cannot exist at this temperature (the 301 table starts with a negative slope at this temperature). Every subsequent isotherm has zero pressure at zero density, and larger pressures at higher densities. In this case, the zero gradient algorithm would skip the vapor dome.

5267 is a weird table here. The Maxwell construction actually contains a slight tilt to it, but before this there are a series of _very_ low density points for which the pressure is a constant, non-zero value for many isotherms. In this case, the zero-gradient algorithm would actually do the right thing.

Let me take a look at the code itself though...

Yurlungur:
> Under most circumstances though, this isn't the case. Looking at the 311 table for 3337 in particular, you can see that this only occurs for the zero-K isotherm since the vapor essentially cannot exist at this temperature (the 301 table starts with a negative slope at this temperature). Every subsequent isotherm has zero pressure at zero density, and larger pressures at higher densities. In this case, the zero gradient algorithm would skip the vapor dome.

But 3337 isn't Maxwell constructed right? I'm not sure I understand why zero-gradient would skip the vapor dome.

That said, I've actually removed the gradient check. I think it's more robust to just sweep up looking for minimum pressure (*or* very close to P(rho=0)) and then enforce that rho of Pmin itself is monotone in temperature.

singularity-eos/eos/eos_spiner_rho_sie.hpp
```diff
@@ -416,21 +415,37 @@ herr_t SpinerEOSDependsRhoSie::loadDataboxes_(const std::string &matid_str, hid_
...
-        PMin_ = dependsRhoT_.P(j, i);
+      Real P = dependsRhoT_.P(j, i);
+      PMin_ = std::min(P, PMin_);
+      if (P <= PMin_ || std::abs(P - PMin_) < SOFT_THRESH) {
+        jmax = j;
+      }
+      // Also look for the rightmost point with a small change iwth
+      // respect to the minimum density point
+      if (std::abs(P - P0) < SOFT_THRESH) {
         jmax = j;
```
jhp-lanl, line 437: I think it's worth considering the outcome of this algorithm for various EOS structure, if nothing else so we can add some comments here to understand perhaps surprising results.
1. _The EOS is not Maxwell-constructed_
  This shouldn't change anything, and the algorithm should correctly find the minimum pressure
2. _The EOS has a **physical** Maxwell construction_
  `P0 = 0` and there is some range of densities where the pressure is some positive constant. If that constant pressure is less than `SOFT_THRESH`, then we basically take the right-most point that below this threshold. Therefore, any points in the vapor phase at a pressure below `SOFT_THRESH` are ignored for determining the density at which the minimum pressure occurs. When the vapor pressure is above `SOFT_THRESH`, the density at which the pressure is below `SOFT_THRESH` is returned.
3. _The EOS has a Maxwell construction, but it's tilted_
  If the pressure is below `SOFT_THRESH`, see above. Otherwise, now the answer may depend on the grid resolution. If the spiner grid is coarse enough, the difference in pressure between subsequent points in the Maxwell-constructed region may not trigger the `std::abs(P - PMin_) < SOFT_THRESH` condition to increase `jmax` to the current `j`. If the grid is fine enough, then the difference is small and the condition _will_ increase `jmax`.
  
It seems to me like this approach is unnecessarily excluding valid vapor-phase points. If the goal is to provide a suitable bound for the PTE solver, then I don't see why these points wouldn't be valid. A vapor should be able to expand infinitely, and Maxwell-constructed tables are essentially allowing for this.

jhp-lanl (commented):
I still don't understand why these changes are needed. If they are for the PTE solver, then a Maxwell-constructed table _shouldn't_ be excluding vapor phase points since the material can smoothly expand to those states. The only states I feel we need to avoid are those for which the the material is thermodynamically unstable, i.e. the pressure slope w.r.t. density is negative.

Yurlungur:
Moving into WIP status for now due to the issue @jhp-lanl  mentioned

singularity-eos/eos/eos_gruneisen.hpp
```diff
@@ -112,7 +112,7 @@ class Gruneisen : public EosBase<Gruneisen> {
       Indexer_t &&lambda = static_cast<Real *>(nullptr)) const;
   template <typename Indexer_t = Real *>
   PORTABLE_INLINE_FUNCTION Real SpecificHeatFromDensityTemperature(
-      const Real rho, const Real temperatummmmmmre,
+      const Real rho, const Real temperature,
```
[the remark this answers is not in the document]
Yurlungur: Might have been my cat lol

singularity-eos/eos/eos_spiner_rho_sie.hpp
Yurlungur: Actually I think this causes problems for non Maxwell constructed equations of state, like copper like you pointed out, which is obviously not ideal. Let me see if I can come up with a less aggressive solution.

Issue #539, opened 2025-08-08
Fix the `newton_raphson` root finding function.


Bismarrck:
This IF check will always fail if `ytarget` is negative.

https://github.com/lanl/singularity-eos/blob/a61b6e6e810f00c0d78efbe010668ebb1fadda12/singularity-eos/base/root-finding-1d/root_finding.hpp#L268

Suggested fix:

```cpp
if (std::abs(yg - ytarget) < std::abs(ytol * ytarget)) break;
```

PR #540, opened 2025-08-08
Fix the `newton_raphson` bug.

Bismarrck:
The IF check will always fail if `ytarget` is negative. This modification will fix this.

https://github.com/lanl/singularity-eos/issues/539#issuecomment-3166312819


jhp-lanl:
@Yurlungur can you review before merging? I don't have as much experience in the root finder, but this seems like an appropriate change

Yurlungur:
Yep this is reasonable. Thanks for the fix!

Issue #542, opened 2025-08-15
Expose several additional options in sesame2spiner


Yurlungur:
It would be useful to expose
- `USE_MAXWELL_TABLE` to get the maxwell constructed table if available
- `EOS_SPLIT_COWAN` to turn on Cowan splitting explicitly for electron/ion splitting if needed
- `EOS_CALC_FREE_ENERGY` to get a free energy if it's unavailable even if it's less good than a first-principles version

Issue #541, opened 2025-08-15
Add extrapolation to rho = 0 for spiner tables


Yurlungur:
Right now spiner's logarithmic interpolation means that it can never extrapolate to exactly rho = 0. An easy fix would be to check if rho < the minimum density on the table, and if it is compute pressure as
$P(rho, T) = \rho \Gamma(\rho_m, T)$
where here $\rho_m$ is the minimum density and $\Gamma$ is the Gruneisen coefficient. And one could do a similar thing for temperature.

Thanks @jhp-lanl for reminding me of this issue.

PR #543, opened 2025-08-15
properly handle vapor dome for Maxwell-constructed EOS's

Yurlungur:
## PR Summary

This MR pre-empts #538 . In investigation of the PTE solver for riot, I noticed that table density often determined whether or not a solver converged. It turned out that this was partly due to the `RhoPMin` machinery not handling the vapor dome of a Maxwell constructed EOS correctly. The vapor dome is problematic for a PTE solver because the Maxwell construction flattens out the slope of $\partial P/\partial \rho$ in that region. This means the Jacobian for the Newton solver may have a trivial direction, which can prevent the solver from escaping the region and finding a different solution. On the other hand, the Maxwell constructed region may contain physically meaningful states we wish to access, so we usually don't want to completely exclude it.

There was another problem, which was that `RhoPMin` was reporting seemingly random values for a Maxwell constructed EOS, where it should report 0. I have resolved these issues with several changes:

1. I add a floor to $\partial P/\partial \rho$ so that the Jacobian doesn't trivialize. In practice, this is actually a *ceiling* on $\partial P/\partial\alpha$ for volume fraction $\alpha$ (for most solvers). This ceiling can be set by the user with the `MixParams` struct.
2. I modify the `rho_at_pmin` logic in the spiner tables to correctly detect a Maxwell constructed EOS (or gas EOS) and in those cases report that the density at Pmin is 0.
3. *optionally* rho_at_pmin may instead be interpreted to report the edge of the vapor dome. It only does so if requested when the EOS is constructed. 

Note that there's also an unfortunate interaction between sesame and eospac for Maxwell constructed equations of state, which is that because eospac's rational function interpolation can overshoot at the edge of the vapor dome, a Maxwell constructed EOS may still have a region of negative pressure near the edge of the vapor dome, when interpreted as a continuous function. It's dangerous to still set rho_at_pmin to 0 in this case, because there genuinely is a region of "negative slope" in the continuum representation output by eospac.

I think fixing this would require modification of EOSPAC, or leveraging a priori knowledge of whether or not a table is Maxwell constructed. This might be possible in `sesame2spiner` in some cases, as I discuss in #542 . For now I let it be. The `rho_at_pmin` algorithm will detect that rho_at_pmin = 0 for most temperatures where eospac doesn't overshoot, but may find a finite rho_at_pmin for some temperatures. I just let this happen.

It may also be worth adding an ability to Spiner to extrapolate to exactly zero pressure/density as described in #541 .

One advantage of these changes is that the PTE solver is now more robust (again). IN particular, some dependencies on spiner table density are now relaxed and coarser tables are possible. Also we can now provide wildly incorrect temperature guesses, for example we can guess a temperature very close to 0 K and the solver will still converge for a difficult state.

Thanks to @jhp-lanl and @pdmullen for your feedback as I worked through the right thing to do here.

As an aside, I have a few more robustness tweaks I'm working on for the PTE solver, so keep an eye out for that. Then it's probably time for another release.

## PR Checklist

### DIFF
--- a/doc/sphinx/src/models.rst
+++ b/doc/sphinx/src/models.rst
@@ -1695,18 +1695,24 @@ The constructor for ``SpinerEOSDependsRhoT`` is given by two overloads:
 
   SpinerEOSDependsRhoT(const std::string &filename, int matid,
                        const singularity::TableSplit = singularity::TableSplit::Total,
-                       bool reproduciblity_mode = false);
+                       bool reproduciblity_mode = false,
+                       bool pmin_vapor_dome = false);
   SpinerEOSDependsRhoT(const std::string &filename, const std::string &materialName,
                        const singularity::TableSplit = singularity::TableSplit::Total,
-                       bool reproducibility_mode = false);
+                       bool reproducibility_mode = false,
+                       bool pmin_vapor_dome = false);
 
 where here ``filename`` is the input file, ``matid`` is the unique
 material ID in the database in the file, ``materialName`` is the name
 of the material in the file, and ``reproducability_mode`` is a boolean
 which slightly changes how initial guesses for root finds are
 computed. The ``TableSplit`` option selects electron, ion, or total
-equation of state tables for partial ionization. The constructor for
-``SpinerEOSDependsRhoSie`` is identical.
+equation of state tables for partial ionization. ``pmin_vapor_dome``
+specifies how to interpret the ``RhoPMin`` function. If it is false
+(the default), it is the usual interpretation. If it is ``true`` and
+the equation of state is Maxwell constructed, then ``RhoPMin`` for
+this EOS will return compression edge of the vapor dome. The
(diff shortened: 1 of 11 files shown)


Yurlungur:
> > I think fixing this would require modification of EOSPAC, or leveraging a priori knowledge of whether or not a table is Maxwell constructed. This might be possible in sesame2spiner in some cases, as I discuss in #542 . For now I let it be. The rho_at_pmin algorithm will detect that rho_at_pmin = 0 for most temperatures where eospac doesn't overshoot, but may find a finite rho_at_pmin for some temperatures. I just let this happen.
> 
> We could tabulate the 411 table (if it exists) and check for whether we're inside or outside the vapor dome to compensate for these artificial concavities, but I think there are smarter things we can do with that information (multiphase fluid).

Yeah I think there's some stuff we could do in the `sesame2spiner` context. And it might be worth pursuing. But I think that's somewhat linearly independent from the changes in this MR.

PR #545, opened 2025-08-19
reduced rho t system and solver improvements

Yurlungur:
## PR Summary

Recently I've been thinking about the PTE solver and how performance and robustness of the linear solve can impact an integrated system. This has motivated a bunch of my recent MRs, and this one.

In this MR, I observe that the $\rho-T$ system can be reduced from $N+1$ equations to $N$ equations by picking a material $s$ and computing that material's volume fraction as 
$\alpha_s = 1 - \sum_{m!=s} \alpha_m$
In the 2-material case, this has the significant advantage of making the Jacobian a 2x2 matrix, just like with `PTESolverPT`. That lets us use the more performant and accurate Cramer's rule (which is always available, but only better in the 2x2 case). For small numbers of materials, the smaller system can make a big difference in time-to-solution, since a 2x2 matrix inversion is something like 3x faster than a 3x3. This has the added bonus of unifying the matrix inversion logic that I wrote for the `PTESolverPT` case.

This does come with some pitfalls, however. For one thing, the material $s$ needs to be chosen. It's not obvious how to pick it... So I default to the material with the largest mass fraction, which seems to perform better in the few experiments I tried. I also implemented the ability for $s$ to be set downstream via the `MixParams` struct.

Another pitfall is the risk of catastrophic cancellation in the sum of the volume fractions (excluding $\alpha_s$). To get around this issue, I implement [Kahan summation](https://en.wikipedia.org/wiki/Kahan_summation_algorithm) (and sprinkle it throughout the rest of the codebase where I think it's useful). 

I also wanted to reduce problems that might come from inverting the Jacobian matrix when it's ill conditioned. To help with that, I also implement a simple modification of the classic Jacobi pre-conditioner, where each element of the diagonal of my pre-conditioning matrix is one over the *maximum* of the corresponding row of the matrix $A$. It *seems* like this pre-conditioner improves the condition number of the matrices we encounter, but I'm not sure whether or not it actually improves our ability to converge. But it's cheap, so I left it in.

I also documented and tested all these changes.

## PR Checklist

- [x] Adds a test for any bugs fixed. Adds tests for new features.
- [x] Format your changes by using the `make format` command after configuring with `cmake`.
- [x] Document any new features, update documentation for changes made.
- [x] Make sure the copyright notice on any files you modified is up to date.
- [x] After creating a pull request, note it in the CHANGELOG.md file.
- [x] LANL employees: make sure tests pass both on the github CI and on the Darwin CI

If preparing for a new release, in addition please check the following:
[...]

PR #548, opened 2025-08-22
Robustness improvements to the reduced rho-T system and clarity in residual tol

Yurlungur:
## PR Summary

My previous MR #545 implemented the reduced rho-T system. However, I noticed a robustness issue which is that excluding the material with the largest mass fraction can be led astray when, for example, a heavy material is in a mixed cell with a light one. To improve robustness, the solver now switches which material to exclude every cycle. I add additional tests to add confidence for this, including a mixed Helium/Copper cell.

I also spent a lot of time thinking about what on earth the residual tolerance in the solver means and is doing. My take is the following: it is checking whether or not we consider the solver to be "good enough" after it starts to fail to find a gradient descent direction. This is indeed a common fail-safe/fallback in solvers, including in, e.g., numerical recipes. 

I revamped this system a little bit to make this functionality more clear and more fine-grained. I add tolerances for each component of the residual, now called `*_sufficient`. When `CheckPTE` is called, it checks both for convergence and "good enough" based on this criteria. The "good enough" condition is only utilized, however, if the residual is not decreasing sufficiently between steps. I made this user-settable and a much tighter check by default.

Note that the default tolerance for the `sufficient` thresholds may look looser than `residual_thresh` looked but it's actually tighter. `residual_thresh` is on the *square* of the residual, because that's how it's formulated in the Newton solver literature, where-as the `_sufficient` thresholds are essentially an $L_1$ or $L_2$ norm. 

I also removed the absolute tolerances for energy and volume fraction, as they did nothing, because energy and volume fraction are normalized the absolute and relative tolerances are identical.

Finally I tweaked the `pte_2mat` example a little bit based on my current experiments. It reports more information now, including timing.

## PR Checklist

### DIFF
--- a/doc/sphinx/src/using-closures.rst
+++ b/doc/sphinx/src/using-closures.rst
@@ -659,29 +659,40 @@ contains the following member fields, with default values:
 
 .. code-block:: cpp
 
-  struct MixParams {
-    bool verbose = false; // verbosity
-    Real derivative_eps = 3.0e-6;
-    Real pte_rel_tolerance_p = 1.e-6;
-    Real pte_rel_tolerance_e = 1.e-6;
-    Real pte_rel_tolerance_t = 1.e-4;
-    Real pte_abs_tolerance_p = 0.0;
-    Real pte_abs_tolerance_e = 1.e-4;
-    Real pte_abs_tolerance_t = 0.0;
-    Real pte_residual_tolerance = 1.e-8;
-    std::size_t pte_max_iter_per_mat = 128;
-    Real line_search_alpha = 1.e-2;
-    std::size_t line_search_max_iter = 6;
-    Real line_search_fac = 0.5;
-    Real vfrac_safety_fac = 0.95;
-    Real temperature_limit = 1.0e15;
-    Real default_tguess = 300.;
-    Real min_dtde = 1.0e-16;
-    std::size_t pte_small_step_tries = 2;
-    Real pte_small_step_thresh = 1e-16;
-    Real pte_max_dpdv = -1e-8;
-    std::int64_t pte_reduced_system_exclude_idx = -1;  // choose index with largest mass fraction
-  };
+struct MixParams {
+  bool verbose = false;
+  bool iterate_t_guess = true;
+  Real derivative_eps = 3.0e-6;
+  Real pte_rel_tolerance_p = 1.0e-6;
+  Real pte_rel_tolerance_t = 1.0e-4;
+  Real pte_rel_tolerance_e = 1.0e-6;
+  Real pte_rel_tolerance_v = 1.0e-6;
+  Real pte_abs_tolerance_p = 1.0e-6;
+  Real pte_abs_tolerance_t = 1.0e-4;
+
+  Real pte_slow_convergence_thresh = 0.99;
+  Real pte_rel_tolerance_p_sufficient = 1.e2 * pte_rel_tolerance_p;
+  Real pte_rel_tolerance_t_sufficient = 10 * pte_rel_tolerance_t;
+  Real pte_rel_tolerance_e_sufficient = 1.e2 * pte_rel_tolerance_e;
+  Real pte_rel_tolerance_v_sufficient = 1.e2 * pte_rel_tolerance_v;
+  Real pte_abs_tolerance_p_sufficient = 1.e2 * pte_abs_tolerance_p;
+  Real pte_abs_tolerance_t_sufficient = 10 * pte_abs_tolerance_t;
+
+  std::size_t pte_max_iter_per_mat = 128;
+  Real line_search_alpha = 1.e-2;
+  std::size_t line_search_max_iter = 6;
+  Real line_search_fac = 0.5;
+  Real vfrac_safety_fac = 0.95;
+  Real temperature_limit = 1.0e15;
+  Real default_tguess = 300.;
+  Real min_dtde = 1.0e-16;
+  std::size_t pte_small_step_tries = 2;
+  Real pte_small_step_thresh = 1e-16;
+  Real pte_max_dpdv = -1e-8;
+
+  // choose index with largest mass fraction
+  std::int64_t pte_reduced_system_exclude_idx = -1;
+};
 
 where here ``verbose`` enables verbose output in the PTE solve is,
 ``derivative_eps`` is the spacing used for finite differences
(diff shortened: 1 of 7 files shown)


singularity-eos/closure/mixed_cell_models.hpp
```diff
@@ -207,15 +216,18 @@ bool solve_Ax_b_wscr(const std::size_t n, Real *a, Real *b, Real *scr) {
...
-PORTABLE_INLINE_FUNCTION Real sum_neumaier(Data_t &&data, std::size_t n,
-                                           std::size_t offset = 0,
-                                           std::size_t iskip = -1) {
+struct IdentityOperator {
+  PORTABLE_FORCEINLINE_FUNCTION Real operator()(const Real x) const { return x; }
+};
+template <typename Data_t, typename Operator_t = IdentityOperator>
+PORTABLE_FORCEINLINE_FUNCTION Real
+sum_neumaier(Data_t &&data, std::size_t n, std::size_t offset = 0, std::size_t iskip = -1,
+             const Operator_t &op = IdentityOperator()) {
```
Yurlungur, line 225: @dholladay00 Kahan summation now capable of taking unary operators

singularity-eos/closure/mixed_cell_models.hpp
```diff
@@ -1148,6 +1183,15 @@ class PTESolverPT
...
   }
 
+  PORTABLE_INLINE_FUNCTION
+  void Finalize() {
+    for (std::size_t m = 0; m < nmat; m++) {
+      press[m] = Pequil * uscale;
+      temp[m] = Tequil * Tnorm;
+      u[m] *= uscale;
+    }
+  }
```
Yurlungur, line 1193: This was a bug in the PTESolverPT. If the solver was passed a PTE solution, it would not overwrite the pressures with `Pequil` and they might be uninitialized memory. This ensures that can't happen.

singularity-eos/closure/mixed_cell_models.hpp
```diff
@@ -2076,16 +2134,19 @@ PORTABLE_INLINE_FUNCTION SolverStatus PTESolver(System &s) {
...
+        close_enough) {
       converged = true;
+      status.slow_convergence_detected = true;
       break;
     }
   }
-  // Call it converged even though CheckPTE never said it was because the residual is
-  // small.  Helps to avoid "failures" where things have actually converged as well as
-  // finite precision allows
-  if (!converged && err < residual_tol) converged = true;
```
Yurlungur, line 2088: this was redundant with the block above. I unified them.

doc/sphinx/src/using-closures.rst
```diff
@@ -659,29 +659,40 @@ contains the following member fields, with default values:
...
+  Real temperature_limit = 1.0e15;
+  Real default_tguess = 300.;
+  Real min_dtde = 1.0e-16;
+  std::size_t pte_small_step_tries = 2;
+  Real pte_small_step_thresh = 1e-16;
+  Real pte_max_dpdv = -1e-8;
+
+  // choose index with largest mass fraction
+  std::int64_t pte_reduced_system_exclude_idx = -1;
+};
```
dholladay00, line 695: This thing is getting big, is it in global memory?

doc/sphinx/src/using-closures.rst
```diff
@@ -659,29 +659,40 @@ contains the following member fields, with default values:
...
+  Real line_search_fac = 0.5;
+  Real vfrac_safety_fac = 0.95;
+  Real temperature_limit = 1.0e15;
+  Real default_tguess = 300.;
+  Real min_dtde = 1.0e-16;
+  std::size_t pte_small_step_tries = 2;
+  Real pte_small_step_thresh = 1e-16;
+  Real pte_max_dpdv = -1e-8;
+
+  // choose index with largest mass fraction
```
pdmullen, line 693: Should this instead be largest volume fraction?

doc/sphinx/src/using-closures.rst
```diff
@@ -690,8 +701,31 @@ evaluations of equations of state when building a jacobian. The
...
+example, digits may be lost in the Jacobian inversion or the residual
+itself may suffer catastrophic cancellation. In these cases, the user
+may still wish to call this a success. This is the reason for the
+``pte_slow_convergence_thresh`` and ``pte_*_tolerance_*_sufficient``
+variables. If between two iterations the residual does not decrease to
+``pte_slow_convergence_thresh`` times its previous value or smaller,
+**and** the errors in the relevant variables are within the
+``_sufficient`` tolerances, the solver will consider that "good
+enough" and report succesful convergence. In general these should be
+set to an order of magnitude or two less strict than the desired
```
pdmullen, line 720: Should we warn if they are not?

pdmullen (approved):
LGTM!

doc/sphinx/src/using-closures.rst
Yurlungur: The struct is passed in to a solver when it's constructed, so each solver owns a copy. Are you concerned about that adding up? I think it's still trivial compared to the size of the state vector + scratch.

We could potentially split this into one params object per PTE solver... we could reduce the size that way. Made an issue #550

doc/sphinx/src/using-closures.rst
Yurlungur: Hmm... I don't think so. Setting them tighter than that is essentially saying "don't use these, only accept convergence if the primary convergence is satisfied" which is a reasonable way to run the code, I think?

pdmullen: Oh I meant moreso just some guardrails on what users are permitted to do, e.g., 

PORTABLE_REQUIRE(_sufficient > counterpart, "This configuration doesn't make sense");

But frankly your documentation of these should be good enough

Yurlungur:
Tests passing on re-git. I also changed the Eigen solver to pivoted QR since it turned out to be a one-line change. 

I note also that I briefly chatted with the Kokkos developers I think it would be pretty straightforward to implement a pivoted QR in Kokkos. I might go ahead and do that.

Yurlungur:
I'm going to merge this to facilitate some time sensitive changes in riot, but I will address additional feedback in a follow-up MR if necessary.

singularity-eos/closure/mixed_cell_models.hpp
```diff
@@ -259,15 +272,11 @@ class PTESolverBase {
...
-    Real vsum = sum_neumaier(vfrac, nmat);
-    for (std::size_t m = 0; m < nmat; ++m)
-      vfrac[m] *= robust::ratio(vfrac_total, vsum);
-  }
+  virtual void Fixup() { NormalizeVfrac(); }
   // Finalize restores the temperatures, energies, and pressures to unscaled values from
   // the internally scaled quantities used by the solvers
   PORTABLE_INLINE_FUNCTION
-  void Finalize() {
+  virtual void Finalize() {
```
jhp-lanl, line 279: We're testing this on GPU, so I guess this works, right?

dholladay00: I don't think we provide overrides to these implementations. My preference is to remove the keyword if it is not needed.

PR #559, opened 2025-09-04
document the intent of the virtual keyword in solvers

Yurlungur:
## PR Summary

This addresses @dholladay00 's concern about MR #548 . After some discussion offline, I believe this comment covers concerns.

## PR Checklist

### DIFF
--- a/doc/sphinx/src/contributing.rst
+++ b/doc/sphinx/src/contributing.rst
@@ -633,6 +633,12 @@ Also note that any new functionality added to the base class needs to be
 mirrored in the :ref:`Variant class <variant section>` so that it is accessable
 when using the ``EOS`` type.
 
+The solvers in ``closure/mixed_cell_models.hpp`` also use inheritence
+but again here the intent is **not** to use run-time polymorphism. The
+base class simply provides (potentially overridable) functionality to
+the derived classes. But you should always instantiate a derived
+class, **not** a pointer to a base class.
+
 .. _CRTP: https://www.fluentcpp.com/2017/05/12/curiously-recurring-template-pattern/
 
 Fast Logs and Approximate Log Gridding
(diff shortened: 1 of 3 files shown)


dholladay00 (commented):
Could you put a comment right above each instance of `virtual` pointing to this comment?

doc/sphinx/src/contributing.rst
```diff
@@ -633,6 +633,12 @@ Also note that any new functionality added to the base class needs to be
 mirrored in the :ref:`Variant class <variant section>` so that it is accessable
 when using the ``EOS`` type.
 
+The solvers in ``closure/mixed_cell_models.hpp`` also use inheritence
```
dholladay00, line 636: good call to add it to the documentation!

Yurlungur:
There are only two and they're right below the comment:
```C++
  // A note on the virtual functions below. We are NOT using runtime
  // polymorphism. The point of these virtual functions is to provide
  // default implementations for functions that are shared between
  // solvers. But a given solver should be treated by value, not with
  // pointers, and avoid relocatable device code.

  // Fixup is meant to be a hook for derived classes to provide arbitrary manipulations
  // after each iteration of the Newton solver.  This version just renormalizes the
  // volume fractions, which is useful to deal with roundoff error.
  PORTABLE_INLINE_FUNCTION
  virtual void Fixup() { NormalizeVfrac(); }
  // Finalize restores the temperatures, energies, and pressures to unscaled values from
  // the internally scaled quantities used by the solvers
  PORTABLE_INLINE_FUNCTION
  virtual void Finalize() {
    for (std::size_t m = 0; m < nmat; m++) {
      temp[m] *= Tnorm;
      u[m] *= uscale;
      press[m] *= uscale;
    }
  }
```
is that sufficient?

PR #563, opened 2025-09-10
Fix for Carnahan-Starling EOS with DensityFromPressureTemperature

bgclayton-lanl:
## PR Summary
This fix corrects some of the fails with the analytic EOS tests regarding the Carnahan-Starling. Mainly it swaps our the root solver for the `regular_falsi` method. I haven't gone through the analysis to determine if the root is unique though. Maybe someone else has done this analysis before?

## PR Checklist


singularity-eos/eos/eos_carnahan_starling.hpp
```diff
@@ -98,20 +98,36 @@ class CarnahanStarling : public EosBase<CarnahanStarling> {
...
+
+    //auto status = findRoot(poly, press, guess, lo_bound, hi_bound, xtol, ytol, real_root);
+    if (status != RootFinding1D::Status::SUCCESS) {
       // Root finder failed even though the solution was bracketed... this is an error
       EOS_ERROR("*** (Warning) DensityFromPressureTemperature :: Convergence not met in "
                 "Carnahan-Starling EoS (root finder util) ***\n");
-      real_root = -1.0;
+      rho = -1.0;
     }
-    return std::max(robust::SMALL(), real_root);
```
Yurlungur, line 114: Why this change? I think there should be a floor on the returned density, it shouldn't be possible to return a negative one. I would suggest
```suggestion
return std::max(rho_low, real_root);
```

Yurlungur (commented):
Thanks for the fix! This looks like this is introducing test failures, but that might just be formatting changes. Can you run 
```bash
make format_singularity-eos
```
in the build directory? Hopefully that's why tests are failing. But we might need to make other changes to see. After tests pass on github, we'll also need to make sure they pass on LANL systems by triggering tests on re-git.

You should also add a changelog entry.

bgclayton-lanl:
> Thanks for the fix! This looks like this is introducing test failures, but that might just be formatting changes. Can you run
> 
> ```shell
> make format_singularity-eos
> ```
> 
> in the build directory? Hopefully that's why tests are failing. But we might need to make other changes to see. After tests pass on github, we'll also need to make sure they pass on LANL systems by triggering tests on re-git.
> 
> You should also add a changelog entry.

I currently don't have a way to run the clang formatting thing. I need to talk with Jeff more about this and see the best way to set this up. I also don't have access to run the re-git process either. But I'll go ahead and update the Changelog. Thank you.

Yurlungur:
Ok Jeff or I can run the formatting and trigger the tests

bgclayton-lanl:
@Yurlungur @jhp-lanl I think I've gotten everything fixed. Let me know if I need to change anything else.

bgclayton-lanl:
Okay, sorry for all the issues. It started opening of the door to other bigger questions. The more fundamental question is how to define convergence criteria. But in any case, all the tests are passing, and I updated the tests to include a unit scaling (based on the reference state) on the epsilon for checking errors. Let me know if there are any other changes to do.

@jhp-lanl @Yurlungur

Yurlungur (approved):
This seems entirely sensible. Thanks for the careful digging and nice work @bgclayton-lanl !

@jhp-lanl or I will run formatting and trigger tests on re-git.

jhp-lanl (approved):
One of the jobs failed, but I'm rerunning it

PR #574, opened 2025-10-02
[trivial] gpu compatibility patch for in-tree builds

Yurlungur:
The last time the GPU compatibility patch was updated for HIP, it was only updated in the spackage, not the version used for in-tree builds. Probably we should centralize it so there's only one location, rather than three copies floating around. For now, though, I just copy the one in the spackage into the location used by cmake for in-tree builds.

## PR Checklist


Yurlungur:
This was required to get in-tree builds working on HIP

PR #607, opened 2026-03-02
siecvfrompreferred

Yurlungur:
## PR Summary

While tamping down on edge cases in EAP, @jhp-lanl discovered that the initialization routines in the PTE solver assume that internal energy cannot be computed from pressure and temperature. However, we have an introspection mechanism in singularity-eos for this purpose, `eos.PreferredInput.` I thus modify the initialization routines in the PTE solver to properly note this, and short circuit a potentially expensive or less robust table inversion for PT tables. I also fix a bug in the `GetPressureFromPreferred` calls, where `PreferredInput` was being compared exactly, rather than using a bitwise mask. 

To test this, I recognize that ideal gas actually doesn't care what input you give it. Density, energy, temperature, and pressure are all equally good. I thus modify it to report all of them as preferred inputs. This automatically passes this machinery through the ideal gas PTE test. Note this isn't a particularly robust test of this modification of the machinery, but I think it's the best we can do without introducing a truly pressure-based EOS model.

## PR Checklist

### DIFF
--- a/singularity-eos/closure/mixed_cell_models.hpp
+++ b/singularity-eos/closure/mixed_cell_models.hpp
@@ -538,20 +538,20 @@ class PTESolverBase {
   // depending on what the EOS wants.
   template <typename EOS_t, typename Indexer_t>
   PORTABLE_FORCEINLINE_FUNCTION static void
-  GetSieCvFromTAndPreferred(const EOS_t &eos, const Real rho, const Real P, const Real T,
+  GetSieCvFromTAndPreferred(const EOS_t &eos, Real rho, const Real P, const Real T,
                             Indexer_t lambda, Real &sie, Real &cv) {
     if (eos.PreferredInput() & thermalqs::pressure) {
       Real Pmin = eos.MinimumPressure();
       Real Pmax = eos.MaximumPressureAtTemperature(T);
       Real Pguess;
-      if (std::isnan(P)) { // note this is dangerous
+      if (error_utils::bad_value(Pguess, "Pressure in GetSieCvFromTAndPreferred")) {
         Pguess = 0.5 * (Pmin + Pmax);
       } else {
         Pguess = std::max(Pmin, std::min(Pmax, P));
       }
-      eos.InternalEnergyFromDensityPressure(rho, Pguess, sie, lambda);
+      eos.DensityEnergyFromPressureTemperature(Pguess, T, lambda, rho, sie);
       cv = eos.SpecificHeatFromDensityInternalEnergy(rho, sie, lambda);
-    } else { // if (eos.PreferredInput() & thermalqs::temperature) {
+    } else { // use density, not pressure.
       sie = eos.InternalEnergyFromDensityTemperature(rho, T, lambda);
       cv = eos.SpecificHeatFromDensityTemperature(rho, T, lambda);
     }


singularity-eos/closure/mixed_cell_models.hpp
```diff
@@ -520,18 +523,40 @@ class PTESolverBase {
...
+        Pguess = 0.5 * (Pmin + Pmax);
+      } else {
+        Pguess = std::max(Pmin, std::min(Pmax, P));
+      }
+      eos.InternalEnergyFromDensityPressure(rho, Pguess, sie, lambda);
+      cv = eos.SpecificHeatFromDensityInternalEnergy(rho, sie, lambda);
+    } else { // if (eos.PreferredInput() & thermalqs::temperature) {
+      sie = eos.InternalEnergyFromDensityTemperature(rho, T, lambda);
+      cv = eos.SpecificHeatFromDensityTemperature(rho, T, lambda);
+    }
```
adamdempsey90:
```suggestion
    } else if (eos.PreferredInput() & thermalqs::temperature) {
      sie = eos.InternalEnergyFromDensityTemperature(rho, T, lambda);
      cv = eos.SpecificHeatFromDensityTemperature(rho, T, lambda);
    } else if (eos.PreferredInput() & thermalqs::specific_internal_energy) {
       if (do_e_lookup) {
        sie = eos.InternalEnergyFromDensityTemperature(rho, T, lambda);
      }
      cv = eos.SpecificHeatFromDensityInternalEnergy(rho, T, lambda);
    } else {
       // abort?
    }
```
Is it guaranteed that an eos that prefers SIE will not get this far?

singularity-eos/closure/mixed_cell_models.hpp
```diff
@@ -520,18 +523,40 @@ class PTESolverBase {
...
     Real P{};
-    if (eos.PreferredInput() ==
-        (thermalqs::density | thermalqs::specific_internal_energy)) {
+    if (eos.PreferredInput() & thermalqs::specific_internal_energy) {
       if (do_e_lookup) {
         sie = eos.InternalEnergyFromDensityTemperature(rho, T, lambda);
       }
       P = eos.PressureFromDensityInternalEnergy(rho, sie, lambda);
-    } else if (eos.PreferredInput() == (thermalqs::density | thermalqs::temperature)) {
+    } else { // if (eos.PreferredInput() & thermalqs::temperature) {
```
jhp-lanl: Good catch on the `else`

singularity-eos/closure/mixed_cell_models.hpp
jhp-lanl: I disagree with this change.

As I understand, I think that you're taking into account the fact that ideal gasses can now have any combination of inputs as preferred, so you don't want to _always_ make an ideal gas guess 1.0e8 for the pressure. Okay I agree with that at least, but you're also going to great lengths to make sure an ideal gas _can't_ use a pressure guess. That I don't agree with. Plus, this complexity in the logic is a bad code smell to me.

I would instead propose asking whether a provided pressure guess is _useful_ (i.e. it's a preferred input for this EOS) and _reasonable_. One way to do this would be to just say
```c++
if (eos[m].PreferredImnput() & thermalqs::pressure && !(error_utils::bad_value(press[m], "press[m]")) {
  press[m] = robust::ratio(press[m], uscale);
} else {
```
This way you have a reliable way to tell the code _not_ to use a pressure guess while also still guarding against FPEs. You don't get the automatic pressure guess of 1.0e8, but with the change to `GetPressureFromPreferred` you will at least get something even if it's more expensive.

singularity-eos/closure/mixed_cell_models.hpp
```diff
@@ -520,18 +523,40 @@ class PTESolverBase {
...
+  template <typename EOS_t, typename Indexer_t>
+  PORTABLE_FORCEINLINE_FUNCTION static void
+  GetSieCvFromTAndPreferred(const EOS_t &eos, Real rho, const Real P, const Real T,
+                            Indexer_t lambda, Real &sie, Real &cv) {
+    if (eos.PreferredInput() & thermalqs::pressure) {
+      Real Pmin = eos.MinimumPressure();
+      Real Pmax = eos.MaximumPressureAtTemperature(T);
+      Real Pguess;
+      if (error_utils::bad_value(Pguess, "Pressure in GetSieCvFromTAndPreferred")) {
+        Pguess = 0.5 * (Pmin + Pmax);
```
jhp-lanl: This is UB. I assume you mean
```suggestion
      Real Pguess;
      if (error_utils::bad_value(P, "Pressure in GetSieCvFromTAndPreferred")) {
```

jhp-lanl: Duplicate comment

singularity-eos/closure/mixed_cell_models.hpp
Yurlungur: This is a good cleanup and I agree. Thanks for the suggestion!

Issue #610, opened 2026-03-05
Requirements for PTE Solver


jhp-lanl:
As part of my work debugging EOS states coming out of the singularity-eos PTE solver, I realized that some of my expectations for the PTE solver weren't being met, and the returned state could possibly be problematic. As a result, I wanted to codify some of my expectations after some discussions with @Yurlungur @jonahm-LANL  .

### 1. Material mass should be constant

We can define $\overline{\rho}_m = \alpha_m \rho_m$, where $\rho_m$ is the material "microphysical" or "primitive" density (i.e. material mass per material volume), $\alpha_m$ is the volume fraction of that material, and $\overline{\rho}_m$ is referred to as the "conserved" density (i.e. material mass per _control volume_). **Conserving material mass implies that $\overline{\rho}_m$ is constant.**. The original $\overline{\rho}_m$ values should be recovered after the PTE solver returns. Another way of stating this is that the mass fractions, $\mu_i = \overline{\rho}_m /  \rho_{tot} $, are constant.

This requirement implies
```math
\sum \mu_m = 1
```
```math
\sum \overline{\rho}_m = \rho_\mathrm{tot}
```
```math
\sum \rho_m \alpha_m = \rho_\mathrm{tot}
```

Since the [singularity-eos PTE equations](https://lanl.github.io/singularity-eos/main/src/using-closures.html#pressure-temperature-equilibrium) are posed in terms of volume fractions and component energies, this then implies that
```math
\sum \alpha_m \neq 1
```
```math
\sum u_m = \sum \overline{\rho}_m \epsilon_m \neq u_\mathrm{tot} = \rho_\mathrm{tot} \epsilon_\mathrm{tot}
```
during intermediate stages of the PTE solver. However, if a factor is used to renormalize the volume fractions _and_ its inverse is also used to scale the microphysical densities, then mass should remain conserved. @jonahm-LANL @Yurlungur this may mean that the existing renormalization is fine.

NOTE: the above link contains an error. The material energy densities should be given by $u_m = \overline{\rho}_m \epsilon_m$, so $u_m \neq \rho_m \epsilon_m$.

### 2. The error associated with a given PTE solver should be consistently located in the outputs

There will always be an error associated with the PTE state that is determined by the convergence tolerances. The _location_ of this error will be PTE method dependent.

For example, in the density-temperature PTE solver, the pressure and energy for a given material are determined by the density and temperature. Therefore there is no error in the output material temperatures, and the volume fractions can be calculated as
[...]

PR #615, opened 2026-03-12
Add pte conservation introspection, detailed documentation, resolves issue #610

Yurlungur:
## PR Summary

In #610, @jhp-lanl reported that the errors in, e.g., mass fractions, out of the PTE solver were not what he was expecting. This was coming from two related issues: 
1. We were not adequately clear in our documentation (or perhaps in our thinking) about what is expected to be conserved up to machine precision. Of course  In general, all solvers should preserve that $\bar{\rho}_m = \rho_m\alpha_m$ is constant, mass fractions sum to 1, volume fractions sum to 1, and $u = \sum_m \rho_m \alpha_m \varepsilon_m$. However, in general these are satisfied only up to the solver tolerance. What we should expect is that mass fractions are unchanged and sum to 1 up to machine precision. But otherwise, only solvers that use density as an independent variable should preserve that volume fractions sum to 1 up to machine precision, and only solvers that use internal energy as an independent variable should preserve energy up to machine precision. (Note machine precision vs solver tolerance here. That's the difference here in expectations.)
2. We were insisting on re-normalizing volume fractions by default at the end of each PTE solver. This wasn't technically wrong, but I think given the expectations above, it was inconsistent. And by doing so for a solver that does not use density as an independent var will move the error due to tolerance out of the volume fractions (which will now sum to 1) and into the mass fractions, which will no longer match up to machine precision.

I resolve this in a few ways:
1. I expose a new method `SumsExactly` for each solver, which reports which quantities should sum exactly after the solver is done, up to machine precision.
2. I document our expectations and how to use this new functionality.
3. I remove the normalization
4. I add tests that check this, using the new introspection to check only the quantities we expect to be conserved for the solver used.

## PR Checklist

### DIFF
--- a/doc/sphinx/src/using-closures.rst
+++ b/doc/sphinx/src/using-closures.rst
@@ -695,6 +695,30 @@ then :math:`\rho_2` implies an effective pressure
 which now differs from the solver-selected pressure by a factor of
 order :math:`\delta_f`.
 
+The ideal gas is a relatively gentle special case. For stiffer
+equations of state, the issues can be more severe. For example, the
+pressure along a reference curve for the Murnaghan equation of state
+is
+
+.. math::
+
+  P(\rho) = \frac{B_0}{B_0'} \left[\left(\frac{\rho}{\rho_0}\right)^{B_0'} - 1\right]
+
+where the bulk modulus :math:`B_s` is given by a linear function in density
+
+.. math::
+
+   B_s = B_0 + B_0' (\rho - \rho_0)
+
+for some reference density :math:`\rho_0`. In this case, the corrected
+pressure due to modifying the volume fractions has the form
+
+.. math::
+
+  P_2' = P\left(1 + \mathcal{O}(\delta_f)\right)^{B_0'},
+
+which could introduce a very large deviation for :math:`B_0' > 1`.
+
 A similar story holds for attempting to modify the internal energys
 :math:`\epsilon` to sum to the bulk internal energy density. The
 internal energy doesn't necessarily impact mass or volume fractions,


doc/sphinx/src/using-closures.rst
```diff
@@ -564,6 +564,66 @@ material pressures.
...
+
+  PORTABLE_INLINE_FUNCTION
+  constexpr static unsigned long ExactlySum();
+
+which returns a bit array that evaulates to true for quantities that
+are satisfied. The possible flags are
+
+* ``singularity::thermalqs::mass_fractions`` for the mass fraction constraint
+* ``singularity::thermalqs::volume_fractions`` for the volume fraction constraint
+* ``singularity::thermalqs::specific_internal_energy`` for the energy constraint
```
adamdempsey90: i guess a nitpick is that it is internal energy not specific that would be conserved so maybe the name here is a at odds with the usage of mass and volume fractions.

singularity-eos/closure/mixed_cell_models.hpp
```diff
@@ -304,6 +366,10 @@ class PTESolverBase {
   PORTABLE_INLINE_FUNCTION
   const MixParams &GetParams() const { return params_; }
 
+  // Reports which quantities are exactly conserved by a given solver
+  PORTABLE_INLINE_FUNCTION
+  constexpr static unsigned long ExactlySum() { return thermalqs::none; }
```
jhp-lanl: Why is this defined in the base class? It seems kind of useless here and it's going to be overridden anyway, right?

test/test_pte.cpp
```diff
@@ -199,13 +201,54 @@ void TestPTE(const std::string name, const std::size_t nscratch_vars,
...
+              rho_new += rho[m] * vfrac[m];
+            }
+            bool rho_close = isClose(rho_new, rho_tot, 1e-12);
+            if (!rho_close) {
+              printf("Mass fractions no longer sum to 1! "
+                     "rho_bulk = %.14e, true = %.14e\n",
+                     rho_new, rho_tot);
+            }
+            in_pte = in_pte && rho_close;
+          }
```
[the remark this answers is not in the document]
jhp-lanl: Actually, would it be possible to check the final mass fractions or $\overline{\rho}$ against their initial values?

PR #618, opened 2026-03-18
pt derivatives from preferred

### LINKED ISSUE #598: P-T derivatives are needed for accurately computing PTE mixture properties
# Tl; dr

I think we need to expose $P$ - $T$ derivatives for accurately computing bulk mixture properties in PTE.

# Details

## Constant volume heat capacity example in PTE

There is no way I have discovered to accurately calculate mixture derivatives for a PTE system from the corresponding derivatives of the individual components of that system. By derivatives I mean the standard three we compute in singularity-eos: the isentropic bulk modulus, the Gruneisen parameter, and the constant volume heat capacity. To illustrate, we can consider the constant volume heat capacity of the bulk.

Essentially if you start with the Amagat sum of the specific internal energies, $e$,
```math
e_\mathrm{tot} = \sum\limits_{m = 1}^N \mu_m e_m,
```
where $\mu_m$ are the mass fractions, you can differentiate this equation with respect to temperature at constant _bulk_ volume to get the mixture heat capacity:
```math
C_{V, \mathrm{tot}} = \left(\frac{\partial e_\mathrm{tot}}{\partial T}\right)_{V_\mathrm{tot}} = \sum\limits_{m = 1}^N \mu_m \left(\frac{\partial e_m}{\partial T}\right)_{V_\mathrm{tot}}.
```
The issue is that
```math
\left(\frac{\partial e_m}{\partial T}\right)_{V_\mathrm{tot}}
\neq \left(\frac{\partial e_m}{\partial T}\right)_{V_m},
```
so this term cannot be equated to a component heat capacity in the averaging equation.

There are many ways to achieve constant mixture volume, but PTE imposes some path through $N + 2$ phase space ($N$ components $+ 2$ thermodynamic state variables). There are probably existence/uniqueness questions here, but that's beyond the scope of this issue.

For a metal and a gas, you can imagine that the metal will expand as the temperature increases and do work on the gas as the mixture volume is held constant. In so much as the heat capacity measures degrees of freedom, you can imagine how this might open additional degrees of freedom for the mixture beyond those of the individual species, leading to a higher mixture heat capacity. The conclusion then is that the total heat capacity cannot be found from mass-averaging the component heat capacities.

## $P$ and $T$ derivatives can be averaged

For the constant pressure heat capacity though, we are taking the derivative with respect to a quantity that is the same for the mixture and the component (by the definition of PTE). Therefore, taking the derivative of the energy sum yields
```math
C_P = \left(\frac{\partial e_\mathrm{tot}}{\partial T}\right)_P = \sum\limits_{m = 1}^N \mu_m \left(\frac{\partial e_m}{\partial T}\right)_P
= \sum\limits_{m = 1}^N \mu_m C_{P, m},
```
meaning that the constant pressure heat capacity can be averaged.

This is also true for the constant pressure thermal expansion coefficient, isothermal bulk modulus, and isothermal derivative of energy with respect to pressure.

Then other quantities, like the mixture sound speed, must be computed from the $P$ and $T$ derivatives of the mixture.

## Side note for other mixing rules

Generally, you can only average component derivatives for quantities where the value held constant is _also_ a property of the mixture. For a uniform strain closure rule, the component volumes change by the same proportionate amount that is related to $\nabla \cdot \mathbf{v}$. In this case, the assumption of uniform strain imposes the condition that all materials expand or contract at the same time. Therefore constant mixture volume implies constant component volume (and visa versa). This also means that derivatives with respect to the mixture volume can be related to derivatives with respect to the component volumes since there is a simple relation between the two.

For more complex mixing rules, other prescriptions need to be used to relate component derivatives to the bulk response of the mixture.

# Feature request

## Feature

I believe that for mixture quantities to be accurately computed, the EOS models must expose pressure and temperature derivatives. For mixture quantities, I think this is best served by a single function that gives
1. $\left(\partial e / \partial P\right)_T$
2. $\left(\partial \rho / \partial P\right)_T$
3. $\left(\partial e / \partial T\right)_P$
4. $\left(\partial \rho / \partial T\right)_P$

It's also possible to provide all of these individually, but that may increase effort (both developer and computational).

There's also the question of independent variables. One idea would be to provide $\rho_m$, $e_m$, $P$, and $T$ and then allow the EOS model to pick its preferred independent variables. I think we want to avoid unnecessary inversions, and I'd request that $P$ and $T$ be considered possible independent variables.

So a possible function could be
```c++
void PTDerivativesFromPreferred(Real const &density, Real const &sie, Real const &pressure, Real const &temperature,
                                Real &dedP_T, Real &drdP_T, Real &dedT_P, Real &drdT_P) const;
```

## Other considerations

For analytic EOS models, thermodynamic identities can be used to convert to $P$ and $T$ derivatives that can averaged to then compute bulk properties. The only error here is associated with floating point representation of the derivatives and propagation of this error in the equations.

For tabular EOS though, there are multiple considerations:
1. For SESAME data, we can generally assume the grid points for $P(\rho, T$ and $e(\rho, T)$ are essentially faithful representations of the underlying EOS models. However, the table derivatives are always approximations of the underlying EOS model.
2. Thermodynamic identities are predicated upon thermodynamic consistency. In general, there is no guarantee that a given interpolation methods preserves thermodynamic consistency, leading to additional errors associated with using tabular derivatives in thermodynamic identities.
3. In the case of spiner, derivatives are stored directly on the table itself. Even if you could guarantee that the derivatives at grid points are completely accurate, the interpolation is likely to introduce errors associated with the interpolation of the derivative quantity.

Regardless, to me this necessitates exposure of the P-T derivatives so that choices can be made in singularity-eos about _how_ to compute these P-T derivatives.

Yurlungur:
## PR Summary

This MR resolves #598 by adding @jhp-lanl 's suggested function `PTDerivativesFromPreferred`, which I think does make sense. To robustly test, and provide a sensible fallback for methods that don't implement the function, I also implement a finite differences approach.

This MR isn't finished (and we can't merge until our internal CI is back up anyway), but I believe it is ready for feedback. The major todos at this point are:
- [x] Thread the derivatives calculation through EOSPAC
- [ ] Thread derivatives calculation through SpinerEOS. I might put this in a second, later MR because it requires changing sesame2spiner. There's a missing derivative table.
- [x] Test more robustly.

## PR Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,17 @@
 
 ### Added (new features/APIs/variables/...)
 - [[PR618]](https://github.com/lanl/singularity-eos/pull/618) Add PTDerivativesFromPreferred for computing derivatives of a mixture in a cell
+
+### Fixed (Repair bugs, etc)
+
+### Changed (changing behavior/API/variables/...)
+
+### Infrastructure (changes irrelevant to downstream codes)
+
+## Release 1.11.1
+Date: 03/25/2026
+
+### Added (new features/APIs/variables/...)
 - [[PR600]](https://github.com/lanl/singularity-eos/pull/600) Added mass fractions for multiphase tables to sesame2spiner and added MassFractionsFromDensityTemperture/InternalEnergy to the SpinerDependsRhoSIE and SpinerDependsRhoT equations of state.
 - [[PR589]](https://github.com/lanl/singularity-eos/pull/589) InternalEnergyFromDensityPressure
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,7 +24,7 @@ endif()
 # declare the project name
 project(
   singularity-eos
-  VERSION 1.11.0
+  VERSION 1.11.1
   LANGUAGES NONE)
 
 list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake")


singularity-eos/eos/modifiers/eos_unitsystem.hpp
```diff
@@ -214,6 +214,21 @@ class UnitSystem : public EosBase<UnitSystem<T>> {
...
+  template <typename Lambda_t = Real *>
+  PORTABLE_INLINE_FUNCTION void
+  PTDerivativesFromPreferred(const Real rho, const Real sie, const Real P,
+                             const Real temp, Lambda_t &&lambda, Real &dedP_T,
+                             Real &drdP_T, Real &dedT_P, Real &drdT_P) const {
+    t_.PTDerivativesFromPreferred(rho * rho_unit_, sie * sie_unit_, P * press_unit_,
+                                  temp * temp_unit_, lambda, dedP_T, drdP_T, dedT_P,
+                                  drdT_P);
+    // scale outputs
+    dedP_T *= robust::ratio(press_unit_, temp_unit_);
```
jhp-lanl:
```suggestion
    dedP_T *= robust::ratio(press_unit_, energy_unit_);
```

singularity-eos/eos/eos_eospac.hpp
```diff
@@ -1710,6 +1715,40 @@ PORTABLE_INLINE_FUNCTION void EOSPAC::DensityEnergyFromPressureTemperature(
...
+  eosSafeInterpolate(&table, nxypairs, R, T, z, dx, dy, "EofRT", Verbosity::Quiet);
+  const Real dedr_T = sieFromSesame(dx[0]);
+  const Real dedT_r = sieFromSesame(temperatureToSesame(dy[0]));
+
+  table = RofPT_table_;
+  eosSafeInterpolate(&table, nxypairs, P, T, z, dx, dy, "RofPT", Verbosity::Quiet);
+  drdP_T = pressureToSesame(dx[0]); // dividing by pressure means inverse conversion
+  drdT_P = temperatureToSesame(dy[0]); // ditto
+  dedP_T = dedr_T * drdP_T;
+  dedT_P = dedT_r + dedr_T * drdT_P;
```
jhp-lanl: An alternative would be to do an $e(\rho, T)$ lookup and a $P(\rho, T)$ lookup and then use identities to compute the P-T derivatives.

Naively, I'd expect EOSPAC to perform density-temperature lookups since those are the preferred inputs (without pre-inversion).

singularity-eos/eos/eos_variant.hpp
```diff
@@ -280,6 +280,20 @@ class Variant {
         eos_);
   }
 
+  // TODO(JMM): Do we need a vectorized version of this call?
```
jhp-lanl: Maybe in the future? In my head, I'd pair a vectorized version of this call with a vectorized PTE solver.

test/test_eos_ideal.cpp
```diff
@@ -198,6 +198,53 @@ SCENARIO("Ideal gas density energy from prssure temperature",
...
+            Real rho = i;
+            Real T = 100 * i;
+            Real P = device_eos.PressureFromDensityTemperature(rho, T);
+            Real sie = device_eos.InternalEnergyFromDensityTemperature(rho, T);
+            Real dedP_T, drdP_T, dedT_P, drdT_P;
+            Real dedP_T_fd, drdP_T_fd, dedT_P_fd, drdT_P_fd;
+            device_eos.PTDerivativesFromPreferred(rho, sie, P, T,
+                                                  static_cast<Real *>(nullptr), dedP_T,
+                                                  drdP_T, dedT_P, drdT_P);
+            singularity::eos_base::PTDerivativesByFiniteDifferences(
```
jhp-lanl: Nice!

jhp-lanl (approved):
Minor comments. Looks good! I think this will be really useful for computing multimaterial mixture derivatives.

singularity-eos/eos/eos_spiner_rho_sie.hpp
```diff
@@ -215,6 +215,14 @@ class SpinerEOSDependsRhoSieTransformable
   PORTABLE_INLINE_FUNCTION void
   DensityEnergyFromPressureTemperature(const Real press, const Real temp,
                                        Indexer_t &&lambda, Real &rho, Real &sie) const;
+  /*
+  // TODO(JMM): For now using FD. Fix this.
```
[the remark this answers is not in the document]
Yurlungur: Yeah I was putting off the spiner version because I was going to add additional tables possibly.

PR #632, opened 2026-04-09
Add a SpinerEosDependsRhoSie constructor that uses an existing eos object to generate a spiner eos

buechlerm:
## PR Summary

Much of this was generated by AI.  I had asked it to emulate sesame2spiner in terms of parameters for grid generation.  Seems somewhat heavyweight, but there are defaults.  Ultimately, if somebody proposes a different way of parameterizing the grids, I'd accept that.

## PR Checklist

- [x] If ML was used, make sure to add a disclaimer at the top of a file indicating ML was used to assist in generating the file.
- [x] If Agentic AI was used, have the AI generate a "proposed changes" markdown file and store it in the `plan_histories` folder, with a filename the same as the MR number.


singularity-eos/eos/eos_spiner_rho_sie.hpp
```diff
@@ -1043,6 +1223,460 @@ inline SpinerEOSDependsRhoSieTransformable<
...
+      T_(j, i) = T;
+
+      // Compute pressure and derivatives at constant sie
+      // Use PressureFromDensityInternalEnergy if available for correct derivatives
+      if constexpr (eos_builder::has_P_rho_sie_v<EOS>) {
+        P = source_eos.PressureFromDensityInternalEnergy(rho, sie);
+        dependsRhoSie_.P(j, i) = P;
+
+        // Bulk modulus will be computed by calcBMod_() after derivatives are populated
+        dependsRhoSie_.bMod(j, i) = robust::EPS();
```
buechlerm, line 1307: We can check that the eos provides an isentropic bulk modulus.  Probably should use that.  If it provides Ks, gamma and Cv should we derive the derivatives from those?  Probably.

jhp-lanl: modulo thermodynamic consistency...

buechlerm: Are you suggesting that you prefer finite difference for derivatives and then evaluation of the properties from those?  The real question I am thinking about right now, is what is the minimum we expect in the source_eos.  I'm inclined to allow (for the dependsRhoSIE case) the source_eos to provide as little as PofRE and TofRE to fill in the tables, taking advantage of ks, gruneisen, and cv it it also has those methods.

jhp-lanl: Honestly, I'm not sure. The advantage of finite differences is that they are guaranteed to be robust to thermodynamic consistency issues. The disadvantage is that they are never that accurate.

The worst possible scenario is using finite differences and then trying to use them in thermodynamic derivatives to compute something else. I ran into this in https://github.com/lanl/singularity-eos/pull/526 where I didn't have a good mechanism to get P-T derivatives (hence https://github.com/lanl/singularity-eos/issues/598). The finite differences there when combined into quantities we want (like a sound speed) could result in pretty large errors.

Some of the formulas from [Menikoff and Plohr](https://doi.org/10.1103/RevModPhys.61.75) using these quantities are pretty complex. I'm not sure anybody has looked at the effects of thermodynamic inconsistencies on the accuracy of quantities computed from these quantities. I'd naively think that $\Gamma$ could be the most susceptible since it relates the pressure and energy tables; I think that _can_ be a source of inconsistency in older tables where a change was made to one without propagating the change to the other.

If I had to pick, I think I'd agree with you on using $C_V$, $\Gamma$, and $B_S$ to compute everything else. But I think it's always worth remembering that those formulas are predicated on thermodynamic consistency, and using them without that consistency could lead to strange results.

buechlerm:
Ok, this is effectively feature complete.  I've made choices on what is the _prefered_ path from an incomplete eos to the tables, but I'm not dead set on them, so if somebody has a very strong preference, I can change it.  @adamdempsey90 was also working on some sesame2spiner changes and if he wants a different way of defining the table grids, I'm open to that as well.  Finally, I said feature complete, but a future feature that is probably required is writing an sp5 file (for restarts at minimum).  I'm putting that off till later.

buechlerm:
There are actually a few more features we would want (I just don't need them for my Ristra work).  Here, I punted on split-eos and mass fractions.  We definitely want to optionally support those as well.  Again, punting on those, till later, unless somebody sees something atrocious about this MR that'll make adding those features difficult, later.

doc/sphinx/src/examples.rst
```diff
@@ -58,6 +58,68 @@ internal energy to evaluate at.
 The example demonstrates how to call the pressure, energy, and
 thermodynamic derivatives of a table at that point in phase space.
 
+Custom EOS to SpinerEOS
+-------------------------
+
+The ``examples/custom_eos_to_spiner.cpp`` example demonstrates how to
+tabulate a custom EOS implementation into ``SpinerEOS`` format using
```
Yurlungur, line 65: Probably worth noting it's currently only spiner eos rho sie for now.

plan_histories/MR632-2026-05-04-add-spiner-constructor-from-eos.md
```diff
@@ -0,0 +1,485 @@
...
+
+The constructor will:
+1. Take a source EOS object (template parameter for flexibility)
+2. Take grid parameters (mimicking sesame2spiner options)
+3. Evaluate the source EOS on a grid to populate all required Spiner DataBox tables
+4. Handle offsets automatically for negative values (especially sie)
+5. Compute derivatives (either from source EOS or via finite differences)
+6. Create a fully-functional in-memory SpinerEOS object
+
+**Note on persistence**: The constructed SpinerEOS exists in memory and can be used immediately. While the underlying DataBox objects have `saveHDF()` methods, a convenience method to save the entire SpinerEOS to an HDF5 file in the correct SP5 format would be a useful future enhancement (see Future Enhancements section).
```
Yurlungur: I agree with the robot. For the future though. ;)

plan_histories/MR632-2026-05-04-add-spiner-constructor-from-eos.md
```diff
@@ -0,0 +1,485 @@
...
+### Finite Difference Helper
+
+```cpp
+template <typename Func>
+Real finiteDifference(Func f, Real x, Real fx) {
+  Real h = std::max(std::abs(x) * 1e-6, 1e-12);
+  Real f_plus = f(x + h);
+  Real f_minus = f(x - h);
+  return (f_plus - f_minus) / (2 * h);
+}
```
Yurlungur: I haven't seen the actual implementation, but this may not be sufficiently robust if the base EOS has bounds of validity. For example, if FD is applied near rho of 0 or T of 0, bad things might happen. So we probably want bounds checking.

We could do so by also adding checks for the min/max functions in the EOS base class. If they're present, they're used, otherwise assume a min of 0 for rho and T and unbounded otherwise.

singularity-eos/eos/eos_spiner_rho_sie.hpp
```diff
@@ -76,6 +78,321 @@ using namespace eos_base;
   mitigated by Ye and (1-Ye) to control how important each term is.
  */
 
+// Grid parameters for constructing Spiner tables from generic EOS
+// Defaults match sesame2spiner behavior
+struct SpinerTableGridParams {
```
Yurlungur: It doesn't have to be in this MR. But I think we should unify this functionality with what's in sesame2spiner. I don't like that this is implemented essentially twice. I also don't like that this is in rho_sie specifically. If we intend to generalize this to all spiner EOS types, we should move it into the shared header.

singularity-eos/eos/eos_spiner_rho_sie.hpp
```diff
@@ -76,6 +78,321 @@ using namespace eos_base;
...
+  Real TSplitPoint = 1e4;
+
+  // Optional: fine grid bounds override (advanced use)
+  Real rhoFineMin = -1.0; // -1 means use diameter
+  Real rhoFineMax = -1.0;
+};
+
+// Method detection helpers for optional EOS methods
+// Isolated in separate namespace for easy C++20 migration to concepts
+namespace eos_builder {
```
Yurlungur: same as above---I'd prefer these be in teh shared header. These are actually generic enough, we should maybe have a header for "EOS concepts" perhaps in `singularity-eos/base`.

singularity-eos/eos/eos_spiner_rho_sie.hpp
```diff
@@ -1043,6 +1365,531 @@ inline SpinerEOSDependsRhoSieTransformable<
...
+  }
+
+  // Determine reference density (for piecewise grids)
+  Real rhoAnchor = std::isnan(params.rhoNormal) ? 1.0 : params.rhoNormal;
+  if (std::isnan(rhoAnchor) || rhoAnchor <= 0 || rhoAnchor > 1e8) {
+    rhoAnchor = std::sqrt(rhoMin * rhoMax); // geometric mean
+  }
+
+  // Construct Bounds objects for rho, T, and sie
+  using Bounds = table_utils::Bounds<3>;
```
Yurlungur, line 1097: this parsing logic (and the logic below) looks pretty much identical to the stuff in sesame2spiner. Is there any way we could share at least this logic? Or does that require too significant of a change to sesame2spiner's build system?

singularity-eos/eos/eos_spiner_rho_sie.hpp
```diff
@@ -1043,6 +1365,531 @@ inline SpinerEOSDependsRhoSieTransformable<
...
+        Real E_minus = source_eos.InternalEnergyFromDensityTemperature(rho, T - h);
+        Real dEdT = (E_plus - E_minus) / (2 * h);
+        dependsRhoT_.dPdE(j, i) = dPdT / dEdT;
+      }
+
+      // Fill the dTdE field on the rho-T grid
+      if constexpr (eos_builder::has_cv_rho_T_v<EOS>) {
+        Real cv = source_eos.SpecificHeatFromDensityTemperature(rho, T);
+        // Should I use robust::ratio here?
+        dependsRhoT_.dTdE(j, i) = 1.0 / cv;
```
Yurlungur: yes, and elsewhere. ;)

singularity-eos/eos/eos_spiner_rho_sie.hpp
```diff
@@ -1043,6 +1365,531 @@ inline SpinerEOSDependsRhoSieTransformable<
...
+        dependsRhoT_.dPdE(j, i) = rho * gamma;
+      } else if constexpr (eos_builder::has_gamma_rho_sie_v<EOS>) {
+        Real sie = sie_(j, i);
+        Real gamma = source_eos.GruneisenParamFromDensityInternalEnergy(rho, sie);
+        dependsRhoT_.dPdE(j, i) = rho * gamma;
+      } else if constexpr (eos_builder::has_P_rho_sie_v<EOS>) {
+        Real sie = sie_(j, i);
+        Real h = sie * 1e-6;
+        Real P_plus = source_eos.PressureFromDensityInternalEnergy(rho, sie + h);
+        Real P_minus = source_eos.PressureFromDensityInternalEnergy(rho, sie - h);
```
Yurlungur: despite the planning document suggesting a finite differences helper, seems it wasn't used. I think we want one to reduce boiler plate. But see my comment above about bounds checking.

singularity-eos/eos/eos_spiner_rho_sie.hpp
Yurlungur: I agree with @jhp-lanl that this can cause problems sometimes. Perhaps a future enhancement is adding a parameter that enforces these are always computed from FD if desired.

test/test_eos_spiner_constructors.cpp
```diff
@@ -0,0 +1,541 @@
...
+
+    params.matid = 1003;
+
+    WHEN("We construct with a fine grid") {
+      SpinerEOSDependsRhoSie spiner_eos(ideal_eos, params);
+
+      THEN("Interpolation should be very accurate") {
+        // Test at random points within bounds
+        constexpr int ntest = 10;
+        for (int i = 0; i < ntest; i++) {
```
Yurlungur, line 270: I don't know if it's necessary to check this on device, since we know spinereosdependsrhosie works on device, but thought I'd mention it.

Yurlungur (changes requested):
Thanks, @buechlerm this is really nice and it's functionality we've been missing for a long time. I have some review comments that entail varying amounts of work. Feel free to push back on them.

singularity-eos/eos/eos_spiner_rho_sie.hpp
jhp-lanl: Yeah I wouldn't have this hold anything up. I just wanted to comment for the future.

I think it would be a good research project for somebody in the future to really drill into what the bad effects of thermodynamic inconsistency are. There _are_ tools to look at this (EOSPAC team has some), but I'm not sure they really tell you how bad things can get, and what the effects are. I think especially with these derivatives, the effects depend on your formulation, and it's possible that the formulation can be written to make sure certain "important" derivatives are consistent.

singularity-eos/eos/eos_spiner_rho_sie.hpp
buechlerm: Checking these off as I get a chance.  But I moved all the method existence verification to base/eos_concepts.hpp  It still has the pre c++20 support.

plan_histories/MR632-2026-05-04-add-spiner-constructor-from-eos.md
buechlerm: Yeah, I let it plan, but didn't let it finish, so unfortunately, I didn't let it carry a good idea through.  

Regarding the bounds checking.  Are you suggesting: if at or below min eval at min using forward difference, if at or above max, eval at max using backward difference, otherwise something like it suggested in the plan?

buechlerm: I've made a start on removing the boiler plate, but I need to think more about the design to facilitate bounds checking.

buechlerm: I've incorporated density and temperature bounds.  Take a look and let me know if you like the design or suggest some changes.

singularity-eos/eos/eos_spiner_rho_sie.hpp
buechlerm: Probably still want to resolve difference with @adamdempsey90 sesame2spiner work, but I did move some things to a common header.

buechlerm:
Took me some time, sorry, but I believe I've addressed comments.

singularity-eos/eos/eos_spiner_rho_sie.hpp
adamdempsey90: I don't anticipate #641 taking too long and it is already working for submodule mode. We could start thinking about how to incorporate what you need in this MR into that MR

buechlerm: I'm sure there are lots of areas to look at, but this particular code is a structure that can be passed to the constructor.   I'll start there and we can expand later.  Do you think we should be using the map that sesame2spiner uses instead.  I see how when parsing from a file that approach works well, but the structure seems more natural (to my small brain anyway) for building in code.

buechlerm: I guess the map might be easier to extend?

singularity-eos/base/eos_concepts.hpp
```diff
@@ -0,0 +1,338 @@
...
+namespace eos_concepts {
+
+template <typename EOS>
+inline constexpr bool dependent_false_v = false;
+
+// C++17 implementation using detection idiom
+#if __cplusplus < 202002L
+
+// Detection helper
+template <typename...>
```
adamdempsey90, line 29: I don't think we need the c++17 concepts anymore, the c++20 implementations below are enough

singularity-eos/base/eos_concepts.hpp
```diff
@@ -0,0 +1,338 @@
+//------------------------------------------------------------------------------
+// © 2021-2026. Triad National Security, LLC. All rights reserved.  This
```
adamdempsey90: All the new file should just be `2026`

singularity-eos/eos/eos_spiner_rho_sie.hpp
adamdempsey90: I think the best thing to do is to move your struct into sesame2spiner and have it use that in place of where it currently uses the Params map (and in standalone mode it would fill the struct from reading the input files). Then, in singularity, you'd pass the struct to the saveMaterials function, it would write your sp5 file and then you'd call the normal ctor. So I think that would solve two things at once: merge the shared code and get this new ctor to dump the table.

buechlerm: For something you do once, I suppose it doesn't really matter, but doesn't it seem kludgy to fill all the databoxes you need for an eos, write it out to file throw it away and the read it in again.  Maybe I'm misunderstanding.

adamdempsey90: That may be, but it would ensure full compliance with whatever sesame2spiner decided on in the end and would ensure full backwards compatibility.

buechlerm: "decides on" in terms of the databoxes?  Isn't that decided by the spiner eos, not sesame2spiner?  Im certainly glad to use the saveMaterial function (though to be clear, the structure we were talking about only contains grid description not eos data), but I think the required data is already decided, right?

adamdempsey90: decides on meaning all of the parameters that sesame2spiner sets for each material (the ones in the input file). Your struct in spinerEOS is an alternative way of specifying all those parameters without having to write an input file, but the defaults might not line up with sesame2spiner (and we might change them / add new ones in the future).

But maybe I'm getting a little confused. As of today, a user builds a spiner table and everything about the EOS is determined by that user through sesame2spiner. The singularity spinerEOS just reads the table and pulls the settings from that. So, my expectation is that this new ctor for spinerEOS is just going to call sesame2spiner with the same parameters that the user could've run the standalone exe with.

Maybe I'm conflating the SESAME use case here with making a table from a different EOS where you don't have sesame2spiner

singularity-eos/base/eos_concepts.hpp
buechlerm: I'd have to change the CMakeList.txt to cxx_std_20.  Are we ready to do that?

singularity-eos/eos/eos_spiner_rho_sie.hpp
buechlerm: Sure, and I'm happy to use something sesame2spiner defines (instead of the struct I'm using here) in terms of the grid description, but the filling of the data is different than what sesame2spiner does.  There's certainly some shared approaches though that could be used in both.

buechlerm: So I believe spinerTableGridParams is now consistent with sesame2spiner and I have made sesame2spiner us it.

sesame2spiner/generate_files.cpp
```diff
@@ -289,8 +292,11 @@ herr_t saveTablesRhoT(hid_t loc, int matid, TableSplit split, const Bounds &lRho
   return status;
 }
 
-void getMatBounds(int i, int matid, const SesameMetadata &metadata, const Params &params,
-                  Bounds &lRhoBounds, Bounds &lTBounds, Bounds &leBounds) {
+// Convert string-based Params to structured SpinerTableGridParams
+// This bridges sesame2spiner's parameter file format with the SpinerEOS constructor API
+SpinerTableGridParams paramsToGridParams(int matid, const SesameMetadata &metadata,
+                                         const Params &params) {
+  SpinerTableGridParams gridParams;
```
Yurlungur, line 299: the naming style of sesame2spiner and singularity-eos is a bit eclectic due to the development history. Should we try to clean this up a bit? I used camelcase here before. Now would be an opportunity to change it if we wanted to.

sesame2spiner/generate_files.cpp
```diff
@@ -300,119 +306,126 @@ void getMatBounds(int i, int matid, const SesameMetadata &metadata, const Params
...
-  // is not needed. This improves resolution at low densities and
-  // temperatures.
+  // Note: Abar and Zbar are not in Params, they come from metadata in sesame2spiner
+  // Offsets are auto-computed by constructRhoBounds/constructTBounds, not in Params
 
-  // Extrapolation and other resolution tricks will be explored in the
-  // future.
-  if (rhoMin < STRICTLY_POS_MIN_RHO) rhoMin = STRICTLY_POS_MIN_RHO;
-  if (TMin < STRICTLY_POS_MIN_T) TMin = STRICTLY_POS_MIN_T;
+  return gridParams;
```
Yurlungur, line 399: Thanks for unifying this. I think it's much cleaner.

sesame2spiner/generate_files.cpp
Yurlungur, line 1: I *think* the logic here is identical but it was always a bit of a mess. If tests pass though I think that is convincing.

doc/sphinx/src/examples.rst
```diff
@@ -58,6 +58,82 @@ internal energy to evaluate at.
...
+
+The example verifies that the tabulated EOS matches the original custom
+EOS with interpolation errors typically less than 1%. The resulting
+``SpinerEOS`` can be used in production simulations for better
+performance, on GPUs via ``GetOnDevice()``, and with mixed-cell closure
+models.
+
+.. note::
+
+  The example uses ``SpinerEOSDependsRhoSie``, but you can also use
```
Yurlungur, line 118: Nice

doc/sphinx/src/models.rst
```diff
@@ -1720,6 +1720,216 @@ constructor for ``SpinerEOSDependsRhoSie`` is identical.
...
+
+  // Use like any other SpinerEOS
+  Real P = spiner_eos.PressureFromDensityTemperature(rho, T);
+  Real T_out = spiner_eos.TemperatureFromDensityInternalEnergy(rho, sie);
+
+.. note::
+
+  This constructor only supports ``TableSplit::Total``. Electron-only
+  and ion-cold splits are not available for generic EOS, as these
+  require additional physics that generic EOS models may not provide.
```
Yurlungur, line 1833: We could also add a note that there's nothing stopping a user from creating a custom analytic EOS separately for each subtable of a sesame table. e.g., build an EOS and then tabulate separately for electrons and ions.

doc/sphinx/src/models.rst
```diff
@@ -1720,6 +1720,216 @@ constructor for ``SpinerEOSDependsRhoSie`` is identical.
...
+
+  - Use ``SpinerEOSDependsRhoT`` when:
+    - Your code primarily uses (ρ,T) lookups
+    - Memory efficiency is important
+    - You don't need fast (ρ,sie) → P lookups
+
+  - Use ``SpinerEOSDependsRhoSie`` when:
+    - Your code uses both (ρ,T) and (ρ,sie) lookups
+    - You need direct P(ρ,sie) without root finding
+    - You're using mixed-cell closure models (PTE, etc.)
```
Yurlungur: Not necessarily accurate. The rho-T formulation is probably more natural with rho-T tables, though it shouldn't end up mattering much. 

The point that table format and PTE formulation may be matched for optimal performance is good, but I think it's more complex than this.

doc/sphinx/src/models.rst
```diff
@@ -1720,6 +1720,216 @@ constructor for ``SpinerEOSDependsRhoSie`` is identical.
...
+  // Use like any other SpinerEOSDependsRhoT
+  Real P = spiner_eos.PressureFromDensityTemperature(rho, T);
+  Real sie = spiner_eos.InternalEnergyFromDensityTemperature(rho, T);
+  Real T_inv = spiner_eos.TemperatureFromDensityInternalEnergy(rho, sie);
+
+.. note::
+
+  **Choosing between RhoT and RhoSie**:
+
+  - Use ``SpinerEOSDependsRhoT`` when:
```
Yurlungur, line 1912: Should we mention something about accuracy of root finds vs lookups?

singularity-eos/base/eos_concepts.hpp
Yurlungur: We aren't ready unfortunately. Tests fail with this swap on NVIDIA due to some issue with std::chrono pulling in std::format. @rberger is working on a fix. It's a tooling version issue.

singularity-eos/base/eos_concepts.hpp
Yurlungur, line 1: :+1:

singularity-eos/base/eos_concepts.hpp
```diff
@@ -0,0 +1,338 @@
...
+inline constexpr bool has_abar_v = has_abar<EOS>;
+
+template <typename EOS>
+inline constexpr bool has_zbar_v = has_zbar<EOS>;
+
+#endif
+
+} // namespace eos_concepts
+
+#endif // SINGULARITY_EOS_BASE_EOS_CONCEPTS_HPP_
```
Yurlungur: missing newline

singularity-eos/base/finite_diff.hpp
```diff
@@ -0,0 +1,72 @@
...
+Real centralDifference(Func &&f, Real x, std::optional<Real> pert = std::nullopt) {
+  //  f(x) is a function that calls some g(x1,x2) with either x1 or x2 fixed
+  Real h = pert ? *pert : std::max(std::abs(x) * 1e-6, 1e-12);
+  Real f_plus = f(x + h);
+  Real f_minus = f(x - h);
+  return robust::ratio(f_plus - f_minus, 2 * h);
+}
+
+template <typename Func>
+Real forwardDifference(Func &&f, Real x, std::optional<Real> pert = std::nullopt) {
```
Yurlungur, line 35: Looks like you have forward and backward differences still at 2nd order. Nothing wrong with that, but I was imagining dropping to 1st order at domain edges. Either way should be fine, though, assuming the function is sufficiently smooth.

singularity-eos/eos/eos_spiner_construction.hpp
Yurlungur, line 1: Nice. This is a better location for this.

singularity-eos/eos/eos_spiner_rho_sie.hpp
```diff
@@ -1043,6 +1056,428 @@ inline SpinerEOSDependsRhoSieTransformable<
...
+    maximumRho = 1.e99; // OK, what is a reasonable value?
+  }
+  Real maximumT = 1.e99; // OK
+
+  // Populate tables - dependsRhoT (sie, P, etc. as function of rho, T)
+  // Use shared construction utility
+  spiner_table_builder::populateDependsRhoT<EOS>(
+      source_eos, lRhoBounds, lTBounds, lRhoOffset_, lTOffset_, minimumRho, maximumRho,
+      minimumT, maximumT, sieMin, sieMax,
+      [this](int j, int i, Real v) { sie_(j, i) = v; },
```
Yurlungur, line 1274: No change needed now, but capturing this explicitly will become a warning in C++20. Unfortunately it's just a complete swap in the desired syntax between the two version. Maybe want to add a comment to this effect.

singularity-eos/eos/eos_spiner_rho_sie.hpp
Yurlungur: This looks a bit cleaner than the prior version. But I guess some boiler plate here is probably just unavoidable.

Yurlungur (approved):
This looks great now. Thanks for the lift! A few nitpicks still, but I think they're non-blocking.

plan_histories/MR632-2026-05-04-add-spiner-constructor-from-eos.md
Yurlungur: Sorry somehow I didn't see your responses. Looks great now!

singularity-eos/eos/eos_spiner_rho_sie.hpp
Yurlungur: I think for now this is a good approach. We might be able to clean up the way that the data is filled via some tricks with generic functions. But this is good enough for now.

singularity-eos/eos/eos_spiner_rho_sie.hpp
buechlerm: I thought you put in an MR recently that made **this** an explicit capture.  I'm sure I'm just misunderstanding.

singularity-eos/base/finite_diff.hpp
buechlerm: I'm happy to revert to 1st order at the boundaries if you prefer.  I should have considered it being unsmooth near the boundaries.

buechlerm: Using first order at the boundaries now.

singularity-eos/eos/eos_spiner_rho_sie.hpp
Yurlungur: Ah no you're right. I'm mis-remembering it as backwards.

singularity-eos/base/finite_diff.hpp
Yurlungur: Sounds good.

doc/sphinx/src/models.rst
buechlerm: I have TableSplit hard coded to Total right now, but I can change that.

Yurlungur: No I don't think it matters for now.

sesame2spiner/generate_files.cpp
buechlerm: The way I understand things, we use sesame2spiner to create materials.sp5 used in the other spinerEOS tests, so yes, they pass.

sesame2spiner/generate_files.cpp
buechlerm: I'm happy to comply with whatever convention we as  team want.  Is there a convention/style that I just didn't follow?

doc/sphinx/src/models.rst
buechlerm: Better?  I'm also ok with just deleting it if my description is too ham-handed.

sesame2spiner/generate_files.cpp
Yurlungur: In this particular code, you followed what I had done.

In newer code I've been trying to use snake case for variables and camel case for classes and methods.

doc/sphinx/src/models.rst
Yurlungur: Yep. It's better! And I like having some guidance documented.

singularity-eos/base/eos_concepts.hpp
buechlerm: Fixed Thanks

sesame2spiner/generate_files.cpp
buechlerm: I'm happy to fix this if now is the time.  I'm also happy to fix it after both this and https://github.com/lanl/singularity-eos/pull/641 are merged in.  What ever we think as a team

Yurlungur: Let's wait---we can do a style update in a single sweep later.

Yurlungur:
@buechlerm is this ready to go? re-git indicates yes? if so, merge when ready

PR #630, opened 2026-04-09
Thread execution spaces through vector API

Yurlungur:
## PR Summary

This MR resolves issue #628 and pulls in https://github.com/lanl/ports-of-call/pull/109 from `ports-of-call`. Here I expose the ability to select an execution space in the `singularity-eos` vector API:
- Each vector API call now optionally accepts an execution space argument. The default is `PortsOfCall::Exec::Device()` which is (when Kokkos is enabled) an alias to kokkos default execution space.
- To handle the huge number of overloads, I heavily leverage #629 so that the changes are relatively minimal.
- I did need to heavily use SFINAE/concepts to ensure that all the different template specializations played nice.
- The *python* API now explicitly calls the `PortsOfCall::Exec::Host()` execution space.
- As usual, EOSPAC is a bit of an odd one out. As far as I know, EOSPAC does not have the ability to select an execution space like this. For now, I simply document that explicitly requesting an execution space with EOSPAC is undefined behavior. It will do whatever the backend decides to do. We should resolve this at some point down the road in collaboration with the EOSPAC devs if we need to.

This MR used generative AI, but agents weren't set loose, so there's no proposed change plan. I made changes by hand and then had the robots copy my changes by example. 

I am still getting the CI to pass, but this is ready for review.

## PR Checklist

- [x] Adds a test for any bugs fixed. Adds tests for new features.
- [x] Format your changes by using the `make format` command after configuring with `cmake`.
- [x] Document any new features, update documentation for changes made.
- [x] Make sure the copyright notice on any files you modified is up to date.
- [x] After creating a pull request, note it in the CHANGELOG.md file.
- [x] LANL employees: make sure tests pass both on the github CI and on the Darwin CI
- [x] If ML was used, make sure to add a disclaimer at the top of a file indicating ML was used to assist in generating the file.
- [x] If Agentic AI was used, have the AI generate a "proposed changes" markdown file and store it in the `plan_histories` folder, with a filename the same as the MR number.

If preparing for a new release, in addition please check the following:
- [ ] Update the version in cmake.
- [ ] Move the changes in the CHANGELOG.md file under a new header for the new release, and reset the categories.
- [ ] Maintainers: ensure spackages are up to date:
  - [ ] LANL-internal team, update XCAP spackages
  - [ ] Current maintainer of upstream spackages, submit MR to spack

PR #634, opened 2026-04-12
sanitize python more

Yurlungur:
## PR Summary

This MR is a follow up of #630 and issue #628 . I found I was still seeing compilation errors on GPU builds with python active, and the reason was that the lambda was marked `__host__ __device__` even though the execution space was host. This MR adds a constexpr branch to switch to a host-only lambda when appropriate.

## PR Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 ### Changed (changing behavior/API/variables/...)
 
 ### Infrastructure (changes irrelevant to downstream codes)
+- [[PR634]](https://github.com/lanl/singularity-eos/pull/634) Disable device lambda when on host execution space
 - [[PR633]](https://github.com/lanl/singularity-eos/pull/633) Make robust::sgn handle unsigned properly
 - [[PR629]](https://github.com/lanl/singularity-eos/pull/629) Use macros and eos_base and eos_variant to reduce boiler plate
 - [[PR626]](https://github.com/lanl/singularity-eos/pull/626) Fix C++20 warnings related to lambdas
--- a/python/module.hpp
+++ b/python/module.hpp
@@ -34,12 +34,12 @@ using singularity::variadic_utils::np;
 
 // Helper function to convert lambda numpy array to double* buffer
 // With std::optional we would add support for a default value of lambda=None
-template<typename T, PORTABLE_FUNCTION Real(T::*Func)(const Real, const Real, Real*&&) const>
+template<typename T, Real(T::*Func)(const Real, const Real, Real*&&) const>
 Real two_params(const T& self, const Real a, const Real b, py::array_t<Real> lambda) {
   return (self.*Func)(a, b, lambda.mutable_data());
 }
 
-template<typename T, PORTABLE_FUNCTION Real(T::*Func)(const Real, const Real, Real*&&) const>
+template<typename T, Real(T::*Func)(const Real, const Real, Real*&&) const>
 Real two_params_no_lambda(const T& self, const Real a, const Real b) {
   return (self.*Func)(a, b, np<Real>());
 }
--- a/singularity-eos/eos/eos_base.hpp
+++ b/singularity-eos/eos/eos_base.hpp
@@ -176,10 +176,15 @@ char *StrCat(char *destination, const char *source) {
     static auto const name = SG_MEMBER_FUNC_NAME();                                      \
     static auto const cname = name.c_str();                                              \
     const CRTP &copy = *(static_cast<CRTP const *>(this));                               \
-    portableFor(                                                                         \
-        cname, s, 0, num, PORTABLE_LAMBDA(const int i) {                                 \
-          OUT[i] = copy.NAME(IN1[i], IN2[i], lambdas[i]);                                \
-        });                                                                              \
+    if constexpr (std::is_same_v<Space, PortsOfCall::Exec::Host>) {                      \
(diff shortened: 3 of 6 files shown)


adamdempsey90 (commented):
Do you have to do this manual switch on every function that could be called from python?

Yurlungur:
> Do you have to do this manual switch on every function that could be called from python?

Every C++ function. The alternative would be to do a for loop in Python directly, which would be a bit of a hack, but maybe worth it. That would prevent the Python from doing efficient EOSPAC calls though.

Yurlungur:
@adamdempsey90 do you object to this? I don't love it, but I do think it's the simplest/cleanest path to getting things working the way we need them to.

adamdempsey90:
> @adamdempsey90 do you object to this? I don't love it, but I do think it's the simplest/cleanest path to getting things working the way we need them to.

I don't know, I'm not a huge fan of having to write the same for loops twice. Do you understand the root cause of why the normal portable lambda can't be used?

Yurlungur:
> > @adamdempsey90 do you object to this? I don't love it, but I do think it's the simplest/cleanest path to getting things working the way we need them to.
> 
> I don't know, I'm not a huge fan of having to write the same for loops twice. 

Neither am I.

> Do you understand the root cause of why the normal portable lambda can't be used?

Yes. Because it marks it as `__host__ __device__`. But if that lambda then calls the `operator[]` for, say, a numpy array, that is not kosher, because the numpy array doesn't have device markings.

adamdempsey90:
> > > @adamdempsey90 do you object to this? I don't love it, but I do think it's the simplest/cleanest path to getting things working the way we need them to.
> > 
> > 
> > I don't know, I'm not a huge fan of having to write the same for loops twice.
> 
> Neither am I.
> 
> > Do you understand the root cause of why the normal portable lambda can't be used?
> 
> Yes. Because it marks it as `__host__ __device__`. But if that lambda then calls the `operator[]` for, say, a numpy array, that is not kosher, because the numpy array doesn't have device markings.

Maybe an alternative then is to have a small struct that wraps the numpy pointer (?) in an unmanaged host view?

Yurlungur:
> > > > @adamdempsey90 do you object to this? I don't love it, but I do think it's the simplest/cleanest path to getting things working the way we need them to.
> > > 
> > > 
> > > I don't know, I'm not a huge fan of having to write the same for loops twice.
> > 
> > 
> > Neither am I.
> > > Do you understand the root cause of why the normal portable lambda can't be used?
> > 
> > 
> > Yes. Because it marks it as `__host__ __device__`. But if that lambda then calls the `operator[]` for, say, a numpy array, that is not kosher, because the numpy array doesn't have device markings.
> 
> Maybe an alternative then is to have a small struct that wraps the numpy pointer (?) in an unmanaged host view?

That imposes a requirement on stride, which we currently don't impose. Though perhaps one could be clever.

adamdempsey90:
> That imposes a requirement on stride, which we currently don't impose. Though perhaps one could be clever.

What do you mean? There's `LayoutStride` that can accept whatever striding you want

adamdempsey90:
> > That imposes a requirement on stride, which we currently don't impose. Though perhaps one could be clever.
> 
> What do you mean? There's `LayoutStride` that can accept whatever striding you want

I should say, you just set the stride at runtime to whatever the numpy array has, but the View would be `LayoutStride` and `Unmanaged`

Yurlungur:
> > > That imposes a requirement on stride, which we currently don't impose. Though perhaps one could be clever.
> > 
> > 
> > What do you mean? There's `LayoutStride` that can accept whatever striding you want
> 
> I should say, you just set the stride at runtime to whatever the numpy array has, but the View would be `LayoutStride` and `Unmanaged`

We have to thread the numpy language for stride/layout into the Kokkos unmanaged view.

Yurlungur:
> > > That imposes a requirement on stride, which we currently don't impose. Though perhaps one could be clever.
> > 
> > 
> > What do you mean? There's `LayoutStride` that can accept whatever striding you want
> 
> I should say, you just set the stride at runtime to whatever the numpy array has, but the View would be `LayoutStride` and `Unmanaged`

Ah. messages crossing in flight. Yeah this is what I mean---the alternative is a bunch of additional helper code in the python module. Does numpy always provide contiguous memory? It never does some weird array of arrays thing?

adamdempsey90:
> > > > That imposes a requirement on stride, which we currently don't impose. Though perhaps one could be clever.
> > > 
> > > 
> > > What do you mean? There's `LayoutStride` that can accept whatever striding you want
> > 
> > 
> > I should say, you just set the stride at runtime to whatever the numpy array has, but the View would be `LayoutStride` and `Unmanaged`
> 
> Ah. messages crossing in flight. Yeah this is what I mean---the alternative is a bunch of additional helper code in the python module. Does numpy always provide contiguous memory? It never does some weird array of arrays thing?

From some consulting with GPT/Gemini and looking at https://pybind11.readthedocs.io/en/stable/advanced/pycpp/numpy.html, I think you should be able to write a very minimal array wrapper class that basically copies the strides/offsets/whatever else you need and the pointer and then provide `PORTABLE_INLINE_FUNCTION` operators that do the correct indexing.

If this works, it seems worth the effort now so that we don't have to always add double loops for every new function we want the python api to have access to.

Yurlungur:
I'm a dummy. The wrapper code *almost* already exists. This MR now superseded with #636 which implements your suggestion @adamdempsey90 .

PR #635, opened 2026-04-13
[trivial] silence the warning about this pointer

Yurlungur:
## PR Summary

Tiny, trivial change. It is safe to copy the `this` pointer in variant, due to how the recursive serialization routines work. However, C++20 warns. This makes the compiler happy.

## PR Checklist

### DIFF
--- a/singularity-eos/eos/eos_variant.hpp
+++ b/singularity-eos/eos/eos_variant.hpp
@@ -768,7 +768,7 @@ class Variant {
   }
   std::size_t DeSerialize(char *src,
                           const SharedMemSettings &stngs = DEFAULT_SHMEM_STNGS) {
-    memcpy(static_cast<void*>(this), src, sizeof(*this));
+    memcpy(static_cast<void *>(this), src, sizeof(*this));
     std::size_t offst = sizeof(*this);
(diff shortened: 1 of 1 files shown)


buechlerm (approved):
So we know that copying the raw bytes into the object here is safe because you wrote it all (and the target is effectly empyt at this point), but is there an approach that guarantees safety, or is this a case where this operation can only be written with that knowledge.  To be clear, I'm not asking you to do anything different, just trying to learn.

Yurlungur:
> So we know that copying the raw bytes into the object here is safe because you wrote it all (and the target is effectly empyt at this point), but is there an approach that guarantees safety, or is this a case where this operation can only be written with that knowledge. To be clear, I'm not asking you to do anything different, just trying to learn.

Copying is safe because we "fill in" after. So we want to use memcpy to copy the top-level structure of the struct. The pointers will be invalid after `memcpy` in the serialized data structure. But the data will be copied in after the struct body. Then after deserialization, the pointers are assigned properly.

adamdempsey90 (approved):

Yurlungur:
Waiting for the final tests to pass. They're stuck in the queue right now.

PR #637, opened 2026-05-06
fix error utils to work on architectures where std::is_normal is not …

Yurlungur:
…supported

## PR Summary

On a V100 architecture, @swjones discovered that `bad_value` was triggering on perfectly normal numbers, like $10^6$. This is apparently because `std::is_normal` is not guaranteed to be available or functional on all GPU architectures. This MR (hopefully!) remedies the situation by (a) implementing the constexpr version of `is_normal` ourselves and (b) testing it to make sure that sticks.

## PR Checklist


singularity-utils/singularity-utils/error_utils.hpp
```diff
@@ -78,7 +87,7 @@ PORTABLE_FORCEINLINE_FUNCTION bool non_positive_value(valT &&value, nameT &&var_
 }
 template <typename valT, typename nameT>
 PORTABLE_FORCEINLINE_FUNCTION bool negative_value(valT &&value, nameT &&var_name) {
-  return violates_condition(std::forward<valT>(value), is_strictly_positive{},
+  return violates_condition(std::forward<valT>(value), is_non_negative{},
```
Yurlungur, line 88: this is just a typo. this condition was identical to the one above.

Yurlungur:
One thing to note, the code
```C++
// Checks whether a value obeys some sort of provided condition. If not, returns true and
// prints the provided error message, variable name, and value (but does not abort!)
template <typename valT, typename condT, typename nameT>
PORTABLE_FORCEINLINE_FUNCTION bool violates_condition(valT &&value, condT &&condition,
                                                      nameT &&var_name) {
  const bool good = condition(std::forward<valT>(value));
  if (!good) {
    printf("### ERROR: Bad singularity-eos value\n  Var:   %s\n  Value: %.15e\n",
           var_name, value);
  }
  return !good;
}
```
is what actually calls the is_normal condition. I am torn about the print statement not being behind a `NDEBUG` preprocessor flag. On the one hand, the code continues if the check fails, and so this could be a very annoying error message. And it could impact performance. On the other hand, I caught this problem because of this print statement. So I'm not sure. Curious what people think.

singularity-eos/base/error_utils.hpp
```diff
@@ -25,14 +28,20 @@ namespace error_utils {
...
 struct is_normal_or_zero {
   template <typename valT>
   constexpr bool PORTABLE_FORCEINLINE_FUNCTION operator()(valT value) const {
     static_assert(std::is_floating_point<valT>::value);
+    using limits = std::numeric_limits<valT>;
+    const valT abs_value = (value < valT{0}) ? -value : value;
     return (value == valT{0}) ||
-           (std::isnormal(_NORMAL_FACTOR * value) && std::isnormal(value));
+           ((abs_value >= limits::min()) &&
+            (abs_value * static_cast<valT>(NORMAL_FACTOR) <= limits::max()));
```
adamdempsey90:
```suggestion
constexpr double PORTABLE_FORCEINLINE_FUNCTION NORMAL_FACTOR() { return 1.0e10;}

struct is_normal_or_zero {
template <typename valT>
constexpr bool PORTABLE_FORCEINLINE_FUNCTION operator()(valT value) const {
static_assert(std::is_floating_point<valT>::value);
    using limits = std::numeric_limits<valT>;
    const valT abs_value = (value < valT{0}) ? -value : value;
return (value == valT{0}) ||
           ((abs_value >= limits::min()) &&
            (abs_value * static_cast<valT>(NORMAL_FACTOR()) <= limits::max()));
```

Can't use globals on device with nvcc

adamdempsey90 (commented):
Might be something worth putting in ports of call since it's a portable std:: function?

adamdempsey90 (approved):

Yurlungur:
> One thing to note, the code
> 
> ```c++
> // Checks whether a value obeys some sort of provided condition. If not, returns true and
> // prints the provided error message, variable name, and value (but does not abort!)
> template <typename valT, typename condT, typename nameT>
> PORTABLE_FORCEINLINE_FUNCTION bool violates_condition(valT &&value, condT &&condition,
>                                                       nameT &&var_name) {
>   const bool good = condition(std::forward<valT>(value));
>   if (!good) {
>     printf("### ERROR: Bad singularity-eos value\n  Var:   %s\n  Value: %.15e\n",
>            var_name, value);
>   }
>   return !good;
> }
> ```
> 
> is what actually calls the is_normal condition. I am torn about the print statement not being behind a `NDEBUG` preprocessor flag. On the one hand, the code continues if the check fails, and so this could be a very annoying error message. And it could impact performance. On the other hand, I caught this problem because of this print statement. So I'm not sure. Curious what people think.

Needs to be behind debug because it's a performance hit

adamdempsey90:
> > One thing to note, the code
> > ```c++
> > // Checks whether a value obeys some sort of provided condition. If not, returns true and
> > // prints the provided error message, variable name, and value (but does not abort!)
> > template <typename valT, typename condT, typename nameT>
> > PORTABLE_FORCEINLINE_FUNCTION bool violates_condition(valT &&value, condT &&condition,
> >                                                       nameT &&var_name) {
> >   const bool good = condition(std::forward<valT>(value));
> >   if (!good) {
> >     printf("### ERROR: Bad singularity-eos value\n  Var:   %s\n  Value: %.15e\n",
> >            var_name, value);
> >   }
> >   return !good;
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > is what actually calls the is_normal condition. I am torn about the print statement not being behind a `NDEBUG` preprocessor flag. On the one hand, the code continues if the check fails, and so this could be a very annoying error message. And it could impact performance. On the other hand, I caught this problem because of this print statement. So I'm not sure. Curious what people think.
> 
> Needs to be behind debug because it's a performance hit

The alternative is to remove the printing of the name and value so that there are no arguments passed to it.

singularity-eos/base/error_utils.hpp
Yurlungur: `constexpr` scalars (and static arrays, insanely) are an exception:
https://godbolt.org/z/dWMP9or8q

Yurlungur:
Now uses the ports-of-call method 
https://github.com/lanl/ports-of-call/pull/112

To support that, I also bumped us to C++20.

.github/workflows/tests.yml
```diff
@@ -44,7 +44,6 @@ jobs:
                   -DSINGULARITY_PLUGINS=$(pwd)/../example/plugin \
                   -DCMAKE_LINKER=ld.gold \
                   -DCMAKE_BUILD_TYPE=Release \
-                  -DCMAKE_CXX_STANDARD=20 \
```
Yurlungur, line 47: no longer needed as we're now just on C++20

singularity-utils/singularity-utils/error_utils.hpp
```diff
@@ -52,16 +58,18 @@ struct is_non_negative {
...
 
-// Checks whether a value obeys some sort of provided condition. If not, returns true and
-// prints the provided error message, variable name, and value (but does not abort!)
+// Checks whether a value obeys some sort of provided condition. If not, returns true
+// and prints the provided error message, variable name, and value (but does not abort!)
 template <typename valT, typename condT, typename nameT>
 PORTABLE_FORCEINLINE_FUNCTION bool violates_condition(valT &&value, condT &&condition,
                                                       nameT &&var_name) {
   const bool good = condition(std::forward<valT>(value));
+#ifndef NDEBUG
```
Yurlungur, line 66: now added for performance and convenience

Yurlungur:
In the bump to C++20, I also bump the kokkos and kokkos-kernels submodules to 5.1.0. Note that this does not (yet) impact the spack builds. When we do our next release, we need to note the updated kokkos dependency.

singularity-eos/base/error_utils.hpp
adamdempsey90: Huh, I've definitely tried to do this across multiple code bases and it never would compile

adamdempsey90:
@Yurlungur what's the status with this? Ready to go?

Yurlungur:
> @Yurlungur what's the status with this? Ready to go?

Unfortunately no. I need to split out the fix vs the C++20 update. Tests are failing with C++20. Will do early next week.

Yurlungur:
CI now passing on darwin. Will merge as soon as the CI completes

Yurlungur:
Note the update to C++20 and Kokkos 5 necessitated treating Kokkos::initialize/finalize more carefully in the python bindings. This is a breaking change, and the Python vector calls must now be wrapped in a `singularity_eos.initialize()` or `singularity_eos.context()` context manager.

PR #643, opened 2026-07-08
only require library component if it is built

buechlerm:
This changes cmake so that when the archive library isn't built it isn't required by default either.  Please review vigorously as I'm not really a cmake expert.
## PR Summary

When closure isn't built an archive file isn't created.  This makes it so that users don't have a default requirement for an archive that doesn't exist.  There is a work around on their part, if we don't like my solution here.

## PR Checklist

- [x] If ML was used, make sure to add a disclaimer at the top of a file indicating ML was used to assist in generating the file.
- [x] If Agentic AI was used, have the AI generate a "proposed changes" markdown file and store it in the `plan_histories` folder, with a filename the same as the MR number.

- [x] Maintainers: ensure spackages are up to date:
  - [x ] LANL-internal team, update XCAP spackages
  - [x] Current maintainer of upstream spackages, submit MR to spack

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Current develop
 
 ### Added (new features/APIs/variables/...)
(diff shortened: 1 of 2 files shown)


rbberger (approved):
LGTM

Yurlungur (approved):
Since this is user-facing, we should probably have a changelog entry that we can point to in a release.

